PR #7, opened 2011-06-15
Added Firefox 5 and 6 to the extension table

evilpie:
- Added information about "function statements" in strict mode
- Removed the css, so the links don't look ugly

Just used the name "firefox 5" and did bother with adding beta, aurora etc.

Cheers
Tom


kangax:
Thanks, Tom! I also fixed some extraneous firefox4 entries that were in few places.

evilpie:
Did an horrific job at copy & paste :)

PR #9, opened 2012-01-03
NodeJS version of the strict mode test

dmitris:
Hi Juriy,

please consider pulling a version of your strict mode tests modified to run under NodeJS.  I removed the HTML tags and modified the output to print "OK" or "Not OK" at the beginning of the lines, for easier overview.

With Node v0.6.6, I'm getting "OK" for everything except:
Not OK arguments.caller; is a TypeError
Not OK (function f() { f = 123; })() is a TypeError

Thanks,

Dmitry

### DIFF
--- a/strict-mode/README
+++ b/strict-mode/README
@@ -1,0 +1,4 @@
+es5-compat-test.js is the adaptaion of the original index.html for NodeJS testing
+
+index.html is the original file by kangax (forked from 
+https://github.com/kangax/es5-compat-table/blob/gh-pages/strict-mode/index.html)
--- a/strict-mode/es5-compat-test.js
+++ b/strict-mode/es5-compat-test.js
@@ -1,0 +1,450 @@
+// adapted for NodeJS from http://kangax.github.com/es5-compat-table/
+//
+// https://github.com/kangax/es5-compat-table/blob/gh-pages/strict-mode/index.html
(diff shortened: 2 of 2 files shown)


kangax:
Hey, sorry for getting to this so late. I'm not quite following what you mean by "removed the HTML tags". How would [this page](http://kangax.github.com/es5-compat-table/strict-mode/) work then?

dmitris:
Hi - thanks for getting back!  The HTML page will not and is not supposed to work - the added file es5-compat-test.js should be used with NodeJS from the command line (like a unit test):

$ node es5-compat-test.js 
OK x = "foo" is a ReferenceError
OK eval('"\010"') is a SyntaxError
OK eval('010') is a SyntaxError
OK __i_dont_exist = 1; is a ReferenceError
OK eval = 1; is a SyntaxError
OK arguments = 1; is a SyntaxError
OK eval++; is a SyntaxError
OK arguments++; is a SyntaxError
Not OK arguments.caller; is a TypeError
OK arguments.callee; is a TypeError

kangax:
If we're adding compatibility with node.js, we need to make it work in both — browsers and node. You have to be able to go to [this page](http://kangax.github.com/es5-compat-table/strict-mode/) and see what works and what doesn't.

kangax:
We already have this.

PR #10, opened 2012-02-07
Add ES6 table

paulmillr:
First of all, when I saw tables were pure HTML, I decided to rewrite everything in webapp-style. It even worked on node.js. But then I encountered many problems (e.g. versioning) so I decided to just copy generated by javascript HTML to new file and send a pull request.

You may see webapp-style rewrite [on branch rewrite](https://github.com/paulmillr/es5-compat-table/tree/rewrite).

### DIFF
--- a/es6/index.html
+++ b/es6/index.html
@@ -1,0 +1,1001 @@
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>ECMAScript compatibility table</title>
+    <link rel="stylesheet" href="../master.css" type="text/css">
+    <script type="text/javascript" src="../master.js"></script>
+</head>
+<body class="es6">
+  <div id="header">
+    <h1>
+      <a href="http://kangax.github.com/es-compat-table/">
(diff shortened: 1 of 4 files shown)


kangax:
Thanks for this!

There's just one thing. Some of these are already in non-standard table, so I'm wondering if we'll be OK with duplication, or if we should remove them from there... Non-standard table has (results for) a wide range of browsers, and es6 one only has latest ones. So, for example, some of the earlier Opera supported "y" flag and you can see that in the table. ES6 one won't be able to show that, since it only focuses on latest browsers only — and rightfully so, IMO.

What do you think?

kangax:
Btw, my FF nightly (13.x) shows "Yes" for both "Maps" and "Sets" (it's "No" in your table).

paulmillr:
I've fixed FX nightly issue.

I think that we should remove duplicates from old table. Opera 10.1 seems to have <1% market share even in russian-speaking countries so I don't think it matters.

PR #11, opened 2012-02-22
trivial code fix and update table for Firefox Nightly

dynamis:
- fix lack of parenthesis
- set javascript version to enable yield keyword
- update for Firefox Nightly

### DIFF
--- a/es6/index.html
+++ b/es6/index.html
@@ -270,7 +270,7 @@
             </td>
             <script>test((function () {
           try {
-            eval('[a * a for (a of [1, 2, 3]][0] === 1');
+            eval('[a * a for (a of [1, 2, 3])][0] === 1');
             return true;
           } catch(error) {
             return false;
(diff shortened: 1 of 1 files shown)


paulmillr:
Js 1.8 isn't es6. Es6 spec soesn't say you should add the attribute to tag.

Same with "const" keyword in webkit, it works, but incorrectly and only without use strict.

And the table now is broken http://kangax.github.com/es5-compat-table/es6/

kangax:
I saw that it broke, but no time to fix at the moment. Will do whenever I get a chance.

kangax:
I fixed the table and specified versions of Chrome and Firefox. As far as "type" attribute, it's not ES6 concern and it certainly shouldn't be specifying its requirements. I haven't checked but I'd be very surprised if it did; none of earlier ES versions step into browser land. Why would they? Browser (and its engine) is just one of many implementations of ES.

PR #30, opened 2012-05-07
Add tests for ES6 code point escapes, `String.fromCodePoint`, and `String#codePointAt`.

mathiasbynens:
Add tests for ES6 code point escapes, `String.fromCodePoint`, and `String#codePointAt`.

See http://mathiasbynens.be/notes/javascript-escapes#unicode-code-point. 

Preview: http://mathiasbynens.github.com/es5-compat-table/es6/

### DIFF
--- a/es6/index.html
+++ b/es6/index.html
@@ -33,7 +33,7 @@
           })();
         </script>
         <span style="position:relative;top:-6px">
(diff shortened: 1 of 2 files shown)


kangax:
Nice! Thanks :)

PR #31, opened 2012-05-07
Better way of highlighting the second cell in each table row

mathiasbynens:
### Before

(There are two `<script>` elements in the Generators row, so the second script gets the outline rather than the table cell.)


kangax:
Thanks!

PR #36, opened 2012-10-12
Added Node 0.8 to ES6 table

rvagg:
Since V8 isn't compiled with experimental addons I thought it would be useful to have a column showing what's available in Node.js. I considered adding an 0.9 column but ES6 support is identical to 0.8 at the moment but that could change before the end of 0.9
Is this reasonable / useful?


kangax:
Thanks! Definitely useful.

PR #40, opened 2012-11-02
Update Map, Set, WeakMap to reflect latest ES6 draft (rev11)

rwaldron:
All of the changes within are additions that reached consensus over the last 4 months and are now present in the ES6 rev11 draft.

Browser support columns updated as well (unfortunate set backs)

Signed-off-by: Rick Waldron waldron.rick@gmail.com


rwaldron:
I've intentially left out the `delete` methods, probably for the same reason you had left them out

rwaldron:
(rebased)

PR #45, opened 2012-11-04
Firefox 18 supports maps

Skalman:
I noticed that the page stated that Firefox 18 does not support the ES6 Map, though my "Current browser" did, which happens to be Firefox 18 (I highly doubt it matters, but I'm using Aurora 2012-11-04).


kangax:
Thanks!

PR #47, opened 2012-11-06
Update and bugfix in ES5

syoichi:
I added Chrome 19+ and Opera 12.10 in ES5. These are "ALL YES".


kangax:
Thanks again! We have a bit of a gap now in Chrome (17, 18). I'll test later, see if they belong to 13-16 column or 19 one.

Issue #29, opened 2012-04-20
easier way to edit "tables"


tomByrer:
What is the easiest & fastest way to add a column (eg new browser version) for your tables please?  I was about to add FF14 for you, & I looked at it,  & said, "Table layouts = cool, not cool I have to hand- edit a new column."  What tools do you use/recommend, or can this be automated please?  New browser versions are added every few months now.

kangax:
I do it manually and yes it's painful. I'm open to suggestions if anyone knows anything.

tomByrer:
I'm not sure either, but I found a great website where one can ask these kinds of topics, have you ever heard of it?
http://stackoverflow.com/questions/10249939/tools-to-collum-rows-on-html5-tables
;P

Skalman:
I would probably prefer to generate the whole table in JavaScript. Are you interested in such a solution if I create one?

kangax:
If we do it with Node.js, sure.

Skalman:
With Node.js it couldn't be hosted on Github. Is a client-side table generation out of the question?

kangax:
I meant to precompile it with Node.js before pushing to github.

Skalman:
So do I hear you correctly in saying that client-side table generation is not ok and that Node.js must be used?

Alright. How would that work though? Is it even possible to prevent anybody from pushing without having precompiled first?

kangax:
Well, since it's about the same effort, I don't see a need to rely on JS
here. Is something wrong with "precompiling" table with Node.js before
pushing to github?

Skalman:
That means that you cannot make a pull request via Github, such as the one I did before, where you just change a single value. That said, Node.js works too. I'll look into it.

kangax:
For pull requests, you would need to change single value followed by `node
build.js` (or whatever). Shouldn't be much trouble. What do you think?

Skalman:
Not a big problem, but it means that I have to download everything to my own computer and run it. The pull request I did was only through Github, no running node, no cloning it to a local repository. I'm going with Node, though. :-)

kangax:
I see what you mean. I don't recall ever using web interface for pull
requests so completely forgot about that :)

kangax:
Fixed in https://github.com/kangax/es5-compat-table/commit/e9ff07aa803972b425fdc04828664b8420db4b2e

PR #48, opened 2012-11-24
A build system to convert data->table

Skalman:
This commit adds a script to generate the tables from data, as discussed in issue #29.

My goal was to generate HTML that was as close to `index.html` (ES5 tests) as possible, while also normalizing and simplifying it. For that reason the `index.html` (ES5) diff is quite readable, which you can't say about the other tables.

It works like this:
- The data is available in `data-es5.js`, `data-es6.js` and `data-non-standard.js`
- `skeleton.html` in the three directories (`./`, `es6/`, `non-standard/`) have the base HTML which is used as a template for the to-be-generated `index.html`
- Run `node build.js` to generate the `index.html` files

What do you think?

Edit: Oh, and you can view the result at http://skalman.github.com/es5-compat-table/.

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,0 +1,229 @@
+// Copyright 2012 Dan Wolff | danwolff.se
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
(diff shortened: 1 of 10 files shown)


kangax:
@Skalman Looks pretty amazing :) Thanks for doing this! Merging.

PR #61, opened 2013-02-10
Add Konqueror 4.9

sbrunner:
All in the title ;-)


kangax:
Thanks @sbrunner! Merged. What about `String.prototype.bold`? Looks like it's missing; I'm guessing it's `true`?

sbrunner:
Good catch, effectively it's true, I do an other pull request for that :-)

PR #65, opened 2013-04-26
Add Fx 19-23 and a note for Fx's ConstDeclaration to es6 table

teramako:
Add:
- Fx 19-23 columns
- Note for Fx's ConstDeclaration.  
  This is not full-compatible for ES6  because desn't support block scope.


kangax:
Closing as it's outdated now

PR #67, opened 2013-05-16
Fixed HTML error

JamyGolden:
There was a small, but annoying, little problem I couldn't leave unfixed any longer ;p
HTML `<del>` tag was used within an HTML attribute causing the html to "break". This fix removes that HTML tag.


kangax:
Thanks! :)

PR #70, opened 2013-06-20
Firefox 18+ support Proxies

nstepien:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

Damn right!


kangax:
thanks!

PR #71, opened 2013-06-21
Added Object.assign Test

thom4parisot:
Cf. http://people.mozilla.org/~jorendorff/es6-draft.html#sec-15.2.3.17

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -839,6 +839,32 @@ exports.tests = [
   }
 },
 {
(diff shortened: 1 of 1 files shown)


kangax:
Cool, thanks. We should add `Object.mixin` as well then.

thom4parisot:
Sure. I'll propose that too. I was unsure if you were adding APIs containing TODOs.

Thanks for commenting this out :-)

PR #73, opened 2013-06-21
No one supports the latest Proxy spec yet

nstepien:
Turns out I was wrong, Firefox 18+ implemented direct proxies as constructors. (?)
Chrome only implemented the old spec.


kangax:
I don't think you're wrong. In my nightly FF `typeof Proxy` is "object". I've been meaning to change it to "object" before, but forgot.

nstepien:
You have to do `new Proxy(...)` in FF while the spec doesn't have it that way: `Proxy(...)` so I guess it's wrong.

PR #74, opened 2013-07-15
update maps and sets for IE11

archaeron:
just tested it with version 11.0.9431.0


kangax:
Thanks! I forgot to update results after updating test (it was failing initialization with array)

archaeron:
great, thank you :)

PR #77, opened 2013-08-29
Add test case for destructuring

### LINKED ISSUE #37: Add destructuring to ES6 compat table
Hi! A test for destructuring (see http://wiki.ecmascript.org/doku.php?id=harmony:destructuring) is missing. Maybe something like this could work:

``` JavaScript
test((function () {
  try {
    return eval('(function() { var { foo } = { foo: 5 }; ; return typeof foo === "number" })()');
  } catch(error) {
    return false;
  }
})())
```

gmarty:
This PR adds a test case for destructuring (currently supported on FF2+ only).

I wasn't sure if there's a specific order for rules, so I just placed it randomly after Block-level function declaration.
This closes #37.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -935,6 +935,43 @@ exports.tests = [
   }
 },
 {
+  name: 'Destructuring',
+  link: 'http://wiki.ecmascript.org/doku.php?id=harmony:destructuring',
+  exec: function () {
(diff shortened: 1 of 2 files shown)


kangax:
Cool, thanks!

PR #78, opened 2013-09-15
Add Firefox 26

apaprocki:
Firefox 26 adds support for ES6 generators (yield). This can be verified in FF nightly (26).


kangax:
Saw it but didn't have time to update. Thanks! We need to also change data file though.

apaprocki:
Oops! Forgot to include the data file -- added it.

kangax:
Excellent. Thanks again!

PR #80, opened 2013-09-18
fixed a couple tests

calvinmetcalf:
made a couple fixes to the tests
- the let test for non-mozilla browsers wasn't running because the variable being used to store the results of part 1 was undefined not falsey
- same with yield except I had to add an actual test (copy and paste) to the callback.
- iterable was using out of date syntax (source [this post](http://domenic.me/2013/09/06/es6-iterators-generators-and-iterables/) by @domenic)
- WeakMap now has a comparable test to Map based on its capabilities.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -591,11 +591,7 @@ exports.tests = [
   link: 'http://wiki.ecmascript.org/doku.php?id=harmony:iterators',
   exec: function () {
     var obj = {
-      iterator: function() {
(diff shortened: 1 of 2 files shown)


kangax:
Hey @calvinmetcalf thanks for the fixes! But we need to change data file as well — https://github.com/kangax/es5-compat-table/blob/gh-pages/data-es6.js

PR #82, opened 2013-09-21
Row selection improvements: no iteration, easier to click link

Skalman:
Three improvements to selecting a row in this commit:
- Don't iterate over all the rows upon selection
- Previously you had to hover the table cell first and then go left to click the link. Now you can hover them directly or tab to them (if your browser supports tabbing links).
- IDs cannot contain space characters, so I changed those to `_`

### DIFF
--- a/build.js
+++ b/build.js
@@ -82,7 +82,7 @@ function dataToHtml(browsers, tests) {
     name, id;
   for (i = 0; i < tests.length; i++) {
     t = tests[i];
-    id = t.name.replace(/\s/g, '_');
+    id = t.name.replace(/^[\s<>&"]+|[\s<>&"]+$/g, '').replace(/[\s<>&"]+/g, '_');
     name = t.link ? ('<a href="' + t.link + '">' + t.name + '</a>') : t.name;
     body.push(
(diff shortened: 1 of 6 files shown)


build.js
```diff
@@ -82,7 +82,7 @@ function dataToHtml(browsers, tests) {
     name, id;
   for (i = 0; i < tests.length; i++) {
     t = tests[i];
-    id = t.name;
+    id = t.name.replace(/\s/g, '_');
     name = t.link ? ('<a href="' + t.link + '">' + t.name + '</a>') : t.name;
```
michaelficarra, line 86: `t.link` needs to have `"` escaped here. `t.name` has to have `<` and `&` escaped here. I'm sure there are a bunch of other characters that need to be escaped, too.

michaelficarra, line 86: `t.name.link(t.link)` should work. It must be a decade since I've seen that method used.

Skalman, line 86: True. I'm going to escape `"`, `<`, `&`, and `>` (for symmetry).

kangax:
@Skalman @michaelficarra thanks for catching these!

PR #83, opened 2013-09-22
Rename "generator expressions" to "generator comprehensions"; fix iterator test.

### LINKED ISSUE #55: Objects with `iterator` property should not be iterable
Although it's not finalized how users can create iterable objects, it's definitely _not_ going to be through a simple `iterator` property. Right now you're just testing against Firefox's nonstandard implementation.

I can't really see a good test for this, however, given that TC39 hasn't decided exactly how it will work. Probably via a symbol that you import from a standard module, but not clear.

domenic:
Fixes #55.

The `Symbol.iterator` property is new as of last Thursday's TC39 meeting; minutes for that should be published soon.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -549,7 +549,7 @@ exports.tests = [
   }
 },
 {
-  name: 'Generator expressions',
+  name: 'Generator comprehensions',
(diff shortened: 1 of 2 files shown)


kangax:
Thanks, Domenic!

PR #85, opened 2013-09-24
Add Math.fround test

topaxi:
See: https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Global_Objects/Math/fround

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -2331,5 +2331,38 @@ exports.tests = [
     node08: false,
     node08harmony: false
   }
+},
+{
+  name: 'Math.fround',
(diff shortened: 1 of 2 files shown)


kangax:
@topaxi Thanks! I don't see `Math.fround` in the spec though.

@rwaldron @domenic Do you guys know of its status?

rwaldron:
Yep, this is legit. It was temporarily known as Math.roundFloat32() (yuck!) but was changed to "fround" at the last meeting https://github.com/rwaldron/tc39-notes/blob/master/es6/2013-09/sept-17.md#55-mathroundfloat32

kangax:
@rwaldron Thanks, Rick. Adding it then.

rwaldron:
No problem! Anytime :)

Issue #75, opened 2013-07-16
Usability enhancements


fbender:
To improve readability of the list, I propose to: 
- [x] Drop/hide obsolete browsers, e. g. Firefox <17 (Fx 17 is ESR, so it will be around for at least a few months), Chrome <21 (usage numbers are near-zero or actually zero). 
- [ ] Make columns hideable (so one can only display the columns he is interested in). This can be achieved with a small JS (which I'm happy to provide, but it may take some time before I can get to it). 
- [x] Highlight current row (and, if possible, column) on mouseover. 
- [x] A way to unfocus browser, by clicking on the focused browser again
- [x] Separate browser+versions into 2 headers
- [x] Fixed header
- [ ] Replace bottom notes with inline tooltips (for convenience)
- [x] Expand subtests by clicking on entire cell (except link) not just "down" arrow

kangax:
@fbender All good suggestions and all of these I've been planning to do in a near future!

kangax:
@fbender If you're still interested in tackling these, I'd be happy to pull them in. Otherwise, I'll try to get to this within a month or so.

fbender:
I'll have a look at it this week, however may not finish until the end of next week (depending on other work). If I don't report here, feel free to take over!

kangax:
Sounds good.

fbender:
I have a minimal (somewhat dirty) fix at https://github.com/fbender/es5-compat-table/commits/es6-ux – do you agree with the course I take? Should I do something else?

No PR yet, because this fix would benefit from a refactoring, and it does not include item 2 (hiding arbitrary columns) & 3 (highlight current row/column) mentioned above. 

Edit: Also, I only did minimal testing in Firefox, though this should work in "modern" browsers incl. IE9 (IE8 may need a script to make the checkbox work). I don't know if the results get skewed because of the hiding, either.

kangax:
@fbender Yes, I think it looks reasonable, although I already have a way of hiding old browsers in ES5 table. Not sure if you've seen — https://github.com/kangax/es5-compat-table/blob/gh-pages/master.css#L40-L46 It's purely presentational and done in CSS.

fbender:
[...]

PR #86, opened 2013-10-07
Usability enhancements, Part 1 (issue #75)

fbender:
This PR hides the columns of obsolete browsers by default (as per issue #75) in the ES6 compat table only. 

The changes work as expected in current versions of the major browsers. Does not work in IE <10 (but fails gracefully resp. does not change the current situation where the ES6 page is already visually broken in IE 9 and older). 

Let me know if I should alter this proposal, and/or apply the same functionality to the other tables, too. 

The code will get some refactoring in subsequent PRs.


kangax:
Hey @fbender, would you be able to resubmit this? Can't merge anymore.

fbender:
Yes, I'll look into it tomorrow or the day after.

fbender:
Sorry, took me a bit longer. Contiues in #88.

PR #93, opened 2013-11-24
Update Firefox results, change Fx 26 to 27

fbender:
Corrected some false positives/negatives (Fx25 through Fx27), whilst finding that an extra Fx26 column does not make sense, but a Fx27 one would.


kangax:
@fbender Thanks! Is FF26 identical to 25 or 27? We should clarify that (e.g.: FX25-26 or FX26-27)

fbender:
Pretty much identical to Fx25 with the exception of `Math.fround()` (see note in Fx27 column). I don't think this needs further clarification, because the "message" is: Fx 25 has these features, and since Fx 27 there were new features added. If you want to "merge" Fx25 and Fx26 into one column, you should move the `Math.fround` result including note to Fx25-Fx26.

kangax:
@fbender Fair enough. But looking at FF27 and 28 I see that they are entirely identical?

fbender:
Yep, that's plausible. However, Firefox 28 is currently in Nightly, so until the next merge (9th Dec, iirc), things could change easily. However if they don't until this point, you can simply rename Fx 28 to Fx 29 (and Fx 27 to Fx 27-28).

kangax:
@fbender perfect, thanks!

fbender:
No worries! Thanks for the quick turnaround!

PR #106, opened 2014-01-18
Symbols

Yaffle:
only, Chrome 33

Traceur is trying to polyflill symbols , seems, but Object.keys().length is not zero, although Object.getOwnPropertyNames is redefined to skip string, which are "symbols" in Traceur

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -1864,6 +1864,55 @@ exports.tests = [
   }
 },
 {
+  name: 'Symbol',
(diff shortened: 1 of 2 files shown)


kangax:
Need to change https://github.com/kangax/es5-compat-table/blob/gh-pages/data-es6.js as well.

Yaffle:
ok....

kangax:
Thanks!

PR #108, opened 2014-02-19
Fixup #103

fbender:
#103 (commit dd1e672) only touched the index.html thereby reverting the changes upon re-generating. This PR properly fixes the issue.

(You really should not accept PRs that _only_ touch index.html – the changes will get lost sooner or later.)


kangax:
I don't :) That was just an oversight. Thanks for noticing.

PR #110, opened 2014-02-24
PhantomJS 1.9.7 added

victor-homyakov:
The most important compatibility note about PhantomJS is that it still hasn't `Function#bind`: https://github.com/ariya/phantomjs/issues/10522


kangax:
Cool, thanks!

PR #112, opened 2014-03-09
Updated Node harmony data

pepkin88:
Updated Node harmony data and upgraded to version 0.11.11 (Math.sign and
Math.trunc were added since 0.11.7).


kangax:
Thanks!

PR #113, opened 2014-04-01
Added data for Phantom JS and Firefox 28

victor-homyakov:
- added Phantom JS data for `Math.cbrt` on ES6 page
- added Firefox 28 data on non-standard page
- fixed IE names on non-standard page


kangax:
Thanks!

PR #114, opened 2014-04-01
Non-standard `String.prototype.replace(substr, newSubStr, flags)`

victor-homyakov:
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Parameters


kangax:
Interesting. Thanks!

kangax:
I'll fill in the results for the other browsers later.

PR #115, opened 2014-04-07
add es5/es6 data for Echo JS (https://github.com/toshok/echo-js)

toshok:
data from current git master


kangax:
Hey @toshok, do you mind resubmitting this PR? I imagine you're even more compliant right now.

toshok:
sure, will do

PR #118, opened 2014-04-28
Update comprehension syntax

arv:


arv:
I wasn't sure how you handle updating the data tables?

kangax:
Thanks! Yeah, I'll need to retest all the browsers to update results in data-es6.js (then re-build table)

PR #119, opened 2014-04-29
firefox 29 supports promises

lholmquist:
firefox 29, released today, supports ES6 promises by default


kangax:
Thanks, but need to change data-es6.js as well.

lholmquist:
@kangax updated

kangax:
Thanks!

PR #130, opened 2014-06-04
Number.prototype.toInteger was removed from ES6

rwaldron:
https://github.com/rwaldron/tc39-notes/blob/master/es6/2013-07/july-25.md#consensusresolution-2
Signed-off-by: Rick Waldron waldron.rick@gmail.com

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -2743,52 +2743,6 @@ exports.tests = [
   }
 },
 {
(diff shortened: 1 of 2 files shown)


kangax:
Thanks!

rwaldron:
No problem! I'm just sorry that I haven't been keeping up with compat table :| I will try to be more diligent

PR #135, opened 2014-06-05
Cleanup

tschneidereit:
Removed some Object.\* functions that aren't in the spec after all, fixed up some naming and linkified some stuff.


tschneidereit:
Oh, silly me. I'll update the data js file.

tschneidereit:
Ok, this is better

kangax:
Thanks, @tschneidereit!

PR #149, opened 2014-06-25
Firefox 33 implemented Symbols (Firefox bug 645416)

cpeterso:
Add Firefox 33, obsolete Firefox 25–29, and note that Firefox 33 implemented Symbols: https://bugzilla.mozilla.org/show_bug.cgi?id=645416


kangax:
Thanks @cpeterso! What's going on with Firefox bug 688857 (Template literals). I see one of the (and the only remaining one) depending bug was marked as resolved today, just few hours ago. Didn't have time to check nightly yet; should we add it as well?

kangax:
@cpeterso also, FF24 should be marked as obsolete as well then?

cpeterso:
@kangax: FF24 is the current ESR ([Extended Support Release](http://www.mozilla.org/en-US/firefox/organizations/)) for enterprises that want support for a full year, not just six weeks. Your compat table is getting pretty crowded and ESR is not for mainstream users, so FF24 can be marked obsolete.

I don't know the status of Template Literals, but I will follow up.

PR #159, opened 2014-07-16
update for ejs after the big ES6 update on kangax's table

toshok:
Hey @kangax, here's the ejs changes after your big update.


kangax:
Great, thanks!

PR #162, opened 2014-07-16
fixed symbol.iterator sample

muratsu:
fixed the previously incorrect Symbol.iterator sample


rwaldron:
LGTM

webbedspace:
No complaint except that it's maybe a little unintuitive to read, due to using postfix ++ and c === 0. I think prefix ++ and c === 1 might work a weensy bit better.

PR #163, opened 2014-07-17
use Object.defineProperty to provide a Symbol.create impl.  fixes #161

### LINKED ISSUE #161: Symbol.create test is wrong
The test does:

`b[Symbol.create] = function() { a = 4; return {};};`

but this doesn't work.  `Symbol.create` is spec'ed to be in the prototype chain of `b` by virtue of it being in the built-in function prototype.  It's a non-writable property there, so the above assignment should silently fail.

To properly test for `Symbol.create` support using this sort of test, it needs to do:

`Object.defineProperty(b, Symbol.create, { value: function() { a = 4; return {}; } });`

toshok:


webbedspace:
Actually, this raises an additional point - the general Symbol test should be updated to use both Object.defineProperty and plain assignment to create symbol-keyed properties on an object.

rwaldron:
LGTM

I had forgotten about the @@create that's in the prototype chain. Whoops!

PR #164, opened 2014-07-17
Tightened the let and const tests

webbedspace:
This changes the ES6 let and const tests, requiring them to implement the "temporal dead zone", the for-let loop's fresh bindings specified in [rev 23](http://wiki.ecmascript.org/doku.php?id=harmony:specification_drafts#april_5_2014_draft_rev_23), and thrown syntax errors for incorrect const usage. Also, strict mode ("extended mode") and the "application/javascript;version=1.8" attribute are no longer added to the test.

Didn't get around to testing if IE11 or EJS passes these altered tests.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -311,44 +311,44 @@ exports.tests = [
   link: 'http://wiki.ecmascript.org/doku.php?id=harmony:const',
   exec: function () {
     try {
-	  return !!Function(
-		+'const foo = 123;'
-		+'var passed = (foo === 123);'
-		
-		 // bar is not hoisted outside of its block,
-		 // baz is not hoisted outside of the for-loop,
-		 // and qux is not defined until its let statement is executed.
-		 
-		+'{ const bar = 456; }'
-		+'for(const baz = 0; false;) {}'
-		+'passed &= (function(){ try { bar; } catch(e) { return true; }}());'
-		+'passed &= (function(){ try { baz; } catch(e) { return true; }}());'
-		+'passed &= (function(){ try { qux; } catch(e) { return true; }}());'
-		+'const qux = 789;'
-		
-		 // uninitialized const is a syntax error (13.2.1.1)
-		 
-		+'passed &= (function() {'
-		+'  try { Function("const a;")();} catch(e) { return true; }'
-		+'}());'
-		
-		 // duplicate consts are syntax errors (13.2.1.1)
-		 
-		+'passed &= (function() {'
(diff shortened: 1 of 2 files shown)


rwaldron:
Might as well add TDZ tests here as well. Reference: https://gist.github.com/rwaldron/f0807a758aa03bcdd58a

webbedspace:
@rwaldron I believed [these lines](https://github.com/webbedspace/compat-table/blob/0220c633551bdbb5dbfcb918f6d0dba242770657/data-es6.js#L408-L409) were sufficient to test TDZ? Or did I fail to include something or other?

rwaldron:
Yes, that's good for `let`, but still needs to be done for `const`. TBH, I had originally written that comment as a line comment in the `const` tests and then just pasted it here instead. Admittedly, when I saw no TDZ test in `const` I didn't read any further, sorry about that.

webbedspace:
Well... isn't there a largely identical const check [right here?](https://github.com/webbedspace/compat-table/blob/0220c633551bdbb5dbfcb918f6d0dba242770657/data-es6.js#L323-L324)

rwaldron:
@webbedspace haha, yep. I overlooked this—sorry for the noise and more importantly: great work :)

rwaldron:
@webbedspace if you wouldn't mind splitting up the tests like this: 

``` js
 // bar is not hoisted outside of its block,

+'{ const bar = 456; }'
+'passed &= (function(){ try { bar; } catch(e) { return true; }}());'

// qux is not defined until its const statement is executed.
// Accessing an undeclared lexical binding will result in a ReferenceError.

+'passed &= (function(){ try { qux; } catch(e) { return true; }}());'
+'const qux = 789;'
```

And for let: 

``` js
// bar is not hoisted outside of its block,

+'{ let bar = 456; }'
+'passed &= (function(){ try { bar; } catch(e) { return true; }}());'

// baz is not hoisted outside of the for-loop,

+'for(let baz = 0; false;) {}'
+'passed &= (function(){ try { baz; } catch(e) { return true; }}());'

// qux is not defined until its let statement is executed.
// Accessing an undeclared lexical binding will result in a ReferenceError.

+'passed &= (function(){ try { qux; } catch(e) { return true; }}());'
+'let qux = 789;'
```

webbedspace:
OK.

rwaldron:
@webbedspace brilliant and thanks for your patience :) 

@kangax LGTM

PR #167, opened 2014-07-24
ES6: Added string-spreading to the spread tests

webbedspace:
A recent [spec change](http://wiki.ecmascript.org/doku.php?id=harmony:specification_drafts#july_18_2014_draft_rev_26) permits strings to be spread, as Firefox currently allows. I added that test to both spread tests, as well as checks that non-iterables throw a TypeError when spread.

**Note**: I accidentally based this branch off #164, so you ought to merge that before considering this.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -311,44 +311,44 @@ exports.tests = [
   link: 'http://wiki.ecmascript.org/doku.php?id=harmony:const',
   exec: function () {
     try {
-	  return !!Function(
-		+'const foo = 123;'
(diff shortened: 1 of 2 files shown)


kangax:
@webbedspace I'm thinking perhaps to add string spreading as a separate test? Also, could you please resubmit this without #164, as I'd like to address that separately and probably sometime later (once I get some free time).

webbedspace:
OK, thanks. See #172 now.

PR #168, opened 2014-07-25
Tightened the arrow functions test

webbedspace:
The arrow test now checks:
- 1 parameter with no brackets
- multiple parameters
- absence of `prototype` and inability to be used with `new` (Traceur fails this check)
- lexical `this` binding
- lexical `arguments` binding (Firefox fails this check)

This last bullet-point gives me pause, though - I think it isn't too useful or helpful to the reader to report Firefox/Traceur failing support when the "core functionality" (which I regard as basic syntax + lexical `this`) is present, but of course it would be quite incorrect to give them a "Yes". So, I feel that after this is accepted, it should be a priority to implement "partial support" test results, differentiating core functionality checks from full support checks.

Note: I accidentally based this branch off #164, so you ought to merge that before considering this.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -311,44 +311,44 @@ exports.tests = [
   link: 'http://wiki.ecmascript.org/doku.php?id=harmony:const',
   exec: function () {
     try {
-	  return !!Function(
-		+'const foo = 123;'
-		+'var passed = (foo === 123);'
-		
-		 // bar is not hoisted outside of its block,
-		 // baz is not hoisted outside of the for-loop,
-		 // and qux is not defined until its let statement is executed.
-		 
-		+'{ const bar = 456; }'
-		+'for(const baz = 0; false;) {}'
-		+'passed &= (function(){ try { bar; } catch(e) { return true; }}());'
-		+'passed &= (function(){ try { baz; } catch(e) { return true; }}());'
-		+'passed &= (function(){ try { qux; } catch(e) { return true; }}());'
-		+'const qux = 789;'
-		
-		 // uninitialized const is a syntax error (13.2.1.1)
-		 
-		+'passed &= (function() {'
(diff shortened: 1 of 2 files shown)


kangax:
Hey @webbedspace!

It's good you mention "partial support" as I've been meaning to address this as well. This PR and let/const changes are both starting to grow into quite extensive set of tests, rather than simple one-liners that we've always had.

Now, there's nothing wrong with that per se... but on a spectrum of simple feature existence (e.g. `typeof Set !== 'undefined'`) to a full-blown conformance test (a la [Sputniktests](https://code.google.com/p/sputniktests/), which I ported to [web version back in the days, btw](https://code.google.com/p/sputniktests/) :) and now we have what... [official test262](http://test262.ecmascript.org/), although not sure if it includes ES6 or not) I always tried to err on a side of simplicity. So that a person could hover over a result and see a relatively simple expression showing the feature at hand. Kind of for educational and introductory purposes.

I'm not against more thorough tests.

I just feel like we then need to come up with some kind of better system of representing results — such as partial (percentage?) support — and possibly 2 versions of a test (simple and complex). If we do that, there's also a lot of room for improvement when it comes to conformance. Practically every test could be made into multiple checks, just following specs. Here's an example of a [shitload of tests Mozilla is adding for computed property names](https://bugzilla.mozilla.org/show_bug.cgi?id=924688#c3) that they're working on right now.

So I imagine this all would require more maintenance. And it would also probably be unwise to duplicate any other efforts creating/maintaining ES6 conformance suite that TC39 might have in mind (or already in the works?).

Just throwing it all out here for discussion, before proceeding with adding anything more extensive.

/cc @rwaldron @domenic

domenic:
We are trying hard to make test262 contribution friendly and _the_ place to put such stricter conformance tests... I'd encourage you to open a PR against test262 and we'll work hard to integrate it and make the process seamless :)

PR #172, opened 2014-07-31
ES6: Added separate spreadable strings check

webbedspace:
This is a re-do of #167.

(You may notice the es6 index.html file has a number of extra changed lines. This is because other contributors didn't freshly rebuild it before committing their recent changes.)


kangax:
Thanks!

Issue #128, opened 2014-06-04
don't fade feature name column


aeosynth:
when a column is highlighted, the feature name column should not be faded along with the others

kangax:
Good suggestion, will do.

PR #176, opened 2014-08-04
Some CSS changes

webbedspace:
This contains:
1) A bugfix for #128.
2) A new feature: when "show obsolete browsers" is enabled, the added columns are drawn using slightly desaturated colours to offset them from the current browsers' columns.


kangax:
Thanks!

PR #182, opened 2014-08-06
List annex B features as optional on non-browsers

webbedspace:
Related to the discussion in #178, which asserts that the red colour for "No" entries in the Traceur/EJS/etc. columns is misleading for Annex B features.... here's a proposal that changes Annex B rows so that the results for non-browsers are de-emphasised, signifying that they are optional. The cells are drawn in gray, and a title attribute states that "This feature is optional on non-browser platforms."

They still say "Yes" and "No", purely to allow the data to remain, but it isn't expressed as important to ES6 compatibility.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -188,7 +188,8 @@ exports.browsers = {
   },
   phantom: {
     full: 'PhantomJS 1.9.7 AppleWebKit/534.34',
-    short: 'PH'
+    short: 'PH',
+    nonbrowser: true,
   },
   node: {
     full: 'Node 0.10',
--- a/data-es6.js
+++ b/data-es6.js
@@ -5310,3 +5310,7 @@ exports.tests = [
   separator: 'after'
 }
 ];
+
+//Shift annex B features to the bottom
+exports.tests = exports.tests.filter(function(e) { return !e.annex_b })
+        .concat(exports.tests.filter(function(e) { return  e.annex_b }));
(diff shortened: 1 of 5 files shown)


kangax:
That's not a bad idea! I was thinking to position those tests on the bottom of a table; they are from Annex after all. Perhaps also exclude from features counting. We can do both — position at the bottom and remove colors. What do you think?

webbedspace:
Actually, I forgot to mention: they already are excluded from the feature count for non-browsers (as in, EJS is "67/96" instead of "67/100", and is actually sorted higher than FF34 at 68/100).

kangax:
Ok good, so let's just put them on a bottom then (plus whatever you already changed)

webbedspace:
This is ready to go.

data-es6.js
```diff
@@ -5300,3 +5310,7 @@ exports.tests = [
   separator: 'after'
 }
 ];
+
+//Shift annex B features to the bottom
+exports.tests = exports.tests.filter(function(e) { return !e.annex_b })
```
kangax, line 5315: Won't work in pre-ES6 filter-less environments, but not sure we care...

webbedspace, line 5315: Well, the build script runs in Node, not the browser.

kangax, line 5315: Wasn't looking carefully. Never mind then :)

PR #183, opened 2014-08-06
Updated Konqueror results to 4.13

webbedspace:
Results obtained via browsershots.net.


kangax:
nice, thanks!

webbedspace:
Just for reference: it's not a mistake that KQ4.13 is listed as supporting Math.acosh but not Math.asinh. As you can see [here](http://marc.info/?l=kde-commits&m=139014989807635&w=2), "asinh" was typoed in the commit that introduced it. :P

kangax:
That's amazing :) I'm also quite surprised to see _any_ ES6 support in KQ at all... I was under impression it's not actively developed, basically dying; even considered dropping it from the tables at some point. I mean, they don't even have full ES5 support — http://kangax.github.io/compat-table/es5/#konq49 — or at least _didn't_ have. Speaking of that, we should update ES5 table as well. I imagine all of those red ones are now green?

webbedspace:
Well, I just ran it through browsershots.org (not .net, sorry) and... [you're right](http://i.imgur.com/ZvPwywM.png). Everything except strict mode and zero-width chars.

Incidentally, if you're looking for things to axe, I've been lately wondering whether Rhino should be dropped or replaced with Nashorn...

Mouvedia:
@webbedspace I am curious, do you have intermediate results between version 4.9 and 4.13?
I am trying to find the first ES5 ready version of Konqueror (KHTML).

PR #187, opened 2014-08-07
Non-standard: Remove features included in ES6

### LINKED ISSUE #150: Non-standard: remove properties specced in ES6 Annex B
The following "non-standard" ES properties/things are now standardised in ES6: `const`, `let`, RegExp `y` flag, function `name` property, `Object#__proto__`, `String#anchor`, `String#blink`, `String#big`, `String#bold`, `String#link` (etc., but NOT `String#quote`),  `String#substr` (actually specced in ES5 annex B) and octal literals (actually specced in ES5 annex B).

webbedspace:
This removes the following items from the non-standard table which are
now in ECMAScript 6:
- function statement
- function "name" property (this has been added to the ES6 table)
- `__proto__`
- const, let
- RegExp "y" flag
- most of the String HTML methods
- String#substr
- octal

Closes #150.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -2433,6 +2433,53 @@ exports.tests = [
   }
 },
 {
+  name: 'function "name" property',
+  link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-setfunctionname',
+  exec: function () {
+    return (function foo(){}).name == 'foo';
+  },
+  res: {
+    tr:          false,
+    ejs:         true,
+    ie10:        false,
+    ie11:        false,
+    firefox11:   true,
+    firefox13:   true,
+    firefox16:   true,
+    firefox17:   true,
+    firefox18:   true,
+    firefox23:   true,
+    firefox24:   true,
+    firefox25:   true,
+    firefox27:   true,
+    firefox28:   true,
+    firefox29:   true,
+    firefox30:   true,
+    firefox31:   true,
+    firefox32:   true,
+    firefox33:   true,
+    firefox34:   true,
+    chrome:      true,
+    chrome19dev: true,
+    chrome21dev: true,
+    chrome30:    true,
+    chrome33:    true,
+    chrome34:    true,
+    chrome35:    true,
+    chrome37:    true,
+    safari51:    true,
+    safari6:     true,
+    safari7:     true,
+    webkit:      true,
+    opera:       true,
+    konq49:      true,
+    rhino17:     true,
+    phantom:     true,
+    node:        true,
+    nodeharmony: true
+  }
+},
+{
   name: 'Function.prototype.toMethod',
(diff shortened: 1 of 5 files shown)


kangax:
@webbedspace Looks good, thanks! Didn't know about "substr" being in ES6 now... As far as "function statement", you mean "block-level function declaration" in ES6 table? I see that it's testing `{function f(){}} typeof f == "undefined"` and there's definitely at least couple more ways to test for it. `if (false) { function f() { } }` being one of them. In pre-ES6 envs. it would simply hoist to the top, IIRC. Also `if (true) { function f(){} } else { function f(){} }` and, finally, with "use strict".

kangax:
Btw, related tests by T.J. Crowder — http://jsbin.com/sedetiyo/1/edit (loose) and http://jsbin.com/sedetiyo/2/edit (strict)

webbedspace:
- Good point about block function hoisting. I guess I'll add an ES6 annex B test for the non-strict hoisting behaviour (given that it's in [B.3.3](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-block-level-function-declarations-web-legacy-compatibility-semantics).)
- [String#substr is actually in ES3.](http://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262,%203rd%20edition,%20December%201999.pdf) (back when Annex B was purely non-normative)

webbedspace:
Regarding those linked tests: test 1 is not adherent to the ES6 B.3.3 spec (which I think states that all block-level function declarations (BLFDs) are hoisted out in order of appearance, and, by implication, later bindings trump earlier bindings), so I wonder if you want a non-standard test for this particular block-level behaviour to be added. (Alas, I don't really have the resources to test that in most of the increasingly aged environs in the non-standard table - IE7, BESEN, etc.)

webbedspace:
Also, octal was in ES3's annex B as well, so it's not been moved to the ES5 or ES6 table either.

(Even though annex B features prior to ES6 were not normative, and their implementation by platforms prior to ES6 is _arguably_ "non-standard", I've elected to only include annex B features on the ES6 list that were not mentioned in either ES3's annex B, or ES5's annex B, and thus are genuinely new to any version of the spec - even if, in the case of the string HTML methods, they predate even some of ES5's annex B features.)

kangax:
> test 1 is not adherent to the ES6 B.3.3 spec (which I think states that all block-level function declarations (BLFDs) are hoisted out in order of appearance, and, by implication, later bindings trump earlier bindings), [...]

I didn't even see that section on BLFD in ES6, since it was always disallowed in pre-ES6 times. Should probably update NFE article with that — http://kangax.github.io/nfe/#function-declarations-in-blocks

Thanks for bringing it up.

kangax:
*disallowed but still omnipresent as a de-facto, of course :)

PR #189, opened 2014-08-08
Tightened the arrow function tests, version 2

webbedspace:
This is an alternative approach to #168. The discussion for that PR emphasised that the tightened test, while more thorough, was worrisomely monolithic and tended to obscure partial functionality. So, this PR simply does the following:
- Adds just the parameter checks to the basic "arrow functions" test
- Adds a separate test, "lexical "this" in arrow functions", which checks that arrows have a bound `this` that can't be re-bound. (Chrome 37 fails this test)

Unlike #168, this does not include the following "pedantic" checks, which, while important for conformance, are less important for the syntax's general day-to-day usage:
- Lexical `arguments`
- No line terminator between parameters and `=>`
- Absence of `prototype` property (throws on `new`), and empty `name` property

In the future, if/when tests can be formally divided into multiple checks, then these separate tests could be combined.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -281,28 +281,27 @@ exports.tests = [
 {
   name: 'arrow functions',
   link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-arrow-function-definitions',
-  exec: function() {
-    try {
-      return !!Function(
-         // 0 parameters
-         'var a = () => 5;'
-        +'var passed = (a() === 5);'
-
-         // 1 parameter, no brackets
-        +'var b = x => x + "foo";'
-        +'passed &= (b("fee fie foe ") === "fee fie foe foo");'
-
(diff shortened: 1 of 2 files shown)


webbedspace:
I've changed my mind about having two tests for arrows and lexical this.
Now they're back in the same test, as per the former arrow test pullreq. (In this case, Chrome 37+ is now failing the test, but the V8 team [openly admits](http://code.google.com/p/v8/issues/detail?id=2700) that their arrows aren't production-ready yet.)

However, the "pedantic" checks are still absent. Firefox's non-conformance regarding these minutiae is now noted in footnotes.

I've also converted the test code to a comment string, as per the others.

jerone:
:+1: on merging. 

The data on the site isn't correct and took me a while to find this information, which corresponds with my tests.

PR #191, opened 2014-08-09
Updated the default parameters test

webbedspace:
Added the following checks:
- Explicitly passing undefined defers to the default
- Later defaults may reference values of earlier parameters

This doesn't add the following "pedantic" checks:
- Earlier defaults may not reference values of later parameters
- Defaults may use the arguments object.
- Defaults affect the function length property.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -430,11 +430,18 @@ exports.tests = [
   }
 },
 {
-  name: 'default function params',
-  link: 'http://wiki.ecmascript.org/doku.php?id=harmony:parameter_default_values',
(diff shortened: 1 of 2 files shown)


webbedspace:
I also feel like a check for the [defaults' scope contour](http://dmitrysoshnikov.com/ecmascript/es6-notes-default-values-of-parameters/) may be too pedantic to add to this main test as well.

These pedantic checks sadly don't have enough in common, nor enough expressive "power" in themselves, to justify inclusion in a separate 'extra' test (like destructuring currently has with defaults and rest)...

Issue #131, opened 2014-06-04
let/const temporal dead zone


rwaldron:
One of the important things that sets ES6 let/const lexical bindings apart from some existing implementations is that those bindings _cannot_ be accessed in a way that calls their internal [[Get]] at anytime prior to declaration, eg:

``` js
function f() {
  console.log(a); // 2. throws, because `a` isn't declared yet.
}
f(); // <-- 1. calls
let a = 1; // never reached because #2 threw an exception
```

This is the same for `const` bindings.

rwaldron:
Any implementation missing these semantics is **wrong** and should be updated to a nasty red "No". As far as I know, only IE11 does this right.

webbedspace:
FWIW the dead zone test correctly throws in Chrome 35 (but not Firefox <= 30).

(I also feel like incorrect or partial implementations should have a slightly different colour than the red shade used for "missing", but that's for another issue.)

rwaldron:
@webbedspace sure, but it's still broken: 

<img src="http://i.gyazo.com/a5c17a7129d34bc9ad0f857d00990868.png">

`let` and `const` aren't restricted to strict mode (the term "extended mode" appears to refer to "strict mode" + experimental JS)

PR #192, opened 2014-08-09
Tightened the let and const tests, version 2  

webbedspace:
This is an alternative to #164. This simply does the following:
- The `const` test now checks that reassigning to the const is a syntax error (I consider this not a "pedantic" detail but a major aspect of `const`, and even though Firefox has for years made reassigning const a mere silent error, I think such an implementation should be rejected.)
- The `let` test now only contains anti-hoisting checks (for blocks and for-loop heads)
- A new test, "let/const temporal dead zone", contains the TDZ checks for both.
- A new test, "let (for-loop iteration scope)", checks that for-let creates new bindings on each iteration.

This does not include the following "pedantic" checks:
- `const x;` is a syntax error
- `let x; let x;` is a syntax error
- `for(const x = 1; ...)` is not hoisted
- `let` and `const` do not require the `application/javascript;version=1.8` tag attribute
- `let` and `const` do not require strict mode

Those last two are quite important, but, as has been dwelt upon, they end up masking partial implementations of core functionality (and should probably be treated on the same level as Chrome's #enable-javascript-harmony flag). Until better means of representing partial functionality are available, they must be held off for now.


webbedspace:
By the way, this also serves to close #131.

webbedspace:
OK, I've now added footnotes remarking about the `application/javascript;version=1.8` and strict mode limitations, which should hopefully prevent the results from misinforming people.

PR #195, opened 2014-08-12
Added Konqueror 4.13 to ES5, plus minor changes

webbedspace:
In addition to adding KQ:
- Removed the "For example" footnotes (on the basis that the example test code contained the example, too).
- Removed the define-property-webkit note from WebKit (but not from Safari 5).
- Renamed the Opera and Chrome rows to reflect the fact of Opera 15+ being Chromium-based rather than Opera 12-based.
- Added ES7 hyperlinks to the ES5 and ES6 pages.
- Set "Show obsolete browsers" off by default. The obsolete browsers are all multiple years old now - their relevance has diminished.

### DIFF
--- a/data-es5.js
+++ b/data-es5.js
@@ -60,7 +60,7 @@ exports.browsers = {
   safari51: {
     full: 'Safari 5.1',
     short: 'SF 5.1',
-    obsolete: false
+    obsolete: true
   },
   safari6: {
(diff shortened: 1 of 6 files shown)


kangax:
Good stuff @webbedspace! Thanks for all the cleanups and additions.

PR #196, opened 2014-08-12
Fixed an ES5 strict test that crashes in Chrome

webbedspace:
The test `(function f() { f = 123; })()` actually throws a SyntaxError on evaluation (`SyntaxError: Assignment to constant variable`), not a TypeError on execution, in the current Chrome. I'm really not sure which is fully correct for either the ES5 or ES6 specs. Nevertheless, I've changed that line to be `eval`ed so that the entire script doesn't crash.

Also, the link to the strict mode page on the ES5 test page is fixed.

### DIFF
--- a/data-es5.js
+++ b/data-es5.js
@@ -1647,7 +1647,7 @@ exports.tests = [
 },
 {
   name: 'Strict mode',
-  link: 'http://kangax.github.com/es5-compat-table/strict-mode/',
+  link: '../strict-mode/',
   exec: function () {
     "use strict";
     return !this;
--- a/es5/index.html
+++ b/es5/index.html
@@ -1339,7 +1339,7 @@ test(function () {
             <td class="no ejs">No</td>
           </tr>
           <tr>
-            <td id="Strict_mode"><span><a class="anchor" href="#Strict_mode">&sect;</a><a href="http://kangax.github.com/es5-compat-table/strict-mode/">Strict mode</a></span></td>
+            <td id="Strict_mode"><span><a class="anchor" href="#Strict_mode">&sect;</a><a href="../strict-mode/">Strict mode</a></span></td>
 <script>
 test(function () {
   "use strict";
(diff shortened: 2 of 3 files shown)


kangax:
> The test (function f() { f = 123; })() actually throws a SyntaxError on evaluation (SyntaxError: Assignment to constant variable), not a TypeError on execution,

I'm not familiar with ES6 spec as much as I'm familiar with ES5. In ES5, AFAIK, this is supposed to be exclusively a runtime error.

Specifically, http://es5.github.io/#x16 says that:

_"An implementation shall not treat other kinds of errors as early errors even if the compiler can prove that a construct cannot execute without error under any circumstances. An implementation may issue an early warning in such a case, but it should not report the error until the relevant construct is actually executed."_

except for:
- Any syntax error.
- Attempts to define an ObjectLiteral that has multiple get property assignments with the same name or multiple set property assignments with the same name.
- Attempts to define an ObjectLiteral that has both a data property assignment and a get or set property assignment with the same name.
- Errors in regular expression literals that are not implementation-defined syntax extensions.
- Attempts in strict mode code to define an ObjectLiteral that has multiple data property assignments with the same name.
- The occurrence of a WithStatement in strict mode code.
- The occurrence of an Identifier value appearing more than once within a FormalParameterList of an individual strict mode FunctionDeclaration or FunctionExpression.
  Improper uses of return, break, and continue.
- Attempts to call PutValue on any value for which an early determination can be made that the value is not a Reference (for example, executing the assignment statement 3=4).

This is an assignment to a reference, and while relatively easily statically recognizable as an error, shouldn't be reported early.

Unless I missed something, or something changed in ES6 error reporting, which I didn't look into yet.

webbedspace:
It's kind of weird, because in ES6, reassignment to a `const` _is_ a SyntaxError. I guess that's how this discrepancy occurred - but, of course, function internal names (for lack of a better term) aren't `const`s, despite being comparably read-only.

PR #197, opened 2014-08-13
Added tests for each Proxy handler

webbedspace:
This adds individual tests for the Proxy `get`, `set`, `has`, `deleteProperty`, `getOwnPropertyDescriptor`, `defineProperty`, `getPrototypeOf`, `setPrototypeOf`, `isExtensible`, `preventExtensions`, `enumerate`, `ownKeys`, `apply` and `construct` handlers. Currently Firefox supports most of them, excepting the prototype handlers and the enumerate handler.

EJS support was inferred from the EJS source.

Also updates the Reflect test to match all the Proxy handlers.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -1938,16 +1938,140 @@ ${a + "z"} ${b.toLowerCase()}` === "foo bar\nbaz qux";
   }
 },
 {
-  name: 'Proxy',
-  link: 'http://wiki.ecmascript.org/doku.php?id=harmony:direct_proxies',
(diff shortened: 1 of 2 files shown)


webbedspace:
Query: would it be more helpful if this was split into multiple pullrex? It may be a tad monolithic.

webbedspace:
I re-did the entire branch of this pullreq because it kept conflicting easily due to those extraneous Firefox footnotes, which are now moved to #262. Anyway PTAL as usual.

PR #198, opened 2014-08-14
ES6: Added super test

webbedspace:
This checks that:
- `super()` in the constructor binds to the superclass's constructor
- `super.X()` in the constructor binds to the superclass's X method
- `super()` in the X method binds to the superclass's X method
- `super` is bound statically, and remains bound even when a method is extracted

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -752,6 +752,84 @@ exports.tests = [
   }
 },
 {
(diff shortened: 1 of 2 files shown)


arv:
FWIW, there are some ideas on how to improve `super` so that we no longer need `@@create`.

Issue #175, opened 2014-08-04
"Global symbol registry is wrong" is wrong


webbedspace:
#160 did not actually fix the global symbol registry test - it merely updated the compiled HTML file instead of `data-es6.js`.

(This may make you wonder why the compiled HTML file is even in version control anyway. Maybe it should be `.gitignore`d from now on?)

kangax:
My bad; can't believe I missed it. The problem with putting html in .gitignore is that we still need it to be in the repo and updated, so AFAICS can't exclude it.

One option would be to include data files dynamically via JS and build HTML accordingly. That way we could avoid entire compilation process, in exchange for (unfavorable but maybe ok?) dependence on JS.

PR #201, opened 2014-08-15
Fix unscopables test

arv:
The unscopables object is now speced as an Object and not an Array.

This now correctly passes in Chrome 38.

### DIFF
--- a/es6/index.html
+++ b/es6/index.html
@@ -2912,7 +2912,7 @@ test(function () {
 test(function () {
   if (typeof Symbol === "function" && typeof Symbol.unscopables === "symbol") {
     var a = { foo: 1, bar: 2 };
-    a[Symbol.unscopables] = ["bar"];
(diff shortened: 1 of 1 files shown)


webbedspace:
FWIW [table 6.1.5.1](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-well-known-symbols) still (incorrectly) calls the unscopables object "an Array of string values" (in both the PDF spec as well as that HTML version).

webbedspace:
OK, I just had a closer look at this, and it turns out this commit (and #203 and #202) didn't actually do anything because you edited the `es6/index.html` file, which is auto-generated by `build.js`, instead of the `data-es6.js` file which provides the data to it. Whoops! (Perhaps [the README](https://github.com/kangax/compat-table/blob/gh-pages/README.md) isn't quite emphatic enough about this point. See also #175.)

PR #202, opened 2014-08-15
Fix invalid Symbol.iterator test

arv:
The test was using Function() which implies global scope. Since
the test tries to access the local variable b there was a
ReferenceError.

Changed the test to use eval instead.

### DIFF
--- a/es6/index.html
+++ b/es6/index.html
@@ -2758,7 +2758,9 @@ test(function () {
         }
       };
     };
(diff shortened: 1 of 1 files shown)


arv:
Test now passes in Chrome 38

webbedspace:
Whoops, sorry.

By the way, the test currently uses for..of to test the iterator's functionality, but I wonder if it should also attempt to use spread array (...) to check, in case for..of isn't available (e.g. Firefox).

Issue #190, opened 2014-08-09
Symbol.for and getElementsByClassName break IE8 :P


webbedspace:
Ever since the "sort by most implemented" feature was added, which used `getElementsByClassName`, IE 8 hasn't been able to display the pages. Also, `Symbol.for` isn't valid ES3.

I suppose on some level it's ironic that the ES5 page no longer loads on one of the browsers it bears results for, but I was wondering whether you are fussed all that much by it. (Note that fixing the `Symbol.for` test would involve changing the displayed test code to `Symbol['for']`, which may give readers the wrong impression about how it's meant to be used in ES6 - or, bunging the entire check inside an `eval`, which would hinder readability a tad.)

kangax:
Well, considering that tables are entirely for developers who're using modern browsers, this is hardly critical... but we could still fix it, considering it's not that complicated.

Speaking about eval and `Symbol.for`, it gives me an idea to display test snippets not like this:

``` js
test(function () {
  try {
    return eval('export var foo = 1;');
  } catch (e) {
    return false;
  }
}());
```

but like this:

``` js
export var foo = 1;
```

try/catch + eval + wrapping test function are all sort of irrelevant to the test in question anyway (they do provide some clarity on the way you'd need to write things to test for features yourself but I'm not sure it's that important).

This way, we could wrap _all_ tests in eval + try/catch for all we care, and only output eval'd string.

What do you think?

webbedspace:
I like this idea, and would like to extend it to say that the tests themselves, stored in the `data-es6.js` file(s), be stored as strings of ES6 code, with the wrapper added during build. That would allow e.g. the Node test script to be simplified, and a command-line Traceur script to be built (I don't know how Traceur test results are currently gathered), along with scripts for other transpilers like [es6-transpiler](https://github.com/termi/es6-transpiler).

(Now, it may be annoying to write all the code in multi-line strings, necessitating escaping line breaks and certain quote marks, so I guess [this hack](http://tomasz.janczuk.org/2013/05/multi-line-strings-in-javascript-and.html) could be employed in place of string literals).

kangax:
> (Now, it may be annoying to write all the code in multi-line strings, necessitating escaping line breaks and certain quote marks, so I guess this hack could be employed in place of string literals).

Yep, we could parse from comments. It's all one environment (Node) so should be fine.

webbedspace:
Might as well close this now, owing to #209 being merged.

PR #203, opened 2014-08-15
Fix Array.prototype[Symbol.unscopables] test

arv:
The unscobables of the Array.prototype is no longer an Array.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -4156,10 +4156,15 @@ exports.tests = [
   name: 'Array.prototype[Symbol.unscopables]',
   link: 'http://people.mozilla.org/~jorendorff/es6-draft.html#sec-array.prototype-@@unscopables',
   exec: function () {
-    return typeof Symbol === "function" && typeof Symbol.unscopables === "symbol"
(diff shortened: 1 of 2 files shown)


arv:
Test now passes in Chrome 38

webbedspace:
The client-side tests are actually supposed to be ES3-compliant outside of the tested feature (at the very least because the ES5 page is expected to run on not-fully-compliant browsers) so Array#every is (currently) frowned upon. Yes, I know the likelihood of a platform implementing Symbol.iterator but not the Array extras is remarkably low, but coupling these two unrelated features is still a bit improper. See #190.

(This is also why plain for-i loops and `instanceof Array` are used in several other tests instead of isArray and forEach).

webbedspace:
(Astute viewers may have noticed that contrary to the above pronouncement, a previous test I contributed for Symbols does use several ES5 methods without checking for existence. I wish to make amends with #206.)

arv:
Changed it to a plain ol for loop.

PTAL

arv:
Updated data-es6.js and ran `node build.js`.

PTAL

kangax:
Thanks!

PR #204, opened 2014-08-16
Fixed the TCO test in buggy browsers

### LINKED ISSUE #199: Konqueror supports proper tail calls???
Actually, it doesn't - but the test shows "Yes" when I run it in Browsershots. The reason seems to be that when KQ evaluates something like `var x = (function f(n){ if (n <= 0) return true; return f(n - 1);}(1e6));`, it runs out of call-stack and creates a RangeError object... but _returns_ it instead of throwing it. That is, the RangeError object is silently assigned to `x`, and execution continues on its merry way.

In the case of the test, since RangeError objects are truthy, the test shows "Yes". Very funny :3

webbedspace:
Specifically, browsers that don't throw errors when the call stack is exceeded. Fixes #199.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -214,7 +214,12 @@ exports.tests = [
   exec: function() {
     "use strict";
     try {
-      return (function f(n){ if (n <= 0) return true; return f(n - 1);}(1e6));
+      return (function f(n){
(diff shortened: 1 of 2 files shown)


kangax:
Thanks!

PR #205, opened 2014-08-16
ES6: Improved the template strings test

webbedspace:
This splits the template strings test into two tests: one that checks basic expression interpolation (e.g. `${expression}` syntax) and one that performs various tagged template function checks (including checks for the "call site array" argument, its standard and raw properties, and the expression value arguments).

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -1334,12 +1334,74 @@ exports.tests = [
   }
 },
 {
-  name: 'Template Strings',
-  link: 'http://wiki.ecmascript.org/doku.php?id=harmony:quasis',
+  name: 'template strings',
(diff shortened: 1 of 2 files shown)


webbedspace:
Minor vacuous aside: You may notice something a little awkward about the test line that goes `return eval('`foo bar\n${a + "z"} ${b.toLowerCase()}`') === "foo bar\nbaz qux";` - the unescaped "\n" in the eval line tests that the template string syntax can work multi-line, by inserting a line break in the eval string argument containing it. However, it is easy for an idle reader to miss this detail, and presume that since the `foo bar\n` on the left matches the `foo bar\n` on the right, the line break's inclusion is arbitrary. Sadly, there isn't really any way to make this read clearer without using the template string syntax itself...

kangax:
@webbedspace Pretty cool, thanks!

PR #206, opened 2014-08-16
Updated the Symbol test to also check Object.defineProperty

webbedspace:
Based on a musing I had [here](https://github.com/kangax/compat-table/pull/163#issuecomment-49293344): this updates the Symbol test to check that:
- Symbols can be used with the `in` operator (`symbol in object`)
- Symbols can be used as keys in `Object.defineProperty` arguments, not just in bare assignments
  This doesn't change any of the test's historical results, though.

This incidentally also alters the Symbol test and the `Object#__proto__` test to not rely on most ES5 `Object` static methods without checking for existence first (but `Object#__proto__` still requires `Object.getOwnPropertyDescriptor` for basic checking).

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -3372,13 +3372,17 @@ ${a + "z"} ${b.toLowerCase()}` === "foo bar\nbaz qux";
   exec: function() {
     try {
       var object = {};
-      var symbol = Symbol();
-      var value = Math.random();
-      object[symbol] = value;
-      return typeof symbol === "symbol" &&
-             object[symbol] === value &&
-             Object.keys(object).length === 0 &&
-             Object.getOwnPropertyNames(object).length === 0;
+      var symbols = [Symbol(), Symbol()]
+      var values = [Math.random(), Math.random()]
+      
+      object[symbols[0]] = values[0];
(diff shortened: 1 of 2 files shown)


webbedspace:
As you may notice from that second paragraph and a glance at the code, this changes the test so that a platform, e.g.:
- Passes if it has "basic" Symbol functionality (assignment, etc). but lacks `Object.keys`
- Fails if it has basic functionality yet reveals the symbol in `Object.keys`
- Passes if it has basic functionality and doesn't reveal it in `Object.keys`

The first point may seem counterintuitive, but my logic is as follows: the Symbol/Object method interconnection is primarily about hiding\* data. If `Object.keys`/`getOwnPropertyNames` is absent, then the symbols can't be revealed through those methods anyway, so it's comparable in behavior to the third bullet-point in terms of functionality offered to the coder. Thus, only when those methods are present and offer a means of revealing the Symbols should the functionality be considered incomplete.

*I of course don't mean Symbols are "hidden" in terms of security, but in terms of reliability - adding Symbol properties to objects won't affect ES5/ES3 code that uses said objects.

PR #209, opened 2014-08-18
ES6 test functions in multi-line strings

webbedspace:
This updates the build script, allowing it to interpret ES6 code expressed in function comments instead of concatenated strings (see #190 iscussion) - which is then wrapped in try-catch, eval() and test() structures automatically.

Also, the runtime script has been updated to unwrap those same structures for the tooltip preview, making the resulting display cleaner - lacking even the test() call wrapper.

Only some ES6 and non-standard tests have been converted to make use of this format, for the moment. The remaining tests are the subjects of various pull-requests currently. If those other requests are resolved, conversion of those tests may then be done.

NOTE: as this drastically affects the output of all HTML files, they have not been included in this commit, to allow branch-merging to pass smoothly. A new build commit will thus have to be done after the merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 ECMAScript 5/6/7/non-standard compatibility tables
+==================================================
 
 [![Bitdeli Badge](https://d2weczhvl823v0.cloudfront.net/kangax/es5-compat-table/trend.png)](https://bitdeli.com/free "Bitdeli Badge")
 
@@ -6,3 +7,5 @@ Editing the tests
 -----------------
 
 Edit the `data-es5.js`, `data-es6.js`, `data-es7.js`, or `data-non-standard.js` files to adjust the tests and their recorded browser results. Run `node build.js` to build the HTML files from these JavaScript sources.
+
+The tests themselves should be written in pure ES3, *except* for the sole ES6 feature being tested (as well as any ES5 features strictly required to use the ES6 feature). The test code is placed in multi-line comments (as in [this hack](http://tomasz.janczuk.org/2013/05/multi-line-strings-in-javascript-and.html)), so that node can parse the data scripts without throwing syntax errors. The `build.js` script will wrap the code in an `eval` call inside a `try`, so the tests themselves do not need to catch errors that non-supporting platforms may throw.
--- a/data-es6.js
+++ b/data-es6.js
@@ -493,7 +493,7 @@ exports.tests = [
 },
 {
   name: 'rest parameters',
-  link: 'http://wiki.ecmascript.org/doku.php?id=harmony:rest_parameters',
+  link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-function-definitions',
   exec: function() {/*
     return (function (...args) { return typeof args !== "undefined"; }())
   */},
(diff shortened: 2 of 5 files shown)


data-es6.js
```diff
@@ -494,13 +490,9 @@ exports.tests = [
...
   link: 'http://wiki.ecmascript.org/doku.php?id=harmony:rest_parameters',
-  exec: function () {
-    try {
-      return eval('(function (...args) { return typeof args !== "undefined"; }())');
-    } catch (e) {
-      return false;
-    }
-  },
+  exec: function() {/*
+    return (function (...args) { return typeof args !== "undefined"; }())
```
kangax, line 498: Ahh, so refreshing to see this. Feels just like writing ES6 on a spot :)

master.js
```diff
@@ -56,10 +56,21 @@ domready(function() {
...
+    
     infoEl.onmouseover = function(e) {
       mouseoverTimeout = null;
+	  
       var scriptEl = this.parentNode.parentNode.getElementsByTagName('script')[0];
-      infoTooltip.innerHTML = scriptEl.innerHTML.replace(/^\n/, '').replace(/</g, '&lt;');
+      var id = "tooltip_" + this.parentNode.id;
+	  
+      infoTooltip.innerHTML = tooltipCache[id] = tooltipCache[id] || scriptEl.innerHTML
+        // unwrap "multi-line comment" code
```
kangax: Perhaps we can just save original ES6 code in a data attribute (or hidden tag, etc.) during build, not to bother with deserialization again when showing tooltip? It would be simpler and less fragile.

kangax: We would literally just have to use something like `scriptEl.getAttribute('data-source')` instead of `scriptEl.innerHTML` followed by all the parsing.

webbedspace: OK, done.

README.md
```diff
@@ -6,3 +7,5 @@ Editing the tests
 -----------------
 
 Edit the `data-es5.js`, `data-es6.js`, `data-es7.js`, or `data-non-standard.js` files to adjust the tests and their recorded browser results. Run `node build.js` to build the HTML files from these JavaScript sources.
+
+The tests themselves should be written in pure ES3, *except* for the sole ES6 feature being tested (as well as any ES5 features strictly required to use the ES6 feature). The test code is placed in multi-line comments (as in [this hack](http://tomasz.janczuk.org/2013/05/multi-line-strings-in-javascript-and.html)), so that node can parse the data scripts without throwing syntax errors. The `build.js` script will wrap the code in an `eval` call inside a `try`, so the tests themselves do not need to catch errors that non-supporting platforms may throw.
```
kangax, line 11: Good idea, thanks!

webbedspace:
Just a final note: as I mentioned, this commit didn't update the HTML files because they would create merge conflicts with the other six PRs. So, these changes hasn't actually gone "live" yet. 

You could make a new commit now that updates the HTML files (thus necessitating that I rebase out the resulting conflicts myself), or you could defer that until you've reviewed more PRs.

kangax:
I'll take a look at other PR's in the next few days, and I suppose I'll wait with build for now, to avoid conflicts on your side.

PR #212, opened 2014-08-19
Corrected #201 and #202

webbedspace:
I just noticed that pulls #201 and #202 actually edited the HTML file instead of the `data-es6.js` file. This fixes that.


kangax:
Thanks @webbedspace! Can't believe I keep missing them :/

Issue #151, opened 2014-06-27
Non-standard: add more SpiderMonkey extensions


webbedspace:
- (`Date#toLocaleFormat`)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleFormat]
- (Function `displayName` property)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/displayName] (What it actually does, when set, is make an anonymous function have a different name than `"(anonymous function)"` on the dev tools' call stack.
- (Error `columnNumber` property)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/columnNumber]
- (Object `__noSuchMethod__` property)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/noSuchMethod] ...(Despite the page name, `"__noSuchMethod__" in Object.prototype` is false in Firefox.)
- (InternalError)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/InternalError]
- (Object `__lookupGetter__` property)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/lookupGetter]

kangax:
@webbedspace thanks for all the suggestions and bug reports! If you can create PR's for all or any of them, that would be amazing. I don't have much time right now between work and all the OS projects, so not sure when I'll be able to get to this.

webbedspace:
OK, I've given it a shot. #155

webbedspace:
Update: `displayName` isn't possible to test for inside page Javascript (given that it only affects console output) so that's sadly out. Also I didn't notice `__noSuchMethod__` was already present.

Other things that this table could do with:
- `for each` loops
- `Iterator` constructor and the `__iterator__` [property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators)
- JS 1.8-style generators (e.g. `yield` in bare functions without the `*`)
- [JS 1.8-style generator comprehensions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators#Generator_expressions)
- [String generic methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#String_generic_methods)

[...]

PR #218, opened 2014-08-21
Added more non-standard tests

webbedspace:
- `Date.prototype.toLocaleFormat`
- error `columnNumber`
- `__lookupGetter__`
- String generics
- `for each...in`
- Iterator
- `__iterator__`
- Generators (JS 1.8)
- Array comprehensions
- Generator comprehensions

### DIFF
--- a/data-non-standard.js
+++ b/data-non-standard.js
@@ -663,6 +663,49 @@ exports.tests = [
     besen: false,
     rhino: true,
     phantom: false
+  },
+  separator: 'after'
+},
+{
+  name: 'Array comprehensions (right-to-left)',
+  link: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Predefined_Core_Objects#Array_comprehensions',
+  exec: function () {
+    try {
+      var a = eval('[i * 2 for (i in { 2: true, "foo": true, 4: true }) if (i !== "foo")]');
+      return a instanceof Array && a[0] === 4 && a[1] === 8;
+    }
+    catch(e) {
+      return false;
+    }
+  },
+  res: {
+    ie7: false,
+    ie8: false,
+    ie9: false,
+    ie10: false,
+    ie11: false,
+    firefox3: true,
+    firefox3_5: true,
+    firefox4: true,
+    firefox5: true,
+    firefox6: true,
+    firefox7: true,
+    firefox12: true,
+    firefox28: true,
+    safari3: false,
+    safari4: false,
+    safari5: false,
+    safari7: false,
+    webkit: false,
+    chrome7: false,
+    opera10_10: false,
+    opera10_50: false,
+    opera15: false,
+    konq44: false,
+    konq49: false,
+    besen: false,
+    rhino: false,
+    phantom: false
   }
 },
 {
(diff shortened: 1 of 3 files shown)


data-non-standard.js
```diff
@@ -182,7 +182,15 @@ exports.tests = [
 {
   name: '"toSource" method',
   exec: function () {
-    return 'toSource' in (function (){}) && 'toSource' in ({});
+    return Object   .prototype.hasOwnProperty('toSource')
```
kangax: Is there a reason for such precision? I think it's unnecessary.

data-non-standard.js
```diff
@@ -546,13 +554,51 @@ exports.tests = [
...
+    rhino: true,
+    phantom: true
+  },
+  separator: 'after',
 },
 {
   name: 'Array generics',
   exec: function () {
-    return typeof Array.slice === 'function' && Array.slice('123').length === 3;
+    return typeof Array.slice === 'function' && Array.slice('abc').length === 3;
```
kangax, line 603: This is to avoid some sort of false positives?

data-non-standard.js
```diff
@@ -585,6 +631,84 @@ exports.tests = [
   }
 },
 {
+  name: 'String generics',
+  exec: function () {
+    return typeof String.slice === 'function' && String.slice(123, 1) === "23";
```
kangax, line 638: There's also String generics? Nice.

data-non-standard.js
```diff
@@ -663,6 +787,47 @@ exports.tests = [
   }
 },
 {
+  name: '"for each..in" loops',
```
kangax, line 793: Let's add a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for_each...in ?

data-non-standard.js
```diff
@@ -704,6 +869,215 @@ exports.tests = [
...
+    phantom: false
+  }
+},
+{
+  name: 'Generators (JS 1.8)',
+  link: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators#Generators',
+  exec:[{
+      type: 'application/javascript;version=1.8',
+      script: function () {
+	  console.log(JSON.stringify(__script_executed));
```
kangax: This is not needed, right?

data-non-standard.js
```diff
@@ -704,6 +869,215 @@ exports.tests = [
...
+          }
+          catch(e) {
+            return false;
+          }
+        }());
+		__script_executed["generators"] = true;
+      }
+  },{
+      script: function () {		
+	  console.log(JSON.stringify(__script_executed));
```
kangax: Also here

data-non-standard.js
```diff
@@ -1033,6 +1407,41 @@ exports.tests = [
   separator: 'after'
 },
 {
+  name: 'Date.prototype.toLocaleFormat',
+  link: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleFormat',
+  exec: function () { return typeof Date.prototype.toLocaleFormat == 'function' },
```
kangax: We should probably use `===` for consistency but not a big deal

data-non-standard.js
```diff
@@ -1203,6 +1612,41 @@ exports.tests = [
   }
 },
 {
+  name: 'error "columnNumber"',
```
kangax, line 1625: Let's add a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/columnNumber

data-non-standard.js
```diff
@@ -1203,6 +1612,41 @@ exports.tests = [
...
+    ie9: false,
+    ie10: false,
+    ie11: false,
+    firefox3: false,
+    firefox3_5: false,
+    firefox4: false,
+    firefox5: false,
+    firefox6: false,
+    firefox7: false,
+    firefox12: false,
```
kangax, line 1642: Hm, I'm surprised it's false in FF12 and below.

kangax, line 1642: But you're right. Looks like introduced in FF17

data-non-standard.js
webbedspace, line 603: Mostly it's to make it 1% more readable (the first time I misread '123' as a number literal, interpreting it as some kind of nonsensical slice index).

data-non-standard.js
webbedspace: Well, I wanted to demonstrate that all of the prototypes possess their own unique 'toSource' methods. If I just ran `"toSource" in Function.prototype`, then that would return `true` if only Object.prototype possessed the toSource method that they all delegated to.

But, you're right - I guess since there's no standard for these, I should be permissive rather than rigorous for these tests. Need to get in the right mindframe.

data-non-standard.js
webbedspace: Whoops, sorry.

data-non-standard.js
webbedspace: (Oh, well, if you mean "just that one line" rather than the entire set of checks in the test, then, well, OK, mostly I wrote it like that for visual consistency with the lines below - but it does seem leery.)

webbedspace:
Hope that commit is all good.

BTW, forgot to mention that this closes #151.

kangax:
I would still want to remove those extra `toSource` checks, but it's fine for now. Thanks for doing all this!

PR #221, opened 2014-08-22
Non-standard: merge IE7-10 and FF5-6's columns

webbedspace:
Since IE 7-10 have identical results, their columns are now merged. The same applies to Firefox 5 and 6. Of course, if a new test is devised which does have different results between these versions, they'll need to be split again, but I feel that's unlikely.


kangax:
Awesome, thanks @webbedspace!

PR #233, opened 2014-08-27
Made some displayed tooltip code for certain tests appear cleaner

webbedspace:
Tests which aren't in comment strings now have their code stripped of the "function () {" wrapper when included in a tooltip, putting them on par with the others.

(For merge conflict avoidance, this commit doesn't update the HTML files.)

### DIFF
--- a/build.js
+++ b/build.js
@@ -213,7 +213,7 @@ function testScript(fn) {
     // if there wasn't an expression, make the function statement into one
     if (!expr) {
       expr = deindentFunc(fn);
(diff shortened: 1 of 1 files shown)


kangax:
Perhaps we should just update all tests to be in comments?

webbedspace:
I guess - I've been leaving untouched the ones that are just basic built-in-function existence checks (such as `return typeof Object.create === "function"`), but not for any reason other than if-it-aint-broke. Obviously the presence of this pullreq belies that notion, though, so I find myself in agreement.

webbedspace:
(But, I shan't get around to it until the other 4 test-modifying PRs meet some resolution.)

PR #235, opened 2014-08-28
Fixes ES5 and non-standard pages in IE 8

webbedspace:
The ES5 and non-standard pages should now work in IE 8 again.
(The ES6 page, puzzlingly, seems to make it freeze or crash. Oh well, maybe later.)

Fixes include:
- Fallback for hsl() colours. (I prefer the hsl format for clarity of reading and understanding - the relation between the "hover" and normal colours is made obvious in HSL, so I opt to leave both colour formats in the CSS.)
- "Show obsolete browsers" checkbox works again
- Fallback for opacity when selecting a column
- Tooltips work again (incorrectly uses String#trim and Array#forEach)

### DIFF
--- a/data-non-standard.js
+++ b/data-non-standard.js
@@ -497,11 +497,16 @@ exports.tests = [
 {
   name: '__lookupGetter__',
   link: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__lookupGetter__',
-  exec: function () {
-    var obj = { get foo() { return "bar"; } };
+  exec: function () {/*
+    try {
+      var obj = eval('{ get foo() { return "bar"; } }');
+    }
+    catch(e) {
+      obj = {}; obj.__defineGetter__("foo", function(){ return "bar"; });
+    }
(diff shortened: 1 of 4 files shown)


data-non-standard.js
```diff
@@ -497,11 +497,16 @@ exports.tests = [
 {
   name: '__lookupGetter__',
   link: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__lookupGetter__',
-  exec: function () {
-    var obj = { get foo() { return "bar"; } };
+  exec: function () {/*
+    try {
```
kangax, line 501: Hm, why the need for `try` if we're already wrapping?

webbedspace, line 501: Notice the `catch` - the purpose of this is to default to `__defineGetter__` if the ES5 getter syntax isn't available. The implicit catch added by build.js (which will be used if _that_ catch block throws) simply returns false, signifying test failure.

kangax, line 501: Ahha, gotcha. Makes sense. Thanks.

Issue #239, opened 2014-09-01
Chrome 37 results not accurate


webbedspace:
A number of features present in Chrome 37 betas have not actually been made available in Chrome stable 37:
- arrows (don't actually change this, just pass #189)
- `String.fromCodePoint`
- `String.prototype.codePointAt`
- for..of

webbedspace:
Also, two quick addenda about Firefox:
- FF 33 has global symbols
- FF 34 has destructuring rest

kangax:
@webbedspace thanks! I noticed FF34 addition today as well (pushed). Will check the rest later.

webbedspace:
Some further discrepancies (done with Opera 24):
- Does not support "Symbol : cannot coerce to string or number" (neither does Opera 23)
- "let : for-loop statement scope (strict mode)" causes the tab to crash (only Opera 24)
- _Does_ support "Symbol.iterator" (but Opera 23 does not)

Cross-reference: #227

Issue #227, opened 2014-08-25
ES6: Add a Chrome 36 (Opera 23) column


webbedspace:
Chrome 36 is identical to Chrome 35 but adds `Array.prototype.fill`.

(I believe the rule for including a new version column is whether just one result differs? Obviously several versions have been omitted in the past, but that doesn't prevent more consistent inclusions from now on.)

kangax:
Yep, if one differs we add a new one.

PR #241, opened 2014-09-07
Differential results

### LINKED ISSUE #224: Record results data differentially
Imagine if the `Array.prototype.values` results were actually recorded like this:

```
{
    tr:          true,
    ejs:        true,
    ie:          false,
    firefox: {
      17:   {
        val: false,
        note_id: 'fx-array-prototype-values',
        note_html: 'Available from Firefox 17 up to 27 as the non-standard <code>Array.prototype.iterator</code>'
      },
      27:   {
        val: false,
        note_id: 'fx-array-prototype-values-2',
        note_html: 'Available since Firefox 27 as the non-standard <code>Array.prototype["@@iterator"]</code>'
      },
    },
    chrome: {
      30: true,
    }
    safari:      false,
    webkit:      false,
    opera:       false,
    konq49:      false,
    rhino17:     false,
    phantom:     false,
    node: {
      harmony: true
    }
}
```

Instead of having a line of `false`s and `true`s, support for a particular sequence of versions could be shown by just assigning `true` to a particular version number that the table has a column for (and then `false` to a further version if the platform drops support).

webbedspace:
Now, if a result from a platform's version is not different from that of its previous version, the result need not be recorded in the data files. Instead, build.js will copy the result over automatically.

What this basically means is, all you usually need to specify for particular platforms is the result for the _earliest version_, plus the result for _the version where support became true_ (if any). For instance, here's the `Number.EPSILON` results:

```
    tr:          true,
    ejs:         true,
    ie10:        false,
    firefox11:   false,
    firefox25:   true,
    chrome:      false,
    chrome34:    true,
    safari51:    false,
    webkit:      false,
    opera:       false,
    konq49:      false,
    rhino17:     false,
    phantom:     false,
    node:        false,
    nodeharmony: true
```

If a certain range of versions needs a note attached, that can also be specified (see these results for `__proto__`):

```
    tr:          true,
    ejs:         false,
    ie10:        false,
    ie11:        true,
    firefox11:   true,
    firefox33:   {
      val: true,
      note_id: 'fx-proto-shorthand',
      note_html: 'Firefox 33 and 34 incorrectly regard both of the following as true: <ul>'
      +'<li><code>var __proto__ = []; ({ __proto__ }) instanceof Array</code>'
      +'<li><code>({ __proto__(){} }) instanceof Function</code>'
      +'</ul>'
    },
    firefox34:   { val: true, note_id: 'fx-proto-shorthand' },
    firefox35:   true,
    chrome:      true,
    safari51:    true,
    webkit:      true,
    opera:       true,
    konq49:      true,
    rhino17:     true,
    phantom:     true,
    node:        true,
    nodeharmony: true
```

I've thus cleared out all unnecessary lines from data-es6.js. This has resulted in a large diff, but you may observe the es6/index.html diff to see that little has changed w/r/t the output results.

Note: browsers are determined to be the same version by comparing their "full" string. As a result, certain older Chrome versions now have the "full" string "Chrome, Opera", even though they predate Opera 15.

ALSO:
- Fixed a few Chrome and Firefox results as per #239 (and closing it).
- Fixed some older Firefox results that referenced "firefox28" instead
  of "firefox27" (which comprises both 27 and 28).

This closes #224.


webbedspace:
You may notice this isn't quite the same as some of the ideas floated in #224. Mainly I've chosen this approach because it would cause the least amount of conflicts with the other pull requests. When those are done, maybe some more semantic changes could be made (like changing `firefox11` to `firefox11+` or something). As it is, the main goal (of vastly lowering the redundancy of `data-es6.js`) is accomplished.

webbedspace:
(A supplementary goal of this is to lay the groundwork for a succinct fix to #227, which would need only a few lines' insertion rather than one per test. Also, I'd like to "fix" the absence of FF26 and Chrome 28, which currently have a footnote each denoting their sole difference to the previous version.)

webbedspace:
Rebased. (This changes the "full" name of Chrome 39 from "Chrome" to "Chrome, Opera", which is needed for build.js to tell that it's the continuation of the previous "Chrome, Opera" column.)

webbedspace:
@kangax You've already mentioned your reservations with #258, but I'm wondering what your opinion of just this one pullreq (which doesn't add or change any tests) is, and whether I should keep updating it or re-work it.

PR #244, opened 2014-09-10
Fixed some JS bugs on the non-standard page

webbedspace:
- Fixed "Show obsolete browsers" checkbox.
- Fixed row-column highlighting when hovering over cells.
- Removed message reading "I hope to test conformance sometime in the future." :P

### DIFF
--- a/non-standard/index.html
+++ b/non-standard/index.html
@@ -3,9 +3,8 @@
   <head>
     <meta charset="utf-8">
     <title>ECMAScript extensions compatibility table</title>
-
-    <link rel="stylesheet" href="../master.css">
-
+    <link href='http://fonts.googleapis.com/css?family=Open+Sans:300,600' rel='stylesheet' type='text/css'>
+    <link rel="stylesheet" href="../master.css" type="text/css">
     <script src="../ready.js"></script>
     <script src="../master.js"></script>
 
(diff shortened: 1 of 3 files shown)


non-standard/index.html
```diff
@@ -3,9 +3,8 @@
   <head>
     <meta charset="utf-8">
     <title>ECMAScript extensions compatibility table</title>
-
-    <link rel="stylesheet" href="../master.css">
-
+    <link href='http://fonts.googleapis.com/css?family=Open+Sans:300,600' rel='stylesheet' type='text/css'>
```
kangax, line 6: Minor quibble: we use double quotes for attributes everywhere else :)

non-standard/index.html
```diff
@@ -52,10 +51,10 @@
     </div>
     <div id="body">
       <p class="warning">Please note that <i>some of these tests</i> represent <strong>existence</strong>,
-        not functionality or full conformance. I hope to test conformance sometime in the future.</p>
+        not functionality or full conformance.</p>
```
kangax, line 54: Long overdue, thanks!

non-standard/index.html
webbedspace, line 6: Well I copy-pasted that one line verbatim from es6/skeleton.html, so...

PR #248, opened 2014-09-12
Improved Promise tests, and added support for async testing

### LINKED ISSUE #247: ES6: Make the Promise test better (without document.write)
There's something of a problem with testing Promise: because resolving is a deferred action, it's not possible, for a `Promise#then` or `Promise#catch`-checking test, to call `test()` there and then to `document.write` a Yes/No value to the HTML. It'll require some re-working of `master.js`.

webbedspace:
Closes #247. This makes the Promise test call `Promise#then` and `Promise#catch` and checks what behaviour results, and also adds `Promise.all` and `Promise.race` tests.

In order to get the Promise test to work, I needed to implement asynchronous test result writing. To do this, test functions may now, instead of returning true/false, instead at some point call `asyncTestPassed()`, which is a function of no arguments, provided by the testing environment, that records the test as passed regardless of execution order. `build.js` has been updated to provide such a function (which manually updates the class and text of the corresponding `<td>`) as has the `node.js` test runner script.

The historical results have not _yet_ been re-verified (WebKit, Chrome 33 or Firefox 29 might actually fail these new tests, unlikely though it seems).

### DIFF
--- a/build.js
+++ b/build.js
@@ -229,7 +229,7 @@ function dataToHtml(browsers, tests, compiler) {
     body.push(
       '<tr' + (subtests ? ' class="supertest"' : '') + '>',
       '\t<td id="' + id + '"><span><a class="anchor" href="#' + id + '">&sect;</a>' + name + footnoter.get(t) + '</span></td>\n' +
-      testScript(t.exec, compiler)
+      testScript(t.exec, compiler, id)
     );
     
     // Function to print out a single <td> result cell.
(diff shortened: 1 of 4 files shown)


webbedspace:
Update: I ran some tests and can now state with certainty that WebKit, Chrome 33+ and Firefox 29+ pass all three tests.

I also forgot to mention that this does not update the HTML file, for merge conflict avoidance.

kangax:
This is now in, right?

webbedspace:
Actually, I still need to re-work this so that it can asynchronously update results when it's in subtests rather than just top-level tests.

PR #250, opened 2014-09-16
Cleaned up master.css a bit

webbedspace:
In addition to making the CSS somewhat more readable, this also changes some of the colours for the headers in each page, to make each of them visually distinct (previously, only the non-standard page had different header colours).

Also, it changes the "No" colour on the non-standard page from red to blue. The reasoning is that since the features conform to no ES standard, the judgmental, hostile red used for the other pages is less appropriate here, and a more neutral, yet distinct, shade is called for.

Finally, I removed the user-agent tooltip `<script>` from the non-standard table, as it was interfering with the DOM structure expected by some of the CSS, and didn't really contribute much (given that it's absent on the other pages).

### DIFF
--- a/es6/index.html
+++ b/es6/index.html
@@ -67,11 +67,11 @@
       </colgroup>
       <thead>
         <tr>
-          <th colspan="3" class="ignore"></th>
-          <th colspan="3" class="ignore" style="background: #efe"><span>Compilers</span></th>
-          <th colspan="15" class="ignore desktop" style="background: #eef"><span>Desktop browsers</span></th>
-          <th colspan="4" class="ignore" style="background: #fee"><span>Server-ish</span></th>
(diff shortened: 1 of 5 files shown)


kangax:
@webbedspace I like everything and would love to pull this in except one thing — can we change this:

``` css
body
    { font-family: 'Open Sans', sans-serif; margin: 0; font-weight: 300; display: inline-block;}
```

to this:

``` css
body { 
  font-family: 'Open Sans', sans-serif; margin: 0; font-weight: 300; display: inline-block;
}
```

I really don't like (and never used) 1st version :) Thanks!

webbedspace:
OK, done. Although, lately I wonder if the blue shade I've used for the non-standard table's "No" cells is a little garish... WDYT.

Issue #245, opened 2014-09-11
IE does not support Uint8ClampedArray


webbedspace:
Need to add a footnote mentioning this, and change its TypedArray support to "false".

webbedspace:
Also, in the non-standard table, WebKit now supports the Error "stack" own property.

webbedspace:
Also, note to myself: the IE data for hoisted function declaration is incorrect: it actually passed (i.e was compatible with IE 8) until IE11.
(I'll get around to doing a PR for these once #241 is resolved.)

zloirock:
![...](http://i.imgur.com/VU19xH6.png?1)

zloirock:
Looks like it was added in [this update](https://support.microsoft.com/en-en/kb/2929437).

PR #255, opened 2014-09-24
Updated WebKit and IE results

webbedspace:
WebKit now supports the ES6 Number constants and some String methods.

This also closes the entirety of #245.


kangax:
Awesome, thanks!

webbedspace:
Thanks - I just rebased the other pullreqs that are prone to conflicts (189, 192, 197, 241), so PTAL.

PR #256, opened 2014-09-25
More specific function name property tests

rwaldron:


data-es6.js
```diff
@@ -2709,49 +2709,66 @@ ${a + "z"} ${b.toLowerCase()}` === "foo bar\nbaz qux";
...
+    };
+    var descriptor = Object.getOwnPropertyDescriptor(o, "foo");
+
+    return foo.name === 'foo' &&
+      bound.name === 'bound foo' &&
+      o.a.name === 'a' &&
+      o.b.name === 'c' &&
+      descriptor.get.name === 'get foo' &&
+      descriptor.set.name === 'set foo' &&
+      (function() {}).name === '' &&
```
rwaldron, line 2728: I need to confirm this one

webbedspace, line 2728: You refer to [19.2.1.1, step 26a?](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-function-p1-p2-pn-body)

webbedspace:
Something makes me think that this shouldn't replace the present test, but complement it: one tests the basic case where a function's name is in FunctionExpression/FunctionDeclaration _Identifier_ position (which is arguably the more immediately useful case for developers), and this one tests all the cases where the name is implicitly derived from _PropertyName_, _BindingIdentifier_ etc. (with obvious exception to `o.b.name === 'c'`, which should probably be added to the present test, and `(function() {}).name === ''` which is arguably a case where the name is explicitly omitted).

Maybe call the first test `function 'name' property (explicit)` and the other one `function 'name' property (implicit)`.

rwaldron:
This set of checks looks at all aspects of ES6 function name property changes. The problem I have with the original test is that it makes it appear as though this is implemented in Firefox, but it's not.

webbedspace:
The main reason I want the two tests is that:
- The _explicitly_ assigned function name was never specified in ES5, but was a prevalent non-standard extension prior. Hence, it counts as functionality new to ES6 by itself, and appears on the ES6 table instead of the non-standard table.
- The _implicitly_ assigned function name (`var foo = function(){}`, "get foo" etc.) apparently has no substantial precedent on the web and is thus new to both ES6 and to _de facto_ standards.
  This may seem a little arbitrary, but a few other features are split up in this way ("string spreading" for instance) simply because they provide more informative results about the state of implementations.

...Alternatively, you could simply have that one test, and then attach a HTML note to all the browsers that formerly had `true` results:

```
{
  val: false,
  note_id: 'function-name',
  note_html: 'Only name properties for explicitly named functions (such as <code>function foo(){}</code>) are supported.'
}
```

Or whatever succinct wording you think would be best.
(Only one of them would have the `note_html` property - see e.g. the Map results in `data-es6` for an example.)

data-es6.js
rwaldron, line 2728: Sorry, should've been more explicit. I don't need to confirm this one with the spec, I know what the spec says, I presented the proposal as representative of the author. I was sitting in an Ecma/TC39 meetin when I made this PR and wanted to confirm that this was indeed the behavior we wanted to continue forward with. I didn't have a chance to bring this up for discussion

rwaldron:
The original strawman explains that the semantics and machinery being proposed for ES6 include specification of the de facto standard and establishing clear definitions for previously unspecified cases: http://wiki.ecmascript.org/doku.php?id=harmony:function_name_property . As you said, the explicit _Identifier_ -> `.name` behavior wasn't defined in ES5, but is defined in ES6 as one aspect of a bigger picture feature.

webbedspace:
I feel like we're starting to talk in circles... I understand that (for instance) `(function e(){}).name==='e'` and `({a:function(){}}).a.name==='a'` are actually the same feature regarding the same property, but the former form has a drastically different support base (almost every platform except IE) than the other kinds (no platforms at all yet), and I want this difference to be visible on the table in some way - be it either through a separate row, or footnotes.

Maybe I'm just splitting hairs here - I guess the former isn't _that_ important for developers, given that the aforementioned IE lack-of-support precludes its unqualified use.
(I know that adding a CanIUse-style "partial support" value to complement "Yes" and "No" would be the most elegant way of solving this, and has been on the radar for awhile, but I haven't gotten around to doing a pullreq for it yet.)

webbedspace:
@kangax If you've been hesitant to merge this because of my reservations above, I'm thinking now it'd be easiest if this just went through as-is, and I did a separate pullreq instating the desired footnote(s) at my discretion. Unless you have a particular opinion on this discussion?

rwaldron:
Sorry for not following up, but the point I is: existing support is meaningless since there was no specification for the behaviour until ES6. Having support for `(function e(){}).name==='e'` and no other aspects of the feature means that the feature is not implemented correctly to the specification. There was no feature, now there is a feature, it has a spec and compliance with that spec should be tested.

webbedspace:
FWIW I added what I believe was all of the extra checks introduced in this pullreq as subtests in #286, [as you can see.](http://kangax.github.io/compat-table/es6/#function_name_property)

rwaldron:
@webbedspace yes, I saw. 

> existing tests into super- and sub-tests. 

This is key, wish I'd had that previously. Nice work

PR #258, opened 2014-10-02
Six of my other pull requests merged together

### LINKED ISSUE #131: let/const temporal dead zone

webbedspace:
This is a merge of #189, #192, #197, #206, #248 and #241.
As such, it closes #131, #247 and #224.

The main reason for merging these is to apply the benefits of #241 to the tests added by the other pullreqs - the results are shorter and more compact. A number of small conflicts between them are also resolved.


webbedspace:
Rebased; PTAL.

kangax:
Hey @webbedspace, sorry for not attending to this sooner and thanks for all the effort! I don't have enough time to look through everything, unfortunately. As I mentioned earlier, I also wouldn't want to mix small (existence) tests with more sophisticated conformance ones (which consist of multiple small ones). I feel like we need support for partial tests, where we can color cell yellow or smth. and show that it passed 3 out of 5 tests (similar to what happens with things like arrow functions in Chrome, which lacks lexical binding). We'd need to adjust total scores accordingly and obviously make it so that each test could consist of multiple small ones. This would also need proper UI to display small tests, etc. But again... this is quite an effort and I have no idea when I'll have time for all this (I've got a full-time startup where we're at the critical point now; trying to keep Fabric.js afloat; trying to maintain all the other side projects — html minifier, compat tables, node-canvas).

PR #259, opened 2014-10-05
Updated footnote for octal/binary literals

webbedspace:
TC39 [recently decided](http://esdiscuss.org/topic/sept-23-2014-meeting-notes#content-2) to allow `Number("0b1")` to evaluate to 1, thus making Chrome's behaviour correct and Firefox's incorrect.


kangax:
Good to know, thanks!

PR #263, opened 2014-10-16
ES6 Math functionality checks

webbedspace:
Some Math functions' tests have been upgraded from mere existence to basic functionality. These are not supposed to test much beyond a basic result or two, to at least determine that they quack like the intended mathematical function.

I don't have personal familiarity with several of these functions, but my research indicates these are basically valid expressions of their behaviour/relation to other Math functions.

This may seem spurious, but note that Konqueror 4.13's `acosh` is actually implemented to behave like `asinh` due to that bug I mentioned [some months back.](https://github.com/kangax/compat-table/pull/183#issuecomment-51328736)

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -4690,7 +4690,8 @@ ${a + "z"} ${b.toLowerCase()}` === "foo bar\nbaz qux";
   name: 'Number.isFinite',
   link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-isfinite-number',
   exec: function () {
-    return typeof Number.isFinite === 'function';
+    return typeof Number.isFinite === "function"
+	  && Number.isFinite(2) && !Number.isFinite(1/0);
   },
   res: {
     tr:          true,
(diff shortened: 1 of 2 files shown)


Yaffle:
for some of your tests specification do not guarantee accuracy
there was a disscussion on the esdiscuss.org about this, but nothing was changed....
although, we can fight with ES spec authors here in compat-table

data-es6.js
```diff
@@ -4890,7 +4894,7 @@ ${a + "z"} ${b.toLowerCase()}` === "foo bar\nbaz qux";
   name: 'Number.EPSILON',
   link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-number.epsilon',
   exec: function () {
-    return typeof Number.EPSILON === 'number';
+    return Number.EPSILON === 2.2204460492503130808472633361816e-16;
```
Yaffle, line 4897: actually, it is not
`2.2204460492503130808472633361816e-16`,
but
`2.220446049250313080847263336181640625e-16`  (`2 ** -52`)
P.S.
It is very strange, that ES6 spec tells to specify this number with such big number of decimal digits, 
`2.220446049250313e-16` is a shortest decimal string, which will be converted to epsilon value for binary64 double. In case JavaScript will switch to binary128, the epsilon will be `2 ** -112`.

Yaffle:
see http://jsfiddle.net/hv1qqyst/embedded/result/ for Math.xxx results with values from https://github.com/kangax/compat-table/issues/392

PR #264, opened 2014-10-16
Add Closure Compiler column

teppeis:
Google Closure Compiler began to support ES6 transpiration.
https://github.com/google/closure-compiler/wiki/ECMAScript6

I created [test tool](https://github.com/teppeis/closure-compiler-es6-compat-table) and added the result to es6 table.


kangax:
That's awesome, thanks!

PR #265, opened 2014-10-16
Updated note about Firefox and __proto__ restrictions

evilpie:
Correct behavior was implemented in https://bugzilla.mozilla.org/show_bug.cgi?id=1061853.


kangax:
Thanks. Can you please update corresponding data-es6.js as well? https://github.com/kangax/compat-table/blob/gh-pages/data-es6.js#L1051

evilpie:
@kangax Sorry I did this on web interface

PR #268, opened 2014-10-20
Added JS engine colour stripe to platform headers.

webbedspace:
Now, platforms' table `<th>` elements have a thin colour strip denoting their engine. Colours have been set for Trident (IE), SpiderMonkey (Firefox/Rhino), JavaScriptCore (Safari, Phantom, WebKit), V8 (Chrome, Opera, Node), Carakan (Opera 12) and KJS (Konqueror). This allows browsers versions to be more easily visually identified, as well as providing context to e.g. the similarity of WebKit and iOS.

Also, this fixes a bug where the platform type headers stopped being bold in the previous CSS pullreq.

### DIFF
--- a/master.css
+++ b/master.css
@@ -173,6 +173,9 @@ label[for="sort"] {
 .also-see a {
     font-weight: bold; color: blue !important; text-decoration: underline !important; 
 }
+.platformtype {
+	font-weight: bold;
+}
 
 /* page-specific header colors */
 .es6 #header {
@@ -215,4 +218,36 @@ label[for="sort"] {
 }
 .this-browser {
     background-color: #eef; color: #55f; text-transform: uppercase; padding: 4px 5px; width: 110px; font-size: 13px; vertical-align: top; 
+}
+
+/* browser engine color stripes */
+
+/* Trident */
+th[class^=ie] {
+  border-bottom: 5px solid #2571ed;
+}
+/* SpiderMonkey */
+th[class^=firefox],
+th[class^=rhino] {
+  border-bottom: 5px solid #E66000;
+}
+/* JavaScriptCore */
+th[class^=webkit],
+th[class^=safari],
+th[class^=phantom],
+th[class^=ios] {
+  border-bottom: 5px solid #b2b2b2;
+}
+/* V8 */
+th[class^=chrome],
+th[class^=node] {
+  border-bottom: 5px solid #4db848;
+}
+/* Carakan */
+th[class^=opera] {
+  border-bottom: 5px solid #CC0F16;
+}
+/* KJS */
+th[class^=konq] {
+  border-bottom: 5px solid #7BD2FF;
 }


webbedspace:
Almost forgot the screenshot:

master.css
```diff
@@ -173,6 +173,9 @@ label[for="sort"] {
 .also-see a {
     font-weight: bold; color: blue !important; text-decoration: underline !important; 
 }
+.platformtype {
+	font-weight: bold;
```
kangax, line 177: Some weird spacing here but no biggie

kangax:
This looks awesome, great idea, thanks @webbedspace

Delapouite:
That's nice! Thanks @webbedspace 

Some bikeshedding : Firefox's orange is very close from the red used in NO cells, whereas Chrome has the green YES for him. This introduces a bias where one may feel that Firefox has bad ES6 support compare to Chrome.
Maybe other colors should be used ?

kangax:
Yeah, I agree, other color would be better. The strips also blend with table colors a little. Maybe move them to the top of the cell? Finally, we should probably add strips to compilers as well, since lack of them "interferes" with height of background progress coloring (you can see it when sorting by num her of features; currently 2nd column is same as 3rd but should be  higher)

kangax:
(don't mind ipad autocomplete)

webbedspace:
FWIW the colours I selected for this were as follows:
- Trident: I sampled the colour of IE11's Metro tile.
- Gecko: I selected Firefox orange from the [official branding page](https://www.mozilla.org/en-US/styleguide/identity/firefox/color/). I'd tried FF light orange and FF yellow, but thought they were too bright in comparison to the other platforms. (I guess now I'll reconsider.)
- Carakan: I selected Opera red from the [official branding page](http://www.operasoftware.com/press/trademark). (Not sure what to do about this).
- V8: I wanted a colour that represented both Chrome and Node, so I selected a somewhat arbitrary green. I guess I could take the official Node colour and make it even yellower.
- KJS: I sampled a cyan from its [official icon](https://www.kde.org/stuff/clipart.php) that wasn't visually too close to Trident's.
- JavaScriptCore: at something of a loss, I used an arbitrary gray that suggests the Safari logo's silver rim.

webbedspace:
On the subject of compilers (and BESEN), I guess they could use JSConf Yellow, maybe slightly permuted if it resembles a new FF colour too much.

kangax:
Alternatively... how about we just use very-dimmed-transparent logos behind?

webbedspace:
The thing is, these colours are _not_ specifically of the browsers themselves, but their internal JS engines, which is why Node and Chrome, or Phantom and Safari, have the same. It's a way of conveying "these platforms share code, hence their results are causally related" (useful for connecting iOS, Safari and Phantom to WebKit, and also of separating Opera 12 from Opera/Chrome).

Also, the single stripe has the desirable effect of visually grouping browsers' versions together (so that the Firefox or Chrome blocks can be distinguished at a glance).

webbedspace:
Proposal: remove the thin full-colour stripe, and instead change the background bar-graph to use a grayed version of the colours.

One concern is that browsers with <10 features are now nigh-invisible, but TBF there's only about 3 of those in the non-obsolete table.

webbedspace:
_Alternatively_ we could just gray-out the stripe as it is:

The issue here, though, is that JavaScriptCore's colour is almost unreadable.

Any opinion on either of these ideas is welcome.

kangax:
That's true. We're highlighting engines, not browsers, and they have no logos... In that case, I think I really like your colored (#1) proposal.

webbedspace:
OK, let's go. #272

Issue #266, opened 2014-10-19
Platform categories broke column highlighting :p


webbedspace:
Click on e.g. the "FF34" column title. Lo and behold, it doesn't work.

PR #271, opened 2014-10-23
Converted master.js to jQuery

webbedspace:
master.js was getting a little hard to maintain, so to make it easier to expand in the future, I converted it to use jQuery. As a result, ready.js is now no longer used.

This also fixes a few recent broken features:
- Highlighting a single column (#266)
- The feature permalink anchor

There are also a few small additions:
- The 'features implemented' count is now faintly tinted red or green corresponding to how many features are implemented.
- Additionally, there's a "bar graph" gray colouring behind the platform titles that visually denotes how many features are implemented, too:

(As these are purely visual garnishes, I haven't felt the need to supply IE fallbacks for them.)


master.js
```diff
@@ -12,223 +18,161 @@ function test(expression) {
...
+        '</b>/' +
+        results.length + '</sup>')
+      // Fancy bar graph background garnish (again, no fallback required).
+      .css({'background-image':'linear-gradient(to top, #ddd 0%, #ddd ' +
+        (featuresCount * 100|0) + '%, transparent ' + (featuresCount * 100|0) +
+        '%,transparent 100%)'});
+  });
+  
+  // Cached array of sort orderings
+  ordering = [];
```
kangax, line 162: Is this declared anywhere?

master.js
```diff
@@ -238,36 +182,30 @@ domready(function() {
...
-        var cells = [].slice.call(row.cells, 3);
-        var sorted = cells.sort(comparator);
-        var scriptEl = row.getElementsByTagName('script')[0];
+      // Sort the platforms
+      var platforms = $('th.current').parent();
 
-        var firstCell = row.cells[0];
-        var secondCell = row.cells[1];
-        var thirdCell = row.cells[2];
+      var cells = [].slice.call(platforms.children('[data-features]:not(.current)'))
```
kangax, line 188: Hm, jQuery doesn't have sort over collection?

master.js
webbedspace, line 162: Ah, good catch.

kangax:
Looks fantastic @webbedspace and definitely long overdue. Thanks for all the effort!

master.js
webbedspace, line 188: Formal jQuery collections are technically not ordered like Arrays, since by and large the DOM order of the nodes trumps their inclusion order.

PR #273, opened 2014-10-25
Marked Firefox 24 as obsolete (ESR expired)

webbedspace:
Firefox 24's extended support end-of-life was a week ago. This also marks Firefox 31 as having ESR until late 2015.


kangax:
Thanks! Oh and congrats on 100 commits ;)

PR #280, opened 2014-10-28
Partial results

### LINKED ISSUE #131: let/const temporal dead zone

### LINKED ISSUE #224: Record results data differentially

### LINKED ISSUE #246: ES6: Add more TypedArray tests
Add tests for all methods which aren't covered by the [Khronos spec](https://www.khronos.org/registry/typedarray/specs/latest/). Technically the TypedArray classes are 'new' to ES, and there's a general rule that the table doesn't have separate tests for _every_ method of a new class (Symbol being excepted because each symbol property exposes engine behaviour), but since TypedArrays were previously specified in the DOM, this exception could be reasonably made.
- `%TypedArray%.from()`
- `%TypedArray%.of()`

`%TypedArray%.prototype` ES5 array methods:
- `%TypedArray%.prototype.indexOf()`
- `%TypedArray%.prototype.join()`
- `%TypedArray%.prototype.lastIndexOf()`
- `%TypedArray%.prototype.slice()`

`%TypedArray%.prototype` iteration methods:
- `%TypedArray%.prototype.every()`
- `%TypedArray%.prototype.filter()`
- `%TypedArray%.prototype.forEach()`
- `%TypedArray%.prototype.map()`
- `%TypedArray%.prototype.reduce()`
- `%TypedArray%.prototype.reduceRight()`
- `%TypedArray%.prototype.some()`
- `%TypedArray%.prototype.sort()`

`%TypedArray%.prototype` ES6 array methods:
- `%TypedArray%.prototype.copyWithin()`
- `%TypedArray%.prototype.find()`
- `%TypedArray%.prototype.findIndex()`
- `%TypedArray%.prototype.fill()`
- `%TypedArray%.prototype.keys()`
- `%TypedArray%.prototype.values()`
- `%TypedArray%.prototype.entries()`

The idea is that each of these bullets would be a single test, which loops through every TypedArray class (including `Uint8ClampedArray`) and checks for (at least) existence of the methods.

### LINKED ISSUE #275: Partial results
I'm giving this a shot:

Obviously I'll also have a fold-down button that reveals the sub-tests on click (with them hidden by default).

More info as it develops.

### LINKED ISSUE #279: Wrong Headers for ES6 table
In the **ES6** table headers, **Rhino** is shown under _Desktop browsers_ and **iOS7** is shown under _Server-ish_

webbedspace:
Closes #275. I tried to split as many tests as time permitted into subtests, reusing as much of the work I'd put into #258 as possible. The subtest groups are currently:
- arrow functions
- const
- let (this closes #131)
- spread
- class
- octal/binary
- template strings
- typed arrays (this closes #246)
- Proxy
- Reflect
- destructuring
- Symbol
- Well-known symbols
- RegExp.prototype
- Math

I could also extend this to Number, Array and String methods in the future, but I think this is enough for now.
New subtests (e.g. the `Number()` check for octal/binary) were designed to eliminate as many footnotes from the table as possible.

The table header browser results tally now uses each subtest tally fractionally - e.g. a test with 2/4 counts as 0.5 of a feature - but the tally itself is rounded to aid readability.

The subtest results data also use differential records, as previously pitched in #241 (this closes #224).

THIS ALSO CLOSES #279!

As usual, feel free to point out bugs, etc.


mathiasbynens:
Preview: https://rawgit.com/webbedspace/compat-table/partial/es6/index.html

webbedspace:
It's been revealed to me that a few specific tests are bugged or inaccurate, so hold on for the next couple hours.

Issue #281, opened 2014-10-29
Add 6to5


UltCombo:
The [6to5 compiler](https://github.com/sebmck/6to5) has a very nice [ES6 support](https://github.com/sebmck/6to5#comparison-to-other-transpilers). Perhaps add it to the ES6 compat table?

webbedspace:
I just hacked the build.js file to pass all the test code through 6to5.transform() while building the HTML, then checked the results in IE10 (a platform known to pass only 3 tests natively, while passing all of ES5). The results I got were as follows:
- arrows
  - everything except "no line break..." and "no "prototype"..."
- const
  - everything except "is block-scoped" and "TDZ" for both strict and non-strict
- let
  - basic support (strict and non-strict)
  - for-loop iteration scope (strict and no-strict)
- defaults (all)
- rest
- spread (all)
- class
  - everything except "is block-scoped" and "extends" (the latter requires `__proto__`)
- super
  - everything except "in constructors" (because that requires `__proto__`)
- object literal extensions (all)
- octal and binary literals
  - everything except the Number() tests
- RegExp 'u' flag
- destructuring
  - everything except "destructuring defaults" and "defaults in parameters"
- Unicode escapes
- Symbol.iterator

(Also it spuriously passed some of the negative `__proto__` tests, but those are due to the tests' code not checking for basic functionality first - whoops.)

(Now, I know what you're thinking: how did it fail the let block-scoping test? Well, I suspect there's a [bug](https://github.com/6to5/6to5/issues/150) in the thing.)

sebmck:
@webbedspace Any chance I can get the modified build source? Would love to iron out all 6to5 bugs.

webbedspace:
I, er, plan to integrate it as a formal pullreq to this repo in the next day or two. (Specifically, as a command-line option to build.js.)

(Incidentally, I'm a little suspicious of that Symbol.iterator result... I'm pretty sure IE10 actually evaluated `Symbol.iterator` to `<td id="Symbol">.iterator`, a.k.a. `undefined`, and the for-of loop just happened to exploit that.)

UltCombo:
Isn't `Symbol.iterator` polyfilled by `require("6to5/polyfill")`?

sebmck:
@webbedspace 6to5 only transforms and worries about syntax. ES6 polyfills are up to the developer, 6to5 never tries to emulate an entire ES6 environment. Although the optional `require("6to5/polyfill");` does it rather well.

[...]

PR #282, opened 2014-10-30
Add 6to5 colmun

e-jigsaw:
For #281 .

I checked [this](https://github.com/e-jigsaw/6to5-es6-compat-table) inspired by [teppeis/closure-compiler-es6-compat-table](https://github.com/teppeis/closure-compiler-es6-compat-table).

But, some tests are maybe incorrect. For example `let`. Compiled code from `let` test raised error

```
[ReferenceError: __let_script_executed is not defined]
```

compiled code is

``` javascript
// let
"use strict";
module.exports = function() {

            try {
              return (function () { "use strict"; __let_script_executed = true; var _foobarbaz2 = 123; return _foobarbaz2 == 123; }());
            } catch (e) {
              return false;
            }

};
```


webbedspace:
I'm not sure it's possible for a compiler to pass "`__proto__` in object literals" or the typed array tests on its own. (I'm aware the table currently lists Traceur as passing the former, but I feel that's an error, at least as compared to EJS and Closure).

e-jigsaw:
@webbedspace Could you're right in "`__proto__` in object literals" case. Error raised where it was confirmed by the following code.

```
function test() {
  var passed = { __proto__ : [] } instanceof Array
    && !({ __proto__ : null } instanceof Object);

  // If computed properties are supported, the following
  // check must also be passed.
  var a = "__proto__";
  try {
    passed &= !({ [a] : [] } instanceof Array);
  }
  catch(e) {}
  return passed;
}

console.log(test());
// => 0
```

But, "typed arrays" case is passed following code:

```
function test() {
  var buffer = new ArrayBuffer(64);
  var passed = true;
  var view;

  // Check that each int type overflows as expected.
  view = new Int8Array(buffer);         view[0] = 0x80;
  passed &= view[0] === -0x80;
  view = new Uint8Array(buffer);        view[0] = 0x100;
  passed &= view[0] === 0;
  view = new Uint8ClampedArray(buffer); view[0] = 0x100;
  passed &= view[0] === 0xFF;
  view = new Int16Array(buffer);        view[0] = 0x8000;
  passed &= view[0] === -0x8000;
  view = new Uint16Array(buffer);       view[0] = 0x10000;
  passed &= view[0] === 0;
  view = new Int32Array(buffer);        view[0] = 0x80000000;
  passed &= view[0] === -0x80000000;
  view = new Uint32Array(buffer);       view[0] = 0x100000000;
  passed &= view[0] === 0;
  // Check that each float type loses precision as expected.
  view = new Float32Array(buffer);      view[0] = 0.1;
  passed &= view[0] === 0.10000000149011612;
  view = new Float64Array(buffer);      view[0] = 0.1;
  passed &= view[0] === 0.1;
  return passed;
}

console.log(test());
// => 1
```

I will fix it.

webbedspace:
Question: which host JS environment are you running these in? Ideally the 6to5 output should be run in something confirmed to have close to zero native ES6 support (like IE10 - taking into account it does support DataView, String HTML and hoisted block functions by itself).

I ask because Typed Arrays refers to a low-level byte manipulation interface, which would not be very feasible for compiled JS code to emulate (and because 6to5's documentation makes no claim to this).

e-jigsaw:
I checked by Node.js(`v0.10.32`)

webbedspace:
That being so, I'd like to run all of the passing compiled code against IE 10 or IE 9 to confirm its actual efficacy.

e-jigsaw:
Oh, add column is hard way... X(

e-jigsaw:
@sebmck Is 6to5 supported IE 10 or IE9??

sebmck:
@e-jigsaw It should. Nothing is output that wouldn't be supported AFAIK.

PR #285, opened 2014-10-31
Modify the super() code slightly so that Closure can compile it.

MatrixFrog:
To see what Closure compiles this to, you can use the Closure Compiler
debugger: http://goo.gl/CYL53X

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -820,10 +820,11 @@ exports.tests = [
   link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-super-keyword',
   exec: function () {/*
     var passed = true;
-    var B = class extends class {
+    var Base = class {
       constructor(a) { return this.id + a; }
       foo(a)         { return a + this.id; }
-    } {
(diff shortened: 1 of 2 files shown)


webbedspace:
If Closure can't handle having a class expression as the extends clause of another class expression, then it hasn't implemented `class` correctly.

Admittedly, this has no bearing on its implementation of `super()`, so I guess this pullreq is fair enough - but this probably means an additional class subtest (#280) may be needed to test the functionality of class-expression-in-extends.

MatrixFrog:
We intentionally implemented only a subset of all valid `class` expressions, to keep the generated ES3 code simple.

kangax:
Do we still need this?

webbedspace:
Actually, I (or we) haven't yet rewritten the class tests to avoid using the problematic form, or making its usage a separate test. Obvs this particular pullreq no longer works (as the class test is now separate subtests).

MatrixFrog:
Okay, closing this then.

PR #287, opened 2014-11-05
Fixed tooltip positioning

webbedspace:
Instead of only positioning the tooltip during the mouseover event, this new code keeps the div constantly positioned to the right of the cursor. 

Also, the tooltip now disappears instantly when moving off the `c` icon, which makes them easier to intentionally dismiss (even if this makes them less like 'native' tooltips).


kangax:
Ok, I think it's fine. Thanks @webbedspace

PR #296, opened 2014-11-07
Removed a no longer true note for Map and Set

phistuck:
Looks like new Map([[key, value]]) and new Set([obj1, obj2]) succeed in Chrome 38.


webbedspace:
FWIW I'm working on a pullreq that converts that footnote into a proper feature subtest.

webbedspace:
OK, as #301 passed, you can cancel this one now.

PR #300, opened 2014-11-10
Fix test code for "arrow functions: no line break between params and =>"

teppeis:
I think this original intermediate arrow function is invalid syntax as ES6.


webbedspace:
Good, thanks. (The purpose of that wrapper is to confirm that arrows are initially supported before checking that the line break is regarded as invalid.)

PR #303, opened 2014-11-11
extra tests for `Map`, see data-es6.js

Yaffle:
Sorry, the compatibility data i have is only for Chrome, Firefox, IE 11.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -1829,19 +1829,19 @@ exports.tests = [
     '`-0` key should be converted to `+0`': {
       exec: function () {/*
         var map = new Map();
-        map.set(-0, -0);
+        map.set(-0, 42);
         var k = undefined;
         map.forEach(function (value, key) {
           k = 1 / key;
         });
-        return k === Infinity;
+        return k === Infinity && map.get(+0) == 42;
       */},
       res: {
         firefox29:   true,
         chrome39:    true,
       },
     },
-    '"set" should return Map object': {
+    'Map.prototype.set returns this': {
       exec: function () {/*
         var map = new Map();
         return map.set(0, 0) === map;
@@ -2007,6 +2007,31 @@ exports.tests = [
         ios8:        true,
         webkit:      true,
         chrome37:    true,
+      },
+    },
+    '`-0` key should be converted to `+0`': {
+      exec: function () {/*
+        var set = new Set();
+        set.add(-0);
+        var k = undefined;
+        set.forEach(function (value) {
+          k = 1 / value;
+        });
+        return k === Infinity && set.has(+0);
+      */},
+      res: {
+        firefox29:   true,
+        chrome39:    true,
+      },
+    },
+    'Set.prototype.add returns this': {
+      exec: function () {/*
+        var set = new Set();
+        return set.add(0) === set;
+      */},
+      res: {
+        firefox33:   true,
+        chrome39:    true,
       },
     },
   },
--- a/data-es6.js
+++ b/data-es6.js
@@ -1826,6 +1826,31 @@ exports.tests = [
         chrome36:    true,
       },
     },
+    '`-0` key should be converted to `+0`': {
+      exec: function () {/*
+        var map = new Map();
+        map.set(-0, -0);
+        var k = undefined;
+        map.forEach(function (value, key) {
+          k = 1 / key;
+        });
+        return k === Infinity;
+      */},
+      res: {
+        firefox29:   true,
+        chrome39:    true,
+      },
+    },
+    '"set" should return Map object': {
+      exec: function () {/*
+        var map = new Map();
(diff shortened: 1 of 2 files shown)


data-es6.js
```diff
@@ -1826,6 +1826,31 @@ exports.tests = [
         chrome36:    true,
       },
     },
+    '`-0` key should be converted to `+0`': {
+      exec: function () {/*
+        var map = new Map();
+        map.set(-0, -0);
+        var k = undefined;
+        map.forEach(function (value, key) {
```
webbedspace, line 1844: This test, as written, fails if the browser supports it but doesn't also support Map#forEach. I believe it should instead be written as:

```
var map = new Map();
map.set(-0, true);
return map.has(0) && map.get(0);
```

webbedspace:
No identical tests for Set/WeakMap/WeakSet?

data-es6.js
```diff
@@ -1826,6 +1826,31 @@ exports.tests = [
...
+          k = 1 / key;
+        });
+        return k === Infinity;
+      */},
+      res: {
+        firefox29:   true,
+        chrome39:    true,
+      },
+    },
+    '"set" should return Map object': {
```
webbedspace: I would refer to the set method by its full name, `Map.prototype.set`, as currently this table uses the convention that prop names in quotes are instance own properties (compare `Function "name" property`). And instead of "Map object", I'd use `this`, to explicate that the original object is returned. Additionally, I'd place this test next to the `Map.prototype.set` existence test, as an obvious grouping of like items.

data-es6.js
Yaffle, line 1844: map.get(+0) == map.get(-0); // Chrome 38 - OK
but with `forEach` you can see, that it is stored as `-0`

Later the spec was changed for this case:
see https://people.mozilla.org/~jorendorff/es6-draft.html#sec-map.prototype.set : `If key is −0, then let key be +0.`

data-es6.js
Yaffle: so how to name this test? "Map.prototype.set should return this object" ?

data-es6.js
Yaffle, line 1844: Possibly, it is better to add `map.get(+0) === map.get(-0)` to the code

data-es6.js
webbedspace: The existing tests seem to avoid using "it should do ..." names, rather sticking to the more terse "it does..." phrasing. So, I guess it'd be "Map.prototype.set returns this".

Yaffle: OK, thanks

data-es6.js
webbedspace, line 1844: Ah, OK. I think fallbacks should be used, then: first check for `Map#forEach`, then try `Map#keys`, then fallback to `for...of` or spread `...` to call the iterator. If those fail, then we can conclude that the keys cannot be retrieved, and mark it failed.

Yaffle, line 1844: hm... but is it really needed? `forEach` iteration is widely supported.

webbedspace, line 1844: I think it's important that A) individual tests are as de-coupled from other features' existence as feasible, B) the table records accurate results for every browser that is listed, all the way back to FF11. (Several other people can correct these results after the pullreq is accepted.) FF24 supports `Map#keys` but not `Map#forEach`. FF18 supports neither but does support `for...of`. (It may seem apparent that any browser predating `Map#forEach` would sure fail the test due to the SameValueZero requirement not being specced until after, but you never can be sure.)

Yaffle, line 1844: but it is a little hard to add `for..of`, as it's not ES5 syntax;
and the "iterator protocol" was updated for `Map#keys` only in FireFox 27, seems ( https://bugzilla.mozilla.org/show_bug.cgi?id=881226 )...

Yaffle:
@webbedspace , I have updated my code.

> . Additionally, I'd place this test next to the Map.prototype.set existence test, as an obvious grouping of like items.

There are no `Map.prototype.set existence test`, because it is inside of the "base test".

Also, I checked that Mobile safari 8.1 fails both.
Looking at https://github.com/google/traceur-compiler/blob/master/src/runtime/polyfills/Map.js I see, that it should fail `-0` test.

webbedspace:
Sorry, that's what I meant - it should ideally be next to the basic functionality subtest, and not at the end of the subtest list.

Yaffle:
OK

Yaffle:
Why not to merge this?

webbedspace:
I guess this is fine now - maybe the test code is a little unclear to read, but that's minor enough that it can be sorted out later once it's merged.

kangax:
Please resubmit coz I can't merge automatically

Yaffle:
@kangax , ok, new pull request - https://github.com/kangax/compat-table/pull/314

PR #309, opened 2014-11-14
ES6 compiler testing pages

webbedspace:
This changes `build.js` so that in addition to outputting `es6/index.html`, it additionally compiles the ES6 tests' code using various ES6 compilers, and creates HTML test pages using the output, in the `es6/compilers/` directory. These pages are "auxiliary" pages, comparable to the ES5 strict mode page, and thus only contain the current browser column.

These test pages are not yet public-facing (they aren't linked from `es6/index.html`) but have been cleaned up to be mostly presentable.

The test pages are primarily designed for gleaning compiler results easily. If you wish to record results based on their browser response, take care to view them in IE 10, or another browser with low native ES6 support (and subtract from the results the few features that the browser does support natively).

Issue #305, opened 2014-11-12
Let shadowing test


webbedspace:
After discovering [this thing](https://github.com/6to5/6to5/issues/150), it's become clear that the block-scoping let tests might need to be in two parts: one for identifiers shadowing identical ones, and one (the existing one) for identifiers leaving scope altogether.

To whit, the former would be an added test that goes like this:

```
let bar = 123;
{
    let bar = 456;
}
return bar === 123;
```

webbedspace:
One problem I've discovered is that block-scoping is actually checked in a few different features - `let`, `const`, `class` - and in a few different contexts within those features - block scope, for-loop head scope, etc., and adding different _kinds_ of block-scoping checks would probably lead to combinatorial explosion and an unpleasant feeling of such tests being "clutter". So, I'm wondering whether to drop the reference-error block-scoping check and replace it just with the easier shadowing check. (This would, of course, require the results to be re-confirmed, as it's regarded as a more lenient test.)

PR #310, opened 2014-11-14
Added 6to5 and IE Technical Preview result columns.

### LINKED ISSUE #306: Update/add IE tech preview
- https://twitter.com/jdalton/status/532632322919518208
- https://twitter.com/dstorey/status/532625543078760448

webbedspace:
6to5's results for certain tests that used eval() (such as const early errors) required some manual confirmation, and some of the block-scope checks were regarded more leniently due to concerns raised in #305, but for the most part they are based off the [test file](https://rawgit.com/webbedspace/compat-table/compiler-testing/es6/compilers/6to5-polyfill.html) in #309.

Closes #306 and #281, and also supplants #282 (which has a few problems separating compiled/polyfilled vs. native results).

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -10,21 +10,27 @@ exports.skeleton_file = 'es6/skeleton.html';
 
 exports.browsers = {
   tr: {
-    full: 'Traceur compiler',
+    full: 'Traceur',
     short: 'Traceur',
-    obsolete: false, // always up-to-date version
+    obsolete: false,
+    platformtype: 'compiler',
+  },
+  _6to5: {
+    full: '6to5',
+    short: '6to5 +<br>polyfill',
+    obsolete: false,
     platformtype: 'compiler',
   },
   ejs: {
     full: 'Echo JS',
     short: 'EJS',
-    obsolete: false, // always up-to-date version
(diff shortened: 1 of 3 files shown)


kangax:
Excellent, thanks @webbedspace!

kangax:
@webbedspace somewhat relatedly, I wanted to mention the discrepancy in total tests (44 vs. 39) which I'm sure you know about. I'm guessing this happens because we don't count annex "features" as part of compilers' score? The thing is that I don't exactly like it. I'm sure it's at least confusing to most people, it looks strange when sorting columns by features, and probably looks like an error. I think we should either count them everywhere to not count at all. What do you think?

webbedspace:
Counting them everywhere is not something I want - non-browsers, Node etc. have no onus to support annex B, and no one wants them be marked down because they didn't polyfill String#blink.
Proposals:
- Replace with a percentage score.
- Replace with "x/39" for non-browsers and "x/39 + x/5" for browsers (to make it more explicit that the extra 5 is "separate" from the 39).
  Probably gonna go with the first one.

kangax:
Yep, percentage would be perfect

PR #312, opened 2014-11-15
Added TypeScript column

webbedspace:
TS only supports a few things right now, but they're there. Note: TS fails all of the class tests because it doesn't allow classes declared anywhere other than module top-level.


kangax:
@webbedspace could you please rebase?

webbedspace:
OK, done.

PR #314, opened 2014-11-15
extra tests for `Map`, see data-es6.js

Yaffle:
see https://github.com/kangax/compat-table/pull/303

I do not know, if this is fixed in "IE Technical Preview"


Yaffle:
thanks!

PR #315, opened 2014-11-15
Changed the block-scoping tests to check shadowing

### LINKED ISSUE #305: Let shadowing test

webbedspace:
Shadowing (where an inner scope's same named variable shadows an outer-scope identifier, but doesn't replace it) is arguably a more commonplace feature of block-scoping, and the compilers tend to support it even if they don't support identifiers fully leaving scope (as in e.g. temporal dead zone). This closes #305.

The compilers had in fact already taken this into their account for their results, mostly. This corrects Traceur's result for BLFD.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -493,8 +493,9 @@ exports.tests = [
     },
     'is block-scoped': {
       exec: function() {/*
+        const bar = 123;
         { const bar = 456; }
-        return (function(){ try { bar; } catch(e) { return true; }}());
+        return bar === 123;
       */},
       res: {
         tr:          true,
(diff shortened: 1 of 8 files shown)


kangax:
Can't automatically merge anymore

webbedspace:
OK, rebased.

webbedspace:
Re-rebased.

PR #317, opened 2014-11-17
Added some results for the new Map/Set tests

webbedspace:
- Added IE Tech Preview results for the 2 new Map/Set tests, and fixed old Firefox/Chrome results.
- Reordered the Map/Set tests so that tests involving the constructor occur before tests involving set()/add(), which then come before specific method tests.
- Fixed something in the Proxy "setPrototypeOf" test which, while not clear whether it's spec-compliant, was causing some results to be unhelpful.
- Also fixed two Reflect results for IETP.


data-es6.js
```diff
@@ -1800,6 +1800,24 @@ exports.tests = [
...
       },
     },
+    'constructor arguments': {
+      exec: function () {/*
+        var key1 = {};
+        var key2 = {};
+        var map = new Map([[key1, 123], [key2, 456]]);
+
+        return map.has(key1) && map.get(key1) === 123 &&
+               map.has(key2) && map.get(key2) === 456;;
```
kangax, line 1810: Extra semicolon?

kangax:
Thanks!

Issue #304, opened 2014-11-12
Iterator consumer tests for array, string and iterable object


webbedspace:
These syntaxen consume iterators:
- [x] `for..of`
- [x] "array" destructuring `[..]`
- [x] spread `...`
- [x] `yield *`
- [x] `Array.from()`
- [ ] `Map()`/`Set()`/`WeakMap()`/`WeakSet()`
- [ ] `Promise.all()`/`Promise.race()`

The thing is, though, to test them you must feed them an _iterable_. Currently, these features' tests only provide them with arrays (except `yield *`, which takes a generator's iterator, and spread, which has separate tests for strings). Ideally, though, they'd have tests for all of the following:
- Arrays (pre-ES6 built-in)
- Strings (pre-ES6 built-in)
- Arbitrary iterables produced from generators (or with `Symbol.iterator` if those aren't supported).

All 3 of these types could have different levels of support (and the results for spread are already indicative of this).

PR #320, opened 2014-11-19
Added extra iterator consumer tests

webbedspace:
Now, for-of, spread and destructuring have tests that check their functionality on strings, arrays and generic iterators alike. This marks off some of #304 (but not all of it).


kangax:
Thanks!

PR #321, opened 2014-11-19
Some small corrections

webbedspace:
- Fixed an unsightly typo on the "WeakSet#add returns this" test.
- Fixed a few results for Node Harmony and Chrome.


kangax:
Thanks!

PR #326, opened 2014-11-21
Added iteration tests for instances of iterables

webbedspace:
This closes another checkbox on #304, but also adds an extra test for the other iteration syntaxes: objects that inherit from iterables:

```
var iterable = function*(){ yield 2; yield 5; }();
Math.max(...Object.create(iterable));
```

This adds such tests for spread, for-of, destructuring and `yield *`.


webbedspace:
Forgive me for one small GitHub Q: as a repo owner, do you get auto-notifications when a pullreq is rebased?

kangax:
Nope. Only when PR is submitted or commented on, afaik. Any further changes (rebase, additional commits, squashing, etc.) I don't see.

PR #328, opened 2014-11-23
Removed Gratipay button for webbedspace

webbedspace:
I currently don't have a Gratipay account, so this button has no purpose.

Also cleaned up some whitespace in data-es6.js while I was at it.


kangax:
Alright, although I'd love to keep it fare (by at least displaying 2 buttons; perhaps you can create an account?)

Thanks for changing skeleton.html; I only updated index.html as a way to quickly test out buttons positioning (they weren't initializing locally via file:///) and then forgot to update the source.

Issue #331, opened 2014-11-24
Proxy "defineProperty" handler needs to return success indication


anba:
The "defineProperty" trap needs to return true, otherwise a TypeError will be thrown, cf. `Object.defineProperty` and 7.3.5 DefinePropertyOrThrow.

``` diff
--- a/data-es6.js
+++ b/data-es6.js
@@ -2580,6 +2580,7 @@ exports.tests = [
           new Proxy(proxied, {
             defineProperty: function (t, k, d) {
               passed = t === proxied && k === "foo" && d.value === 5;
+              return true;
             }
           }),
           "foo",
```

Issue #330, opened 2014-11-24
Invalid test case for TDZ in default function parameters with var-declaration


anba:
The third test case in "default function parameters - temporal dead zone" is incorrect.

``` javascript
(function(a=function(){ return b; }){ var b = 1;}())
```

The default parameter is never executed and even if the code is changed to:

``` javascript
(function(a=function(){ return b; }()){ var b = 1;}())
```

It will only try to resolve the global variable "b", which will most likely result in a ReferenceError, but that's not TDZ related.

webbedspace:
OK, I'll rewrite it like so and make it a separate test ("separate scope"):

``` js
(function(a=function(){ return typeof b === 'undefined'; }){ var b = 1; return a(); }())
```

Issue #334, opened 2014-11-24
"Symbol.hasInstance" needs to use Object.defineProperty


anba:
The test case for "Symbol.hasInstance" needs to use `Object.defineProperty` to install the hasInstance handler, because the inherited property on `Function.prototype` is non-writable (and non-configurable), cf. http://wiki.ecmascript.org/doku.php?id=strawman:fixing_override_mistake .

``` diff
--- a/data-es6.js
+++ b/data-es6.js
@@ -4005,7 +4005,9 @@ exports.tests = [
         var passed = false;
         var obj = { foo: true };
         var C = function(){};
-        C[Symbol.hasInstance] = function(inst) { passed = inst.foo; return false; };
+        Object.defineProperty(C, Symbol.hasInstance, {
+          value: function(inst) { passed = inst.foo; return false; }
+        });
         obj instanceof C;
         return passed;
       */},
```

Issue #333, opened 2014-11-24
Test "hoisted block-level function declaration" not compliant to Annex B semantics


anba:
The test 'hoisted block-level function declaration' in `data-es6.js` does not conform to the semantics described in AnnexB (ES6, draft Oct 14, 2014). To match the proposed semantics for ES6, the test case needs to be rewritten as follows:

``` diff
--- a/data-es6.js
+++ b/data-es6.js
@@ -2987,9 +2987,9 @@ exports.tests = [
   link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-block-level-function-declarations-web-legacy-compatibility-semantics',
   exec: function () {/*
     // Note: only available outside of strict mode.
-    var passed = f() === 2 && g() === 4;
     if (true) { function f(){ return 1; } } else { function f(){ return 2; } }
     if (false){ function g(){ return 3; } } else { function g(){ return 4; } }
+    var passed = f() === 1 && g() === 4;
     return passed;
   */},
   res: {
```

PR #335, opened 2014-11-25
Fixed several test code bugs

### LINKED ISSUE #329: Incorrect "generic iterables" tests when native generators are present
The following test cases are invalid:
- for..of loops - with instances of generic iterables
- generators - yield *, instances of iterables
- destructuring - with instances of generic iterables
- spread (...) operator - with instances of iterables, in calls
- spread (...) operator - with instances of iterables, in arrays

The test case failure can be reduced to:

``` javascript
var iterable = __createIterableObject(1, 2, 3);
var iterator = Object.create(iterable)[Symbol.iterator]();
iterator.next(); // <- throws a TypeError
```

The TypeError gets emitted because the `iterator` does not pass the tests in 25.3.3.2 GeneratorValidate.

webbedspace:
Fixes #329, #330, #331, #332, #333, #334, via the following:
- Changes the definition of `__createIterableObject` to use `Symbol.iterator` first, thus correcting the "instances of generic iterables" tests (which wrongly failed implementation with both `Symbol.iterator` and generators).
- Changes the "`yield *`, generic iterables" and "`yield *`, instances of iterables" tests to use `__createIterableObject` instead of generators to produce the iterable.
- Split off the defaults "temporal dead zone test" into a "separate scope" test.
- Fixed the Proxy "define property" handler test not making the handler return a boolean.
- Fixed several mistakes in the function "name" property tests.
- Redid the "hoisted block-level function declaration" test, updating its results accordingly.
- Fixed the `Symbol.hasInstance` test to use Object.defineProperty, to overcome a limitation of Function instances.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -830,11 +830,19 @@ exports.tests = [
             eval("(function(a=b,b){}())");
             return false;
           } catch(e) {}
-          try {
-            eval("(function(a=function(){ return b; }){ var b = 1;}())");
-            return false;
-          } catch(e) {}
           return true;
+        }());
+      */},
+      res: {
+      },
+    },
+    'separate scope': {
+      exec: function(){/*
+        return (function(a=function(){
+          return typeof b === 'undefined';
+        }){
+          var b = 1;
+          return a();
         }());
       */},
       res: {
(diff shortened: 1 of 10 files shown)


webbedspace:
Merging this only auto-closed #329 and not the others. IDG how GitHub works.

kangax:
I was never quite sure myself, so just looked it up and according to https://help.github.com/articles/closing-issues-via-commit-messages/ it has to be "closes #330, closes #331", etc.

webbedspace:
OK. Anyway, you can close #330, #331, #332, #333 and #334 now.

Issue #336, opened 2014-11-26
Add nested destructuring rest test


webbedspace:
I've been told that this is now allowed:

``` js
var [a, ...[b, ...c]] = [1, 2, 3, 4];
```

Thing is, I can't find any notes on what the result would be. I believe it would be:

``` js
return a === 1 && b[0] === 2 && b[1] === c && c[0] == 3 && c[1] === 4;
```

...but I'm not 100% sure.

webbedspace:
OK, I've figured it out: the actual test is

``` js
var a = [1, 2, 3], first, last;
[first, ...[a[2], last]] = a;
return first === 1 && last === 3;
```

If this used `[first, a[2], last] = a` then `last` would be 2.

Issue #327, opened 2014-11-22
November spec changes


webbedspace:
- [x] Remove the WeakMap/WeakSet #clear method.
- [x] Rename the RegExp.prototype methods: 
  - RegExp.prototype[Symbol.match]
  - RegExp.prototype[Symbol.replace]
  - RegExp.prototype[Symbol.search]
  - RegExp.prototype[Symbol.split]
- [x] Add RegExp.prototype.flags
- [x] New proxy subtest: Array.isArray returns true for an array's proxy.
- [x] New proxy subtest: JSON.stringify stringifies an array's proxy as if it were an array.
- [x] New misc. subtest: `new RegExp(/./,'g')` no longer throws.
- [x] Remove well-known symbol Symbol.isRegExp
- [x] Add well-known symbol Symbol.species

Also, assigning to consts is now a runtime error instead of a syntax error. However... I don't want to penalise implementations that currently have it as a syntax error. So, I want the rest to remain as is (since it returns true for both runtime errors and syntax errors) but renamed with the "syntax" word removed.

PR #338, opened 2014-11-27
November ES6 spec changes

webbedspace:
This adds/removes/changes the following tests:
- RegExp.prototype methods
- Symbol.species
- String.prototype.includes
- WeakMap/WeakSet.prototype.clear
- Array.isArray on proxies
- JSON.stringify on proxies

This closes some but not all of #327.

It also includes a nested destructuring rest test (as described in #336) which I think is accurate, but maybe could do with an expert opinion.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -503,7 +503,7 @@ exports.tests = [
         ie11:        true,
       }
     },
-    'redefining a const is a syntax error': {
+    'redefining a const is an error': {
       exec: function() {/*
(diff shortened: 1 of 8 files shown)


Yaffle:
> RegExp.prototype[Symbol.match]

Such titles do not show the goal of RegExp-releated changes in ES6

kangax:
Thanks @webbedspace, as always!

PR #339, opened 2014-11-28
Removed ES6 compiler test pages from version tracking

webbedspace:
Since these are not public-facing, take a good while to re-compile, and make more complicated diffs, there's little reason to keep them tracked. Their purpose is primarily to provide results for the proper ES6 table, and so they can simply be re-compiled and viewed locally.


webbedspace:
Rebased.

PR #343, opened 2014-12-02
Added RegExp#flags and yield-sending tests.

### LINKED ISSUE #341: Add yield-as-expression tests
I think two tests are needed:
- A basic test that determines if passing a value to an iterator's .next() will cause the preceding `yield` to evaluate to the passed value.
- A different test to determine `yield`'s operator precedence (see [here](https://gist.github.com/lukehoban/8678463)).

webbedspace:
Closes #341.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -1499,6 +1499,47 @@ exports.tests = [
         nodeharmony: true,
       },
     },
+    'sending': {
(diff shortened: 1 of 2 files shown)


webbedspace:
Rebased.

kangax:
Thanks!

PR #359, opened 2014-12-12
Added React JSX results

### LINKED ISSUE #356: Show ES6 support in JSX
Inspired by http://stackoverflow.com/questions/27414384/what-es6-features-are-supported-in-jsx

### LINKED ISSUE #357: Typo in «function "name" property - symbol-keyed methods»
There is a small typo in [this line](https://github.com/kangax/compat-table/blob/1a8f15951691056d20d8d42ac7e9568412a7685d/data-es6.js#L3696):

s/sym/sym1/

webbedspace:
This closes #356, but also includes a fix to close #357 as well.

Note: JSX technically supports `extends`, but its compiled output fails the current test (that the prototype chain is correct) which I think is fair.

### DIFF
--- a/build.js
+++ b/build.js
@@ -32,6 +32,7 @@ var to5        = require('6to5');
 var esnext     = require('esnext');
 var es6tr      = require('es6-transpiler');
 var traceur    = require('traceur');
+var reacttools = require('react-tools');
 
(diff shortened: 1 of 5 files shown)


kangax:
Awesome, thanks @webbedspace!

PR #366, opened 2014-12-14
Added computed property destructuring test

webbedspace:
This also re-compiles es6/index.html after that EchoJS pullreq.


kangax:
This one is quite... mindbending

Issue #257, opened 2014-10-01
[].find is not available on chrome


jonathanong:
at least without experimental flags enabled

webbedspace:
This is a duplicate of #188 (sorta).

PR #369, opened 2014-12-15
Separated Chrome results into flagged and unflagged values

### LINKED ISSUE #188: Show Chrome support without #enable-javascript-harmony
Some developers may be interested in what ES6 features Chrome supports for stable users, not just those with #enable-javascript-harmony enabled. At the moment this could be a new column, but I'm thinking it'd be more correct if the current column had some augmentation to it: all of the features that are present but require #enable-javascript-harmony have a badge (similar to caniuse's prefix marker) in their squares.

webbedspace:
Now, Chrome features only available with the user-side "Experimental Javascript Features" flag are marked with a special "flagged" cell, and do not contribute to its ES6 support percentage. This basically closes #188 and #257.

The "flagged" cell's colour is red with a green corner triangle, to vaguely suggest that the red field is "peeling off" revealing green underneath.


kangax:
@webbedspace any chance you can push it to your gh-pages so that i can take a look at http://webbedspace.github.io/compat-table/es6/

webbedspace:
You could use https://rawgit.com/webbedspace/compat-table/chrome/es6/index.html instead.

kangax:
That works, thanks. What's interesting is that our 2 Chrome columns are almost identical, having only 1 test difference — `%TypedArray%.prototype.forEach`. I'm saying it because alternatively, we could have "Chrome" and "Experimental Chrome" columns instead, and it could make it a little more clear (I'm a bit concerned those green corners might go unnoticed). On the other hand, 2 columns won't scale well once future Chrome versions start to diverge in results...

kangax:
Another thing is that you can't see percentage of "Experimental Chrome". For example, at some point they might differ as much 10% and I think it would make sense to clearly show that. I'm not sure how though. Perhaps a toggle checkbox in Chrome header? Or global checkbox? But if we do it globally, would we also need to change Node + Node Harmony columns accordingly?

webbedspace:
Just to answer the last bit: I feel like the Node `--harmony` flag is quantifiably different to the Chrome "Experimental JavaScript" flag, because the former is in the author's control and the latter has to be configured by every recipient end-user. Which is to say, the latter boils down to "No", whereas the former boils down to "Yes". So, if there was a checkbox for revealing flagged support, I think Node's should be on by default and Chrome's should be off (which implies they'd be column-specific).

webbedspace:
Update: I decided the checkbox proposal was a bit too confusing/complicated, so instead I just had flagged support be shown as a second bar behind the first bar. This, at the least, conveys the size of the difference between the two levels of support, even if only unflagged support has a percentage listing. [PTAL](https://rawgit.com/webbedspace/compat-table/chrome/es6/index.html), as usual. I don't feel motivated to work too much more on this.

webbedspace:
(Rebased)

kangax:
Thanks! I'll take a look soon-ish (not by the computer now)

kangax:
Checked on a phondate, I think it's pretty great (at least for now) so thanks a bunch, as always! Merging.

kangax:
Can't auto-merge so please feel free to do it yourself.

PR #373, opened 2014-12-28
Added ES6-Transpiler column, and updated other compilers' results.

webbedspace:
This adds a column for [ES6 Transpiler](https://github.com/termi/es6-transpiler), and updates the results for Traceur and JSX, as well as correcting a result for IETP.

One thing I'm not too sure of is whether compiler features that require native `__proto__` should be listed as "Yes" or "No". Currently I'm using "No" because they don't work in IE10 (which is arguably a mandatory feature for a compiler).


kangax:
Awesome, thanks @webbedspace

PR #374, opened 2014-12-29
Sort compilers by most ES6/7 support

jamiebuilds:
It seems like for the most part that each group is sorted by most supported (browsers sub-sorted by version) with the exception of "server-ish". So I moved 6to5 to the first position since it now has the most support for transpilers.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -10,15 +10,15 @@ exports.target_file = 'es6/index.html';
 exports.skeleton_file = 'es6/skeleton.html';
 
 exports.browsers = {
-  tr: {
-    full: 'Traceur',
-    short: 'Traceur',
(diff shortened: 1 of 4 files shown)


webbedspace:
_Well_, this apparently sorting is historically a coincidence - until recently IE's support was lowest among popular browser brands. And, development of both compilers (Traceur and 6to5) is at a steady pace, and I don't know about anyone else, but I don't want to have to shift around the columns whenever one slightly edges out the other.

jamiebuilds:
Would it be better to work out an actual sorting order enforced in the build or clientside?

However, I believe 6to5 was already pushed to it's current place from being last in the list after it had better support than most of the others.

PR #375, opened 2014-12-30
Improved Promise tests, and added support for async testing

### LINKED ISSUE #247: ES6: Make the Promise test better (without document.write)

webbedspace:
Closes #247. This makes the Promise test call `Promise#then` and `Promise#catch` and checks what behaviour results, and also adds `Promise.all` and `Promise.race` tests.

In order to get the Promise test to work, I needed to implement asynchronous test result writing. To do this, test functions may now, instead of returning true/false, instead at some point call `asyncTestPassed()`, which is a function of no arguments, provided by the testing environment, that records the test as passed regardless of execution order. `build.js` has been updated to provide such a function (which manually updates the class and text of the corresponding `<td>`) as has the `node.js` test runner script.

This is a re-do of #248, now fully up-to-date with the current codebase.

### DIFF
--- a/build.js
+++ b/build.js
@@ -184,6 +184,7 @@ function dataToHtml(skeleton, browsers, tests, compiler) {
   var head = $('table thead tr:last-child');
   var body = $('table tbody');
   var footnoteIndex = {};
+  var rowNum = 0;
   
   function interpolateResults(res) {
     var browser, prevBrowser, result, prevResult, bid, prevBid, j;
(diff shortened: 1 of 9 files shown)


kangax:
Thanks @webbedspace. I like this implementation, pulling in. Will document all this at some point in readme.

kangax:
@webbedspace have you seen shifted cells in Promise row? I think it happened after this PR. Could you please take a look.

PR #376, opened 2014-12-30
Added 'built-in prototypes are not instances' misc. test

### LINKED ISSUE #284: ES6 spec incompatibility miscellany
Considering ES5 now has the `parseInt` test, we might as well add some checks for these:
- [x] Duplicate property names are now permitted in strict mode: `(function(){'use strict';0,{a:1,a:1}}())`.
- [x] `new Date(NaN).toString()` yields `"Invalid Date"`
- <del>`[-1,0,1,NaN].sort(function(a){return a})` yields `[-1,0,NaN,1]` (Note that `[-1,0,1,NaN].sort()` still yields `[-1,0,1,NaN]`.</del>
- [x] A semicolon is no longer required at the end of a do-while statement.
- [x] The `= 0` bit of `for(var i = 0 in {})` is now illegal.
- <del>The `var e` bit of `try{}catch(e){var e;}` is now illegal.</del>
- [x] `Object.freeze`, `Object.seal`, `Object.preventExtensions`, `Object.getOwnPropertyDescriptors`, `Object.getPrototypeOf`, `Object.isExtensible`, `Object.isSealed`, `Object.isFrozen` and `Object.keys` no longer throw TypeError when passed a primitive.
- [x] `Boolean.prototype`, `Number.prototype`, `RegExp.prototype`, `Array.prototype`, `String.prototype` and `Date.prototype` no longer behave like instances of their constructors (hence, `Boolean.prototype.valueOf()` should throw, as should `RegExp.prototype.exec("")` and `Array.prototype.push(0)`).

webbedspace:
Closes #284.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -5271,6 +5271,28 @@ exports.tests = [
       res: {
       },
     },
+    'built-in prototypes are not instances': {
+      exec: function(){/*
+        try { 
+          Boolean.prototype.valueOf(); return false;
+        } catch(e) {}
+        try {
+          Number.prototype.valueOf(); return false;
+        } catch(e) {}
+        try {
+          String.prototype.toString(); return false;
+        } catch(e) {}
+        try {
+          RegExp.prototype.source; return false;
+        } catch(e) {}
+        try {
+          Date.prototype.valueOf(); return false;
+        } catch(e) {}
+        return Array.prototype.length === undefined;
+      */},
+      res: {
+      },
+    },
   },
   separator: "after",
 },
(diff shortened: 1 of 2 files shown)


kangax:
Took me a bit to understand that this is to check that properties/methods don't throw. Makes me wonder if perhaps we should add small explanation to some tests.

webbedspace:
You mean _do_ throw (the test requires all of them (except the last one) throw due to the prototypes lacking the requisite internal slots for each method).

RegExp#source is admittedly a bit of a puzzle: unlike Arrays' "length" own-property, it's specced to be a getter, and as such even using it as a statement should cause it to throw.

You may also notice that Function.prototype is missing - since a few libraries use Function.prototype as a "no-op" function, such as a callback, then the ES6 spec can't change that.

kangax:
Yeah, ironically I meant "do throw" of course :) about Function.prototype — makes sense. Back in the days, I was advising _against_ using it as no-op. It's more of a show off trick to show that you did read the spec at some point :) and sacrifices clarity for the rest of the population at a questionable expense (does it really hurt to create extra empty function? Ok, you can use Function() if you're worried about unnecessary closure memory consumption in (ancient) engines that don't optimize, but at least it's simple and clear) Anyway...

PR #377, opened 2014-12-30
Added extra Array.from and Array.of tests.

### LINKED ISSUE #349: Array.from: array-like vs. iterator
Array.from has two behaviours: if the passed object is iterable, it behaves like `Array.of(...obj)`. If it's not, it tries to read numeric-index properties from the object. Both behaviours seem worthy of tests.

Note that other iterator-consumer methods like `new Map(obj)` do not give special treatment to array-likes.

webbedspace:
Closes #349. There's a problem, though: I currently can't obtain accurate test results for IE Technical Preview, EchoJS, or Firefox 32 and 33. If you or anyone else can fill any of those in, those'd be appreciated.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -4639,9 +4639,9 @@ exports.tests = [
   name: 'Array static methods',
   link: 'http://people.mozilla.org/~jorendorff/es6-draft.html#sec-properties-of-the-array-constructor',
   subtests: {
-    'Array.from': {
+    'Array.from, array-like objects': {
       exec: function () {/*
-        return typeof Array.from === 'function';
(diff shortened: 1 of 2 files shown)


kangax:
```
Array.from({ 0: "foo", 1: "bar", length: 2 }) + '' === "foo,bar";
```

FF32,33=true
IETP=true

```
Array.from(iterable) + '' === "1,2,3";
```

FF32,33=true
IETP=false

```
Array.from(Object.create(iterable)) + '' === "1,2,3";
```

FF32,33=error
IETP=false

```
class C extends Array {}
...
```

FF32,33=error
IETP=true

webbedspace:
OK, updated. The EJS results I went with were based on examining the source and referring to other results regarding instances-of-iterables.

PR #384, opened 2015-01-03
Add more computed properties tests

kangax:
Add "computed shorthand methods" and "computed accessors" tests.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -1429,27 +1429,29 @@ exports.tests = [
         return ({ [x](){ return 1 } }).y() === 1;
       */},
       res: {
+        ie11tp:      true,
         tr:          true,
         _6to5:       true,
         es6tr:       true,
-        firefox34:   true
+        firefox34:   true,
       }
     },
     'computed accessors': {
       exec: function() {/*
         var x = 'y',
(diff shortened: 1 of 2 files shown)


kangax:
@webbedspace I've added 2 more "computed properties" tests. Please double-check if everything looks fine. What's interesting is that Chrome currently supports `{ [x]: ... }` notation but not `{ [x]() { ... } }` one (Firefox is OK). There's at least 1 more test we could add — computed generator `{ * [Symbol.iterator]() { yield 'foo' }` but I'm not sure if putting it under object literal extensions makes sense (it's really a combination of 2 features; but then, so is computed shorthand methods...)

kangax:
@webbedspace I also didn't have a chance to test IETP yet

webbedspace:
Let's see here...
- FF does _not_ support computed accessors until FF34.
- IETP supports all object literal extensions.
- If you look closely, there's already a shorthand generator test, so a computed shorthand generator test would fit next to it.

webbedspace:
@kangax Merge conflict resolved.

kangax:
thanks @webbedspace!

PR #385, opened 2015-01-06
Fix for "Current-Browser" column

marcelerz:
Currently, the "Current Browser" column is broken, moving all the columns two to the left, displaying false data for each browser.

This was introduced by the "cheerio" work. 
Code that was previously:

``` javascript
            <script>
              document.write('<th class="current" title="' + navigator.userAgent + '">Current browser</th><th></th>');
            </script>
```

is now:

``` javascript
            <script>
              document.write('<th class="current" title="' + navigator.userAgent + '">Current browser</th>
<th></th>
');
            </script>
```

The change in this PR will remove the "new-line" within the JavaScript code.

### DIFF
--- a/build.js
+++ b/build.js
@@ -383,7 +383,7 @@ function dataToHtml(skeleton, browsers, tests, compiler) {
 
   $('#footnotes').append(allFootnotes());
 
-  return $.root().html().replace(/(<\/t\w>)/g, "$1\n");
+  return $.root().html();
 }
 
 function capitalise(s) {


kangax:
Thanks @marcelerz!

webbedspace:
The reason that line break was there was actually a (failed) attempt to insert a text node line-break, and keep the `<td>`s on separate lines in the compiled HTML to reduce diff sizes. If you observe [this](https://github.com/kangax/compat-table/commit/cd9c583522b5952dbe31800d97c293546c29cd9e), you'll see that several separate lines are now conflated, meaning that a change to any one of that particular test's results will involve replacing the entire line. Whether or not this is a good or bad thing is up for question, but my intention was to try and keep things the way they were.

PR #387, opened 2015-01-07
Add io.js 1.0.0

### LINKED ISSUE #351: Io.js
Any plans to add the node fork io.js?

As far as i know, they're using a newer v8 version which should support a lot of new es features :)

ruimarinho:
The upcoming version of io.js 1.0.0 will likely ship with v8 3.30.37, which provides a nice boost from the latest node unstable release (0.11.14 at this time). The result is a considerable jump from **4%** to **32%** in terms of ES6 compatibility.

Closes #351.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -283,6 +283,11 @@ exports.browsers = {
     note_id: 'harmony-flag',
     note_html: 'Flagged features have to be enabled via <code>--harmony</code> flag'
   },
+  iojs: {
+    full: 'io.js 1.0.0',
+    short: 'io.js',
+    platformtype: 'engine',
+  },
   ejs: {
     full: 'Echo JS',
     short: 'Echo JS',
@@ -531,6 +536,7 @@ exports.tests = [
         konq49:      true,
         phantom:     true,
         node:        true,
+        iojs:        true,
         ios7:        true,
       }
     },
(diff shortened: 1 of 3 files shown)


webbedspace:
Does the support require the command-line `-harmony` switch? If so, then support must be given as `flag` instead of `true`, and the corresponding Node footnote should also be given (albeit without repeating the `node_html` property).

Also, you need to edit [getBrowserColour()](https://github.com/kangax/compat-table/blob/gh-pages/master.js#L202) so that iojs is included in the regex for V8 engines.

kangax:
@webbedspace speaking of flags, I was thinking to show completeness as 4%/23% for such engines; right now we show number for non-flagged support only. What do you think?

ruimarinho:
No, no runtime flag is required on io.js to achieve this result set.

webbedspace:
@kangax I'd previously turned down this idea on the basis it'd make the top a bit crowded. But maybe if the "flagged" percent was placed vertically, above the normal percent (which would remain aligned with the other normal percents)?

kangax:
@webbedspace I'm ok with that. We'd just need to move header titles ("Compilers", etc.) slightly higher.

kangax:
(made iojs colored as V8)

PR #388, opened 2015-01-08
Update io.js 1.0.0 results

ruimarinho:
On the TC meeting of 07-01-2015, it was approved to upgrade the v8 version of io.js to 3.31.74.1. This will be the version that ships with the v1.0.0 release and brings long-awaited ES6 features with it, specifically:
- Binary and octal literals.
- New String methods.
- Block scoping.
- Template literals.

In addition, some `const` and `let` results were also corrected. All results are based from the latest nightly.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -610,6 +610,7 @@ exports.tests = [
         chrome:      flag,
         konq49:      true,
         node:        flag,
+        iojs:        true,
       }
     },
     'is block-scoped (strict mode)': {
@@ -629,6 +630,7 @@ exports.tests = [
         ie11:        true,
         firefox36:   true,
         node:        flag,
+        iojs:        true,
       }
     },
     'redefining a const (strict mode)': {
@@ -651,6 +653,7 @@ exports.tests = [
         firefox11:   true,
         chrome21dev: flag,
         node:        flag,
+        iojs:        true,
       }
     },
     'temporal dead zone (strict mode)': {
@@ -665,6 +668,7 @@ exports.tests = [
         firefox36:   true,
         chrome19dev: flag,
         node:        flag,
+        iojs:        true,
       },
     },
   }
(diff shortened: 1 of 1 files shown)


rektide:
Pardon, which if any of these do we need `--harmony-*` (those're still a thing right?) or `--harmony` flags for? Also, is there a resource to help figure these questions out with?

Updates: read iojs/io.js#90#issuecomment-68809524

ruimarinho:
@rektide, not anymore (on io.js). I'm currently preparing a resource to clarify this. All of these feature have shipped, by default, on the version of v8 that will ship with io.js 1.0.0.

Fishrock123:
None of these. These features have landed in v8 stable and have been unflagged. `--harmony` still exists for `--harmony`-flagged v8 features.

rektide:
iojs/iojs#90 has seen recent activity describing some of this change away from `--harmony` use

kangax:
Thanks @ruimarinho!

So io.js implements most (all?) class behavior but only under strict mode? I'm a little worried to see so many tests duplicated under strict mode (even though, yes, we already have few like that) — what if any other engines implement any other behavior only under strict? Would we need to add tests for those as well? Feels can-of-worm-ish.

All the other updates are +1 from me, of course.

@webbedspace what are your thoughts on this?

ruimarinho:
@kangax, that's right. Unfortunately, this is a current limitation of v8. I also dislike duplicating tests for strict mode only, especially because it is unfair for engines that have not been tested against it but already have known support for the non strict mode.

Fishrock123:
Classes and Object Literals have been moved behind a flag. 
Please see: https://github.com/iojs/io.js/issues/251

(Also, io.js core now uses Octet literals from es6)

ruimarinho:
@kangax, strict mode for classes was removed for now although the duplication problem remains - in ~6 weeks they will be reintroduced in V8. I've updated the results to remove Classes and Object Literals, so I believe this is ready to be merged. This link will be featured on the official FAQ.

kangax:
Cool, merging then

PR #391, opened 2015-01-12
Separate Object static method primitive tests

teppeis:
Because some implementations (Chrome 40, io.js and es6-shim) support a part of Object static methods for primitives, it's better to separate the tests.

FYI: https://gist.github.com/teppeis/5284739aa1c6823b2ea9

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -3943,6 +3943,100 @@ exports.tests = [
   },
 },
 {
+  name: 'Object static methods accept primitives',
+  link: 'http://people.mozilla.org/~jorendorff/es6-draft.html#sec-properties-of-the-object-constructor',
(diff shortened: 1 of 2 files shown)


kangax:
I suppose that makes sense, so +1 from me. /cc @webbedspace

webbedspace:
Technically, the getOwnPropertyNames test can be simplified to `return s.sort() + '' === "0,length"`, but w/e. (Also not sure why the getPrototypeOf test doesn't compare to String.prototype directly.)

teppeis:
@webbedspace Thanks!

> Also not sure why the getPrototypeOf test doesn't compare to String.prototype directly.

Right. It's better.

PR #393, opened 2015-01-13
Categorised the ES6 tests

### LINKED ISSUE #292: Categorising the ES6 tests
How does this sound?

Optimisation:
- proper tail calls

New bindings:
- let
- const
- block-level function

New function constructs:
- arrows
- generators
- class
- super

Literal extensions:
- rest
- object literal extensions
- octal and binary
- unicode code point escapes

Assignment / call extensions:
- defaults
- destructuring
- template strings

Iterator consumers:
- for-of loops
- spread

New standard built-ins:
- Map
- Set
- WeakMap
- WeakSet
- Typed Arrays
- Promise
- Proxy
- Reflect

Built-in extensions:
- Object static methods
- Function 'name'
- Function#toMethod
- String static methods
- String proto methods
- Array static methods
- Array proto methods
- RegExp y/u flags
- RegExp proto methods
- Number properties
- Math methods

These labels don't need to appear on the table itself.

"Iterator consumers" is maybe a bit of a misnomer because `yield *`, `Array.from`, `Map`/`Set`/`WeakMap`/`WeakSet` constructors, `Promise.all` and `Promise.race` also consume iterators.

webbedspace:
Closes #292.
See https://rawgit.com/webbedspace/compat-table/category/es6/index.html


kangax:
Nice! I like. I would do 2 things — capitalize section names and change order of sections, moving optimization before or after "misc" (I don't think it deserves to be on top). What do you think?

webbedspace:
I personally have kept TCO on top for the past half-year for purely evangelical reasons: I feel it's the most unkindly under-reported ES6 feature, considering the expressive power it provides, and the sheer difficulty in polyfilling or transpiling it. I think it's hardly at the same level of triviality as, say, the new Math functions or ES5 miscellany.

kangax:
That's a fair argument :) Ok, let's keep it on top then. It'll sure make Crockford happy once someone finally pioneers TCO.

PR #404, opened 2015-01-21
Marked io.js flagged features, and corrected some results.

webbedspace:
Specifically, this corrects several tests that used 'window' instead of 'global', and their corresponding results. This also marks the presence of class support in io.js by adding a footnote explaining that it's in strict mode only. This is not ideal (there's no way for the live tests to probe whether the presence or absence of strict mode affects a feature, without separate tests existing for each) but should provide more-correct information for now.

Also, this splits one of the super tests into two, handling statement and expression forms of super().

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -300,12 +300,13 @@ exports.browsers = {
     short: 'Node',
     platformtype: 'engine',
     note_id: 'harmony-flag',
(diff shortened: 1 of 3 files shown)


kangax:
This makes it even more important to show both flagged and unflagged percentages in headers.

PR #405, opened 2015-01-23
Improve TypeScript build time

teppeis:
Currently, building TypeScript tests is too slow because it calls `child_process.execSync` for each test.
TypeScript 1.4 publishes [Compiler API](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API) and [typescript-simple](https://github.com/teppeis/typescript-simple) uses it to compile without child process.
- Current: 293.6s
- Improved: 1.8s

### DIFF
--- a/build.js
+++ b/build.js
@@ -51,6 +51,7 @@ process.nextTick(function () {
   var esnext     = require('esnext');
   var traceur    = require('traceur');
   var reacttools = require('react-tools');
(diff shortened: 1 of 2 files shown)


kangax:
That's pretty nice, thanks @teppeis. Didn't check but merging.

teppeis:
Thanks!

PR #407, opened 2015-01-24
Updated capabilities for TypeScript in light of upcoming 1.5 alpha

DanielRosenwasser:
Because TypeScript has an ES6 emit target, we technically support many of the capabilities; however, because most of them are runtime dependent on a valid ES6 implementation, I've left a note on them.

I'm not sure what would be the best thing to do. If this is acceptable, great; however, please let us know what would be the most appropriate.

### DIFF
--- a/build.js
+++ b/build.js
@@ -333,8 +333,12 @@ function dataToHtml(skeleton, browsers, tests, compiler) {
       // Create the cell, and add classes and attributes
       var cell = $('<td></td>');
       cell.addClass(result === true ? "yes" : result !== null ? "no" : "");
-      if(result === "flagged") {
+      if (result === "flagged") {
         cell.addClass("flagged");
+      }
+      else if (result === "needs-polyfill-or-native") {
+        cell.attr('title', "Requires native support or a polyfill.");
+        cell.addClass("needs-polyfill-or-native");
       }
       cell.attr('data-browser', browserId).addClass(
         browsers[browserId].obsolete ? "obsolete" :
@@ -353,6 +357,7 @@ function dataToHtml(skeleton, browsers, tests, compiler) {
       if (result !== null) {
         cell.text(result === "flagged" ? "Flag" : result === true ? "Yes" : "No");
       }
+
       if (footnote) {
         cell.append(footnote);
       }
--- a/data-es6.js
+++ b/data-es6.js
@@ -2995,6 +2995,10 @@ exports.tests = [
         node:        true,
         iojs:        true,
         ios7:        true,
+        typescript:  {
+            val: needsPolyfill,
+            note_id: "typescript-es6",
+        },
       }),
     },
     'Uint8Array': {
@@ -3023,6 +3027,10 @@ exports.tests = [
         node:        true,
         iojs:        true,
         ios7:        true,
+        typescript:  {
+            val: needsPolyfill,
+            note_id: "typescript-es6",
+        },
       },
     },
     'Int16Array': {
@@ -3092,6 +3100,10 @@ exports.tests = [
         node:        true,
         iojs:        true,
         ios7:        true,
+        typescript:  {
+            val: needsPolyfill,
+            note_id: "typescript-es6",
+        },
       }),
     },
     'DataView (Uint8)': {
@@ -3199,6 +3211,8 @@ exports.tests = [
     var eqFn = ' === "function"';
     var obj = {};
     for (var m in methods) {
+      methods[m].typescript = { val: needsPolyfill, note_id: "typescript-es6", },
+
       obj['%TypedArray%' + m] = {
         exec: eval('0,function(){/*\n  return typeof '
           + [
@@ -3236,6 +3250,10 @@ exports.tests = [
       res: {
         tr:          true,
         babel:       true,
+        typescript:  {
+            val: needsPolyfill,
+            note_id: "typescript-es6",
+        },
         ejs:         true,
         es6shim:     true,
         ie11:        true,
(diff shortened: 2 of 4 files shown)


webbedspace:
This isn't what "support" means in the context of the ES6 table. TypeScript is only listed in the same context that Traceur and 6to5 are: as an ES6->ES5 compiler. Compiling ES6 syntax into... the same ES6 syntax... is meaningless.

DanielRosenwasser:
I understood this might be a point of contention. Some of these features are in the works of being downleveled (e.g. template strings). However, I'd say that by and large, there's a strong chance that new runtime-supported constructs such as `Symbol` will not be, but we will be _compatible_ in the sense that we semantically respect and can understand the code, and this is a compatibility table.

Again, I understand if this is not acceptable. Technically, we'd be able to support many of the proposed types using a polyfill like core-js. I think this is something that is worthy of discussion on our end.

webbedspace:
OK, I suppose there's some merit in showing how TS outputs code in the context of platforms with native support. Nevertheless, the footnoted cells must be marked "No" to indicate the output will not pass on a non-supporting platform - that is what the "Yes" cell represents for compilers/polyfills, and what the competitive-looking overall percentage measures. (For comparison, see 6to5's `extends` test result. It is marked "No" because it cannot pass on IE 9 or IE 10, despite being usable on most other browsers.)

All of the `class`/`super` results should also be marked "No" (while bearing the same explanatory footnote), because they have non-standard limitations on their use, and thus falls short of "basic support".

DanielRosenwasser:
So `Yes` is probably not the correct categorization, but neither is `No`. TypeScript doesn't mandate a specific polyfill runtime library to use - rather, we follow a "bring your own polyfill" model so as to be as general and framework independent as possible. So while we support references to ES6 types and semantic understanding of the operations which they support, polyfilling is not something we aim to achieve.

In short, what about `N/A`?

DanielRosenwasser:
Hey @webbedspace, to come back to this, wondering if `N/A` is at all a possibility.

webbedspace:
Here's what I really want: each of the compilers, when the given ES6 test code is freely passed through largely unmodified (such that an additional build step or polyfill may then process it), will receive a result a square, _still regarded as No for percentage calculation_, but coloured lurid orange and with mouseover title text reading "Requires native support or a polyfill", similar to the Annex B squares for compilers.

The main problem, though, is that the maintainability of this may be difficult. Since there's no definitive ES6 implementation yet, it isn't just an easy case of running each compiler's test HTML through it and comparing with a zero-ES6 implementation - each compiler's results for each test may have to be re-compared by eye to see if it works. Additionally, it isn't obvious what colour the "supertests" squares should be - if some compiler has two reds, two oranges and two greens, should the supertest square be a chartreuse "2/6" and with a special sigil not dissimilar to the current "flagged" sigil? I worry that this may get more than a bit obtuse.

DanielRosenwasser:
So I'm in the process of updating this PR to reflect an upcoming release of TypeScript and I'm encountering a few tough calls.

Firstly, from what it sounds, your preference is to check whether or not we have the correct runtime semantics; however, in tests for something like `const`, we _do_ give errors, but we still emit as a `var`. What are your feelings on this? You do still get an error, but instead you get it ahead of time.

Second, we have a pretty big disparity with respect to classes; as you're familiar with this, we support classes, but only at the declaration level. Clearly for tests with class expressions, we won't be able to pass those yet, but since it gives a somewhat misleading perception, is there anything we could do to improve the situation?

webbedspace:
Hmm, mapping the concept of native engine syntax errors onto compilers/polyfills is indeed a bit of an ill fit. Generally I've been regarding support as "Yes" if the compiler halts compilation, and "No" if it passes without complaint. I think I might lean to "Yes" for this middle-road behaviour.

About the `class` syntax: much like the `let` syntax, it's a bit of a mess of caveats across the board. The line between "Yes, but footnoted" and "Flag" is not that thick - originally "Flag" meant "requires a program flag on either the developer's side or the user's side" but has lately crept to mean "requires some unofficial condition be fulfilled, like strict mode or a defined constructor." I interpret TS's restriction - "must be a script/module top-level declaration" - more harshly than those other two because, unlike the requirement of extra boilerplate, it more substantially limits the ways in which the syntax can be used - you can't declare a class within a function body, which is one of the features that the basic "class statement" test kind of assumes (insofar as it's analogous to the `function` statement).

But, I _could_ add an all-new "class statement, function scope" test solely to allow TS to fail it and pass the more basic one (which is what I did for `let`, whereby I duplicated all the tests for strict mode so that Chrome/Node/iojs could pass one and fail the other). This kind of feels like gerrymandering the tests, though, which is why I tend to avoid it unless it produces more informative results for multiple platforms (and would continue to provide information for future platforms, like new V8-derived ones). As of now it looks like no other platforms impose this specific restriction on their class syntax interpretation (and few to no engines have yet to begin implementing class syntax, with even AppleWebKit weighing in recently), but, I'll think about it.

DanielRosenwasser:
Hey @webbedspace, I've updated the PR. While I'm flying out to NY and won't be able to fix the current linter issues until potentially tomorrow, you can take a look at http://DanielRosenwasser.GitHub.io/compat-table/es6 to get a feel for it.

As a note, while TypeScript 1.5 will warn users that classes outside of module scopes are not yet supported, we do now perform rudimentary emit.

DanielRosenwasser:
Just fixed the change and merged recent changes. What do you think? Preview it at [here](https://danielrosenwasser.github.io/compat-table/es6).

webbedspace:
Some thoughts:

Since the start of this pullreq and now, I've used "orange" mean "about 10% support", so now its use here for "no support, but native fallback is allowed by the compiler" - a somewhat different meaning - is called into question.

Also, I feel like some of the uses of "needs-native-or-polyfill", especially in the built-ins section, appear a bit dubious in relation to other "untyped" compilers like Babel or Traceur, which "support" natives or polyfills implicitly by design - although I do understand why that point is significant for TypeScript, where the "type" of Map or Set needs to be understood by the compiler.

Anyway, these are quite minor points, and I'll sort them out myself after merging.

PR #409, opened 2015-01-25
Marked Rhino and Safari 6 as obsolete.

webbedspace:
OS X 10.8 was EOL'd some time last year (and even if it hasn't, SF 6.1 is the current version for 10.8). 

Also hyphenated certain platform names.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -26,13 +26,13 @@ exports.browsers = {
   },
   es6tr: {
     full: 'ES6 Transpiler',
-    short: 'ES6<br>Transpiler',
+    short: 'ES6<br>Trans-<br>piler',
     obsolete: false,
     platformtype: 'compiler',
   },
   closure: {
     full: 'Closure Compiler v20141120',
-    short: 'Closure<br>Compiler',
+    short: 'Closure',
     obsolete: false,
     platformtype: 'compiler',
   },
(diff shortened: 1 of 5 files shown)


kangax:
Hm, Rhino is obsolete forever now?

webbedspace:
While it isn't really "obsolete" in the sense that another engine has officially replaced it (although I think Nashorn really has replaced it as JDK 8's JS engine), I feel like Rhino falls into the same boat as BESEN, which is to say that it's a no-longer-updated engine with, as you can see, zero non-legacy ES6 support, so there isn't much point listing it prominently in the table anymore.

Maybe an argument could be made that its negative results are still important enough to list, insofar as there are still active projects that use Rhino (such as RingoJS). Personally I feel like the table's default view has gotten a bit wide lately, hence this pullreq.

kangax:
I hear you... perhaps we should rename "obsolete" to "obsolete and rare/exotic" or some such?

gbrail:
Rhino is no longer no longer updated ;-)

I would indeed like to use this table to track ES6 compatibility for Rhino.

I wasn't able to figure out from the source how to update the table by running the code in Rhino. Is there a way to automate it? Or are the Rhino results updated manually?

webbedspace:
If you want to add new Rhino results, the `data-es6.js` file must be updated manually based on your own research.

PR #410, opened 2015-01-26
Added yet more destructuring tests.

webbedspace:
Also, reordered the compilers in build.js, and removed a console.log() from master.js.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -3651,7 +3651,10 @@ exports.tests = [
     'with arrays': {
       exec: function(){/*
         var [a, , [b], c] = [5, null, [6]];
-        return a === 5 && b === 6 && c === undefined;
+        var d, e;
+        [d,e] = [7,8];
+        return a === 5 && b === 6 && c === undefined
+          && d === 7 && e === 8;
       */},
       res: (temp.destructuringResults = {
         tr:          true,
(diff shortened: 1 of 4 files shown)


kangax:
@webbedspace out of curiosity, which criteria did you reorder compilers by?

kangax:
@webbedspace inspired by Axel's latest by any chance? :) Been meaning to read it — looked great — but didn't get a chance yet.

data-es6.js
```diff
@@ -3715,6 +3739,44 @@ exports.tests = [
...
+        return a === 1 && b === 2 && c === 3 && d === 4;
+      */},
+      res: {
+        firefox16:    true,
+        es6tr:        true,
+      },
+    },
+    'throws on null and undefined': {
+      exec: function(){/*
+        var n = null;
```
kangax: Hm, why store them as variables instead of just using inline?

webbedspace: One of the compilers (can't remember which) throws a compile error when it encounters a null literal as a destructuring source, which is technically a spec violation (it should throw a TypeError at runtime), but behaves on-spec if it's in a variable.

webbedspace:
Compilers are reordered in terms of how long they take to build their pages, with Closure currently the slowest.

data-es6.js
kangax: Interesting. I feel like we probably shouldn't be "masking" it then. But it's certainly tricky if it completely throws compiler off not letting even get output file...

webbedspace: Actually, nevermind - it turns out the compile error was for some other aspect of the test's structure (namely, that `var {a}` appeared twice, which for some reason it didn't like) which has since been rewritten.

kangax:
Thanks!

PR #412, opened 2015-01-27
make the table head sticky

geecu:
Make the table head sticky in browsers that support CSS sticky positioning. (mentioned in #75)

### DIFF
--- a/master.css
+++ b/master.css
@@ -13,6 +13,18 @@ body {
 }
 table {
     margin-bottom: 2em; width: 100%;
+    border-spacing: 0;
+}
+thead {
+    position: -webkit-sticky;
+    position: -moz-sticky;
+    position: -ms-sticky;
+    position: -o-sticky;
+    top: 15px;
+    z-index: 2;
+}
+.es6 thead {
+    top: 0;
 }
 th {
     font-weight: normal; text-align: center; padding: 1em 0.45em; position: relative; min-width: 40px; font-size: 13px; background: #eee; vertical-align: middle;
@@ -35,6 +47,9 @@ td .anchor {
 }
 td:first-child {
     min-width: 320px;
+}
+td, th {
+    border: 1px solid #fff;
 }
 a[href^='#'] {
     text-decoration: none;
@@ -350,7 +365,18 @@ label[for="sort"] {
     width: 260px;
 }
 .num-features {
-    position: absolute; top: -15px; right: 0px; font-size: 11px;
+    position: absolute;
+    top: -15px;
+    right: 0px;
+    font-size: 11px;
+    left: 0px;
+    background-color: #eee;
+    border-bottom: 1px solid #fff;
+    text-align: right;
+}
+.es6 .num-features {
+    left: auto;
+    background-color: transparent;
 }
 .num-features b {
     font-weight: bold;


kangax:
Hm, I'm seeing pretty bad lagging on FF when starting to scroll down. Anyone else? /cc @webbedspace

master.css
```diff
@@ -13,6 +13,18 @@ body {
 }
 table {
     margin-bottom: 2em; width: 100%;
+    border-spacing: 0;
+}
+thead {
+    position: -webkit-sticky;
+    position: -moz-sticky;
+    position: -ms-sticky;
+    position: -o-sticky;
```
webbedspace, line 22: None of the three prefixes above exist. Only `sticky` (for Firefox) and `-webkit-sticky` (for Safari and iOS) are [actually implemented](http://caniuse.com/#feat=css-sticky).

kangax, line 22: `position: sticky` doesn't work for me either (in FF). Might be because we're dealing with table-\* elements here.

geecu, line 22: Opened #414 to correct according to @webbedspace's comment

master.css
```diff
@@ -36,6 +48,9 @@ td .anchor {
 td:first-child {
     min-width: 320px;
 }
+td, th {
+    border: 1px solid #fff;
```
geecu, line 52: This seems to cause the scroll performance in FF... I changed the border-collapse to collapse and added this one because otherwise the scrolling rows would appear below the sticky header...

master.css
geecu, line 22: @kangax: Yes, FF doesn't seem to support sticky for table elements ( http://jsbin.com/sejoqapoda/1/edit )

kangax, line 22: For the record — https://bugzilla.mozilla.org/show_bug.cgi?id=975644 Looks like that ticket is _just about_ to be fixed. But you never know.

master.css
```diff
@@ -350,7 +365,18 @@ label[for="sort"] {
     width: 260px;
 }
 .num-features {
-    position: absolute; top: -15px; right: 0px; font-size: 11px;
+    position: absolute;
+    top: -15px;
+    right: 0px;
+    font-size: 11px;
+    left: 0px;
+    background-color: #eee;
```
webbedspace, line 373: I'm not sure why this is necessary, except that it puts somewhat ungainly gray boxes around the feature% totals in every table _except_ ES6.

webbedspace, line 373: (OK, I think I understand their purpose now, but IMHO, they should really be white boxes.)

master.css
webbedspace, line 52: JSYK this also messes up the mauve outline of the "current browser" result cells.

master.css
```diff
@@ -350,7 +365,18 @@ label[for="sort"] {
...
+    border-bottom: 1px solid #fff;
```
webbedspace, line 374: This, on the other hand, still puzzles me.

geecu, line 374: The .num-features span was positioned absolutely above the th element, it didn't have a background, so the number of features would appear on top of the underlying table rows (after scrolling). I added a background color to the span and made it stretch the whole width of the th element

webbedspace, line 374: I was puzzled by the `border-bottom` line - specifically, that it isn't overridden in the ES6 table, leaving misaligned white lines beneath the percentages (in Firefox, at least).

PR #416, opened 2015-01-28
addition of a note about FF bug with strict mode

Yaffle:


webbedspace:
Just 2 notes: 
1) This doesn't actually compile because you left off a comma.
2) We already have [an entire separate table for strict mode minutiae](http://kangax.github.io/compat-table/strict-mode/), and this really belongs there instead of in a footnote.

Yaffle:
@webbedspace 
oh, excuse me for this

kangax:
Damn it... thanks @webbedspace, I really need to add jshint hook to catch things like this

Yaffle:
on ES6, ES7 pages the elements are links to the MDN, not to the separate tables...

PR #417, opened 2015-01-28
Add tests for some IE bugs

zloirock:
- IE 11 `WeakMap` not support frozen objects as keys
- IE TP `Math.hypot` accepts only 2 or 3 arguments
- IE TP `Reflect.enumerate` returns not iterable iterator

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -3001,6 +3001,27 @@ exports.tests = [
         iojs:        true,
       },
     },
+    'Support frozen objects as keys': {
+      exec: function () {/*
+        var f = Object.freeze({});
+        var m = new WeakMap;
+        m.set(f, 42);
+        return m.get(f) === 42;
+      */},
+      res: {
+        _6to5:       true,
+        ejs:         true,
+        ie11tp:      true,
+        firefox11:   true,
+        chrome21dev: flag,
+        chrome36:    true,
+        safari71_8:  true,
+        ios8:        true,
+        webkit:      true,
+        node:        flag,
+        iojs:        true,
+      },
+    },
   },
 },
 {
@@ -3570,8 +3591,9 @@ exports.tests = [
         var obj = { foo: 1, bar: 2 };
         var iterator = Reflect.enumerate(obj);
 
+        var passed = Symbol.iterator in item;
         var item = iterator.next();
-        var passed = item.value === "foo" && item.done === false;
+        passed    &= item.value === "foo" && item.done === false;
         item = iterator.next();
         passed    &= item.value === "bar" && item.done === false;
         item = iterator.next();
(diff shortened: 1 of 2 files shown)


data-es6.js
```diff
@@ -3570,8 +3591,9 @@ exports.tests = [
         var obj = { foo: 1, bar: 2 };
         var iterator = Reflect.enumerate(obj);
 
+        var passed = Symbol.iterator in item;
```
webbedspace: This line can't really be added because it unnecessarily requires that the engine also support a publicly available `Symbol.iterator`, a strongly related but nonetheless distinct feature. It's still possible for an engine to support `Reflect.enumerate` without _yet_ exposing the `@@iterator` symbol as `Symbol.iterator`.

(Really, it should first check if it exposes `Symbol.iterator`, and only fail it if it both does this, and neglects to put a `Symbol.iterator`-keyed method on `Reflect.enumerate`'s returned iterator.)

zloirock: I thought about it, but currently only IE TP adds `Reflect.enumerate`. All modern browsers has `Symbol.iterator`. Problem can be only if es6-shim will add `Reflect` module without `Symbol`. Ok,  I will add check.

data-es6.js
```diff
@@ -3569,9 +3590,10 @@ exports.tests = [
       exec: function() {/*
         var obj = { foo: 1, bar: 2 };
         var iterator = Reflect.enumerate(obj);
-
+        var passed = 1
+        if(typeof Symbol != undefined && 'iterator' in Symbol)passed &= Symbol.iterator in iterator;
```
kangax: Can you please make it clean and consistent with the rest of the code — semicolon after `passed = 1`, space after `if`, `passed &=` on new line and wrapped in block.

data-es6.js
```diff
@@ -5796,6 +5801,26 @@ exports.tests = [
         res: methods[m]
       }
     };
+    obj['Math.hypot'] = {
+      exec: function(){/*
+        return Math.hypot(1) === 1 && Math.hypot(9, 12, 20) === 25
```
kangax: Perhaps let's add 0-argument case as well then? (so that we cover 0, 1, 3 (and 2 as part of last one) )

``` js
Math.hypot() === 0 && 
Math.hypot(1) === 1 &&
Math.hypot(9, 12, 20) === 25
```

data-es6.js
webbedspace: "All modern browsers"? You neglect Safari, iOS, and IE 11 (which is still the latest stable version of that browser line).

zloirock: I do not think the Safari is a modern browser :) About IE11 - I think, it will be added with `Symbol`.

zloirock:
ping

kangax:
Can't merge it. Could you please resubmit with new es6.html or just without it.

zloirock:
Ok, now will create a new PR.

PR #418, opened 2015-01-29
Add es7-shim to es7 compat table

ljharb:
https://www.npmjs.com/package/es7-shim

### DIFF
--- a/build.js
+++ b/build.js
@@ -36,7 +36,8 @@ process.nextTick(function () {
   handle(require('./data-es5'));
   var es6 = require('./data-es6');
   handle(es6);
-  handle(require('./data-es7'));
+  var es7 = require('./data-es7');
+  handle(es7);
   handle(require('./data-non-standard'));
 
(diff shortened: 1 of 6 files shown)


kangax:
Need to change data-es7.js as well, or all of this will be overwritten on the next "compile". Check out https://github.com/kangax/compat-table#editing-the-tests

ljharb:
ah, thanks, i just used the web ui. I'll clone the repo properly and fix it today.

ljharb:
@kangax so, the es7-shim doesn't auto-shim unless you run it through browserify (which thus automatically autoshims). Is there a way I can specify this behavior? Is there a reason that all polyfills aren't run through browserify already?

ljharb:
nvm, i found the array in build.js.

ljharb:
All updated!

kangax:
LGTM, thanks!

@webbedspace mind taking a look if all is good?

PR #419, opened 2015-01-30
Add tests for some IE bugs

zloirock:
- IE 11 `WeakMap` not support frozen objects as keys
- IE TP `Math.hypot` accepts only 2 or 3 arguments
- IE TP `Reflect.enumerate` returns not iterable iterator

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -2999,6 +2999,27 @@ exports.tests = [
         iojs:        true,
       },
     },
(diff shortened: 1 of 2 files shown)


kangax:
Great, just need this fixed — https://travis-ci.org/kangax/compat-table/builds/48886638#L249

zloirock:
Fixed.

PR #420, opened 2015-01-30
Fixed "symbol keys are hidden to pre-ES6 code" test

zloirock:
In Traceur symbols are not hidden from `for-in` loop, [try in browser without native `Symbol`](http://google.github.io/traceur-compiler/demo/repl.html#var%20object%20%3D%20%7B%7D%3B%0Avar%20symbol%20%3D%20Symbol%28%29%3B%0Aobject%5Bsymbol%5D%20%3D%201%3B%0A%0Afor%20%28var%20x%20in%20object%29%20alert%28x%29%3B), but passes this test.

The problem in strict comparison - Core-js and Traceur `Symbol` returns object, but `for-in` enumerates object keys as primitives.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -4857,9 +4857,10 @@ exports.tests = [
         var object = {};
         var symbol = Symbol();
         object[symbol] = 1;
-
-        for (var x in object){}
-        var passed = (x !== symbol);
+        var passed = 1;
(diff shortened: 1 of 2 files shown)


webbedspace:
This conflicted with a recent commit, so I just went and applied the fix manually. Thanks.

PR #426, opened 2015-02-05
Update Closure Compiler v20150126

teppeis:
- Closure Compiler [v20150126](https://github.com/google/closure-compiler/wiki/Releases#february-2-2015-v20150126) is released
- Compile global.__createIterableObject with each compiler if possible


data-es6.js
```diff
@@ -3981,7 +3983,9 @@ exports.tests = [
         } catch(e) {}
         return true;
       */},
-      res: temp.destructuringResults,
+      res: Object.assign({}, temp.destructuringResults, {
+        closure:     false,
```
teppeis, line 4110: current status of "destructuring - throws on null and undefined" is wrong. it's not supported.

build.js
```diff
@@ -178,12 +178,21 @@ function handle(options, compiler) {
   var skeleton = fs.readFileSync(__dirname + path.sep + options.skeleton_file, 'utf-8');
   var html = dataToHtml(skeleton, options.browsers, options.tests, options.compiler);
 
+  var createIterableObject = [];
+  try {
+    var compiled = options.compiler(
+      'global.__createIterableObject = function(a, b, c) { return function*() { yield a; yield b; yield c; }(); };'
```
webbedspace, line 197: Isn't the point of `__createIterableObject` that it should work on platforms that support either `Symbol.iterator`, or generators, as both are valid means of creating iterables? This doesn't have the `Symbol.iterator` fallback.

teppeis, line 197: @webbedspace `Symbol.iterator` fallback already exists [in HTML](https://github.com/kangax/compat-table/pull/426/files#diff-d79397253b80964bde92888b84a6e368R20).
This added code is the fallback for a compiler that has no polyfill for generator function or `Symbol.iterator` but can compile generator function.

The fallback process is:
1. Generator function polyfill (but I think any compilers cannot polyfill it in eval)
2. `Symbol.iterator` polyfill
3. Generator function compilation (new!)

I can add No.4 "`Symbol.iteraotr` compilation", but IMO it is not great important.
Because almost compilers don't compile `Symbol.iterator`, but polyfill it.

teppeis:
I'll try to cleanup and re-send PR. thanks.

PR #430, opened 2015-02-11
Added initial subclassing test categories

webbedspace:
Currently, subclassing is tested for Array, Function, RegExp, Promise, Number, String and Boolean. What remains: Map, Set, WeakMap, WeakSet, typed array stuff, Proxy, Date.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -1407,7 +1407,7 @@ exports.tests = [
         _6to5: true
       },
     },
-    'extends (author-defined classes)': {
+    'extends': {
       exec: function () {/*
         class B {}
         class C extends B {}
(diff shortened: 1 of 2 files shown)


data-es6.js
```diff
@@ -6072,6 +6039,271 @@ exports.tests = [
...
 {
+  name: 'Array is subclassable',
+  category: 'subclassing',
+  link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-array-constructor',
+  subtests: {
+    'basic functionality': {
+      exec: function () {/*
+        class C extends Array {}
+        var c = new C();
+        c.push(2,4,6);
```
kangax: Hm, `push` is a good test but it doesn't check arrays' special `[[DefineOwnProperty]]` behavior (which changes length and is what's usually considered to make arrays "magic"). I just double-checked ES6 and `push` sets length explicitly. 

What do you think about something like this instead?

``` js
class C extends Array {}
var c = new C(1,2,3);

var initialLength = c.length;

c[5] = 'x';
var lengthAfterSetting = c.length;

c.length = 1;
var typeAfterClearing = typeof c[5];

return initialLength === 3 && 
       lengthAfterSetting === 6 && 
       typeAfterClearing === 'undefined';
```

kangax:
Thanks, as always!

PR #437, opened 2015-02-15
Added feature significance weightings to supertests.

webbedspace:
These weightings are used to calculate the percentages more accurately - a supertest rated "large" (representing a major, transformative language feature) is worth 1, one rated "medium" (representing a significant feature that's less universally useful, or is primarily connected to another feature) is worth 0.5, and one rated "small" (representing a useful but subtle improvement from ES5) is worth 0.25. They are based solely on my informed judgment, and could easily be disputed.

This replaces the previous weighting system installed recently, which made every supertest equal to 1, but every annex B supertest equal to 0.2.

Also: some faint circles have been added to the supertests' `<td>` names, denoting their weighting (large, medium or small) to the reader.

### DIFF
--- a/build.js
+++ b/build.js
@@ -290,6 +290,9 @@ function dataToHtml(skeleton, browsers, tests, compiler) {
 
     var testRow = $('<tr></tr>')
       .addClass("subtests" in t ? 'supertest' : '')
+      .attr("significance",
+        t.significance === "small" ? 0.25 :
+        t.significance === "medium" ? 0.5 : 1)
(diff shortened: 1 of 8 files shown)


webbedspace:
Sample: https://raw.githack.com/webbedspace/compat-table/gh-pages/es6/index.html

kangax:
I like!

Let's just add these to legend as well? Also, probably a good idea to copy your explanation and paste to README under new ("how scores are determined" or smth like that) section.

PR #441, opened 2015-02-16
Update `es6-shim` with `Reflect` and `Object.gOPN` primitives

ljharb:
`es6-shim` now has the `Reflect` API, as well as wrapping `Object.getOwnPropertyNames` to accept primitives.

Also, the `Reflect.enumerate` test was broken, so I fixed it - you may want to check on all the other columns' values for that.

Also fixes an issue from #439 with an incomplete 6to5 → babel rename.


ljharb:
Thanks!

PR #442, opened 2015-02-17
`es5-shim` support

ljharb:
Add the `es5-shim` to the `es5` table.

I copied the structure of the HTML page from the `es6` one so they're now more consistent, and made the appropriate edits.


build.js
```diff
@@ -55,15 +59,28 @@ process.nextTick(function () {
   var traceur    = require('traceur');
   var reacttools = require('react-tools');
   var tss        = require('typescript-simple');
+  var identity   = function (x) { return x; };
```
webbedspace, line 62: Trivia: you could actually just use `String` - it also returns any string passed into it.

webbedspace:
You need to add `es5/compilers/*` to .gitignore.

ljharb:
Done :-)

build.js
ljharb, line 62: That's true if `code` is always a string, but I didn't want to assume that :-)

data-es5.js
```diff
@@ -1170,6 +1225,8 @@ exports.tests = [
     return typeof Array.prototype.every == 'function';
   },
   res: {
+    es5shim: assign({ val: true }, sparseNote),
```
webbedspace, line 1228: This `assign` isn't really necessary - footnotes are keyed by ID, and using the same object will only create one footnote.

ljharb, line 1228: I realized that afterwards - but this way, the HTML isn't arbitrarily encoded in only one of them (like in es6), and I don't have to use strings to add the note to a new one later.

PR #451, opened 2015-02-22
Supertest cells are now coloured by percentage.

webbedspace:
Preview: https://raw.githack.com/webbedspace/compat-table/gh-pages/es6/index.html

Also, in order to enable these cells to be correctly shaded when the mouse hovers over them, I replaced the multiple cell colour declarations with CSS filters. Support for these is in the latest Firefox, Chrome/Opera and Safari. As it's purely a visual garnish, a lack of IE support is not really that important.


kangax:
Nice, thanks!

PR #458, opened 2015-03-04
Changed browser selection that hide other browsers

### LINKED ISSUE #457: Can you frozen the first column?
I want to one platform as iojs that's nearly the last one, I compare the feature with the scrollbar left-to-right. 

Or hide the others when I choose one platform.

popomore:
Only show one colunm when selected

https://popomore.github.io/compat-table/es6/#iojs

Closes #457

### DIFF
--- a/master.css
+++ b/master.css
@@ -92,18 +92,18 @@ tr.category td {
     margin-right: 0.25em;
 }
 
-/* selected cell opacity */
+/* selected cell visibility */
 table.one-selected td {
-    opacity: 0.3; -ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=50)";
+    margin: 0;
 }
-table.one-selected td:first-child {
-    opacity: 1;
+table.one-selected th.platformtype,
+table.one-selected td[data-browser],
+table.one-selected th[data-browser] {
+    display: none;
 }
-table.one-selected tr.selected td {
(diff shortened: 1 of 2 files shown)


webbedspace:
This doesn't work when I click an obsolete browser's column.

popomore:
See that, I will figure it out soon.

popomore:
Fix it.

Also fix selecting an obsolete browser. Example http://kangax.github.io/compat-table/es6/#firefox17

Known issue: the count of the colspan of category has not changed when column changed.

kangax:
I like the idea, but the implementation could use some tweaking. On my wide screen, horizontally-huge cells look bad. They also don't scale well when I resize it down (or on tablets, mobile, etc. I imagine). It's also not clear to someone unfamiliar with the UI that clicking somewhere else will bring other columns back.

popomore:
How about add a compare button, platform can be chosen not only one, then change the hash like `#iojs+nodejs`.

The changed hash will trigger a new overlay show the result and given a return button.

kangax:
I like it. But what do you mean by overlay? How would that look?

popomore:
Maybe it's a `div` with large z-index cover the page.

popomore:
Can I refactor it with browserify?

kangax:
What exactly and why?

webbedspace:
Closing this because of non-response, and because I believe sticky header solutions (like #532) are more useful.

Issue #467, opened 2015-03-13
Better tests for Intl specs.


keithamus:
Following on from #371 (and #358), the Intl specs could do with improving.

The following needs to be added to get an idea of levels of support from browsers/engines:
### Intl Object
- [ ] Add full set of tests for Intl.Collator object (mimicking [test262 test suite](https://github.com/tc39/test262/tree/83b07ff15eadb141c3d6f4d236a8733b720041d2/test/intl402))
- [ ] Add full set of tests for Intl.NumberFormat object (mimicking [test262 test suite](https://github.com/tc39/test262/tree/83b07ff15eadb141c3d6f4d236a8733b720041d2/test/intl402))
- [ ] Add full set of tests for Intl.DateTimeFormat object (mimicking [test262 test suite](https://github.com/tc39/test262/tree/83b07ff15eadb141c3d6f4d236a8733b720041d2/test/intl402))
### Date Object
- [ ] Add full set of tests for Date#toLocaleString() (mimicking [test262 test suite](https://github.com/tc39/test262/tree/83b07ff15eadb141c3d6f4d236a8733b720041d2/test/built-ins/Date/prototype/toLocaleString))
- [ ] Add full set of tests for Date#toLocaleDateString() (mimicking [test262 test suite](https://github.com/tc39/test262/tree/83b07ff15eadb141c3d6f4d236a8733b720041d2/test/built-ins/Date/prototype/toLocaleDateString))
- [ ] Add full set of tests for Date#toLocaleTimeString() (mimicking [test262 test suite](https://github.com/tc39/test262/tree/83b07ff15eadb141c3d6f4d236a8733b720041d2/test/built-ins/Date/prototype/toLocaleTimeString))
### Number Object
- [ ] Add full set of tests for Number#toLocaleString() (there is no test262 suite for this - see tc39/test262#195. Maybe we should PR a set before porting them to compat-table?)
### String Object
- [ ] Add full set of tests for String#localeCompare() (mimicking [test262 test suite](https://github.com/tc39/test262/tree/83b07ff15eadb141c3d6f4d236a8733b720041d2/test/built-ins/String/prototype/localeCompare))
### Array Object
- [ ] Add full set of tests for Array#toLocaleString() (mimicking [test262 test suite](https://github.com/tc39/test262/tree/83b07ff15eadb141c3d6f4d236a8733b720041d2/test/built-ins/Array/prototype/toLocaleString))
### Object Object
[...]

PR #469, opened 2015-03-17
Notes about support Intl on server side

okuryu:
- Node v0.12: Intl support only English by default, it needs to build with full ICU data if needs all locales supported.
- io.js: Intl support is not enabled by default, it needs to build with --with-intl option if needed.

### DIFF
--- a/data-esintl.js
+++ b/data-esintl.js
@@ -124,18 +124,24 @@ exports.browsers = {
     short: 'Node .12',
     obsolete: false, // current version
     platformtype: 'engine',
+    note_id: 'only-english',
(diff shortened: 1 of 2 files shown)


keithamus:
This checks out. @okuryu - the existing tests are quite slim, and so don't give the full picture. See #467 on notes on improving them. I'll be doing it when I get an opportunity to, but if you're keen feel free to take some up :smile: 

:+1: from me @kangax

kangax:
Cool, thanks @keithamus

okuryu:
@keithamus Thanks! I didn't know about #467, I'll work to contribute for it if there are things that I can. :smiley:

PR #470, opened 2015-03-18
[ES7] Add String.prototype.at

zloirock:
[Proposal](https://github.com/mathiasbynens/String.prototype.at)

Also, enable `Object.getOwnPropertyDescriptors` for `babel` / `core-js`.

[Result](https://raw.githack.com/zloirock/compat-table/gh-pages/es7/index.html)

Not sure about other ES7 proposals which shim available in `core-js`, but this, I think, should be added.

### DIFF
--- a/data-es7.js
+++ b/data-es7.js
@@ -351,7 +351,7 @@ exports.tests = [
   */},
   res: {
     tr: false,
-    babel: false,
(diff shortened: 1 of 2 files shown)


kangax:
Thanks!

PR #475, opened 2015-03-20
Update ES6 data for IETP March flight

dstorey:


zloirock:
Why not updated `es6/index.html`?

Why you add breaking features to browser? Additional work for polyfills for detect and replace it. Only little part of standart library problems (from previous version, don't think they are fixed in current):

`Reflect.ownKeys` exist, but not works
`Reflect.enumerate` exist, but returns not iterable
all `Reflect` methods should throw error on primitive, but not in IETP
`Math.asinh` acuracy
`Math.hypot` accepts only 2 or 3 arguments
`String.raw` "lacks substituting" bug
`Set#keys` !== `Set#values`
`String#startsWith` & `String#endsWith` not throw error on regexp

data-es6.js
```diff
@@ -2708,7 +2715,6 @@ exports.tests = [
       res: {
         tr:          true,
         babel:       true,
-        ie11tp:      true,
```
webbedspace, line 2711: Actually, I did some research, and the truth is somewhere in the middle. `/./u` in IETP will wrongly select the first of the surrogate pair, but `/^.$/u` will correctly select the entire character.

abchatra, line 2711: Is it possible to share the test input for which /./u is failing?

PR #479, opened 2015-03-22
Fix the link for String.prototype.localeCompare

okuryu:
URL needs to be lower case.


Yaffle:
why?
Possibly, the spec might be changed to have `id="sec-String.prototype.localeCompare"`.

Issue #480, opened 2015-03-24
Iterator return() and throw() tests


webbedspace:
Destructuring return():

``` js
var passed = false;
var iterator = __createIterableObject('fee','fie','foe');
iterator.return = function() { passed = true; };
var [a,b] = iterator;
return passed;
```

For-of return():

``` js
var passed = false;
var iterator = __createIterableObject('fee','fie','foe');
iterator.return = function() { passed = true; };
for(var i of iterator) { if (i === 'fie') break; }
return passed;
```

For-of throw():

``` js
var passed = false;
var iterator = __createIterableObject('fee','fie','foe');
iterator.throw = function() { passed = true; };
try { for(var i of iterator) { if (i === 'fie') throw 0; } } catch(e) {}
return passed;
```

Destructuring throw()...?:

``` js
var passed = false;
var iterator = __createIterableObject('fee',undefined,'foe');
iterator.throw = function() { passed = true; };
try { var [a,b = (function(){ throw 0; }())] = iterator; } catch(e) {}
return passed;
```

zloirock:
Delegated `yield`, `Array.from`, collections. Recently wrote a part, can add it.

Issue #454, opened 2015-02-26
Add test for safe iteration closing?


zloirock:
`for-of`, destructuring, delegated yield, `Array.from`, collections constructors, `Promise` static methods [shoud call](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-iteratorclose) `return` method of iterator on abort iteration, if it's available.

PR #481, opened 2015-03-24
[ES6] Implicit iterator closing tests

zloirock:
#480 #454

Tests of implicit `.return` call. Not added tests for `Promise.all` / `Promise.race` - strange case.

Added results for traceur and babel. Missing babel results should be fixed in next version: [partial destructuring](https://github.com/babel/babel/issues/1047), [delegated yield, throw -> return](https://github.com/facebook/regenerator/issues/185).

[Result](https://raw.githack.com/zloirock/compat-table/gh-pages/es6/index.html#Implicit_iterator_closing).

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -4935,6 +4935,172 @@ exports.tests = [
   },
 },
 {
+  name: 'Implicit iterator closing',
+  category: 'misc',
+  significance: 'small',
+  link: 'http://people.mozilla.org/~jorendorff/es6-draft.html#sec-iteratorclose',
+  subtests: {
+    'for-of, break': {
+      exec: function () {/*
+        var closed = false;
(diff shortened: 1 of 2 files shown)


webbedspace:
OK, but this doesn't actually constitute a unique feature, so I'll redistribute these tests to their relevant places.

zloirock:
I thought about it, but the extraction of the iterator closing in a separate category seems more logical.

zloirock:
@webbedspace You in vain replaced `__createIterableObject(null, null, null)` to `__createIterableObject(1, 2, 3)`, now `Map` test isn't correct.

Upd: my mistake, did not look into the spec, it should throw error / close iterator too. Looks like fixed in one of latest revisions.

PR #485, opened 2015-03-28
Weakset basic functionality correction

visualjeff:


webbedspace:
Actually, the point of this test is to add the same object twice, thus confirming that this is possible and that the second time is a no-op. as per the definition of a Set. (However, since it's a WeakSet, there's no way, by definition, to confirm that the object doesn't exist twice within it, so as it is, it's a fairly anemic test compared to the standard Set's test.)

Anyway, the real issue is that unused `obj2 = {}` should be removed.

visualjeff:
I did wonder about that.

PR #487, opened 2015-03-31
Setting unstable platforms "on" by default

deltakosh:


kangax:
I've been thinking about that as well... Merging, thanks.

deltakosh:
Cool :) thanks for your quick validation

PR #488, opened 2015-04-01
renaming IE11TP to Project Spartan

deltakosh:


data-es6.js
```diff
@@ -34,7 +34,7 @@ exports.browsers = {
     full: 'Closure Compiler v20141120',
     short: 'Closure',
     obsolete: false,
-    platformtype: 'compiler',
+    atformtype: 'compiler',ei11tp
```
kangax: Typo?

data-es6.js
```diff
@@ -66,12 +66,10 @@ exports.browsers = {
...
     obsolete: false
   },
-  ie11tp: {
-    full: 'Internet Explorer',
-    short: 'IE<br>Technical<br>Preview',
-    unstable: true,
-    note_id: 'ie-experimental-flag',
-    note_html: 'Have to be enabled via "Experimental Web Platform Features" flag'
+  projectspartan: {
+    full: 'Project Spartan',
```
kangax: Does it have to be "Project Spartan"? We'll need a shorter name for at least "short" version (fitting stuff in a table is hard). Also, I'd just call test identifier "spartan" (similar to "tr", and not traceur; or "ie11" and not "internet_explorer_11")

deltakosh: Spartan is fine and I will submit a new PR when the new name will be announced :)

data-es6.js
deltakosh: clearly..Sorry about that :(

data-es6.js
```diff
@@ -66,12 +66,10 @@ exports.browsers = {
...
     obsolete: false
   },
-  ie11tp: {
-    full: 'Internet Explorer',
-    short: 'IE<br>Technical<br>Preview',
-    unstable: true,
-    note_id: 'ie-experimental-flag',
-    note_html: 'Have to be enabled via "Experimental Web Platform Features" flag'
+  spartan: {
+    full: 'Project Spartan',
```
webbedspace: The `full` name for the Trident/Chakra line of browsers has to be changed to "Internet Explorer / Project Spartan", because the "differential results" code in build.js uses the `full` name as a key to determine engine version succession - for instance, setting `ie10: true` causes IE 11 to have a true result as well, because its `full` name is the same and it appears after it in the `browsers` object (which is of course considered ordered by definition order for the purposes of `Object.keys`).

Ideally there'd be a different value, maybe called `family`, that determined this, but currently this is how it is.

deltakosh: Actually trident is not the engine of spartan

data-es6.js
```diff
@@ -420,7 +418,7 @@ exports.tests = [
         typescript:  true,
         ejs:         true,
         closure:     true,
-        ie11tp:      true,
+        spartan:      true,
```
webbedspace: If I may be so fussy... proper alignment, please?

data-es6.js
webbedspace: Well, what's important is the JS engine, and I believe it still uses Chakra (or at the very least inherits all of the functionality of Chakra, such that an IE10/11 `true` remains so for Spartan unless otherwise noted).

deltakosh: Yes that's right. I'll update it:)

data-es6.js
deltakosh: No worry I'm like you for my projects:)

kangax:
@deltakosh could you please resubmit?

deltakosh:
Did I missed something?

kangax:
@deltakosh just need a fresh rebase

PR #515, opened 2015-04-20
update es6 ejs status

toshok:
did a quick pass this morning after adding initial generator support to ejs.  found a few other things we'd forgotten to update as well.


kangax:
@toshok thanks! could you just please pay attention to indentation next time

toshok:
ah, sorry about that.  will do in the future.  did the patch in emacs and
it has its own idea of indentation :/

PR #519, opened 2015-04-23
`es7-shim` now has `Map#toJSON` and `Set#toJSON`

ljharb:


ljharb:
ping - freshly rebased :-)

PR #525, opened 2015-04-28
Change JSX implementation to jstransform@11.0.0

appsforartists:
Per https://github.com/facebook/jstransform/pull/81, `jstransform` has replaced `react-tools` as the way to access React's transpiling infrastructure.


appsforartists:
Thanks @webbedspace.  How long does it take for the page to update?

PR #526, opened 2015-04-28
Add better tests for Typed Objects

thomasfosterau:
The current test for Typed Objects tests for the existence of a global `StructType` object. In reality, `StructType` isn't a global object - it's actually a property of the global `TypedObject` object. 

This pull request fixes the current test for `StructType` and adds tests for a number of other Typed Objects features.

### DIFF
--- a/data-es7.js
+++ b/data-es7.js
@@ -236,10 +236,112 @@ exports.tests = [
   name: 'typed objects',
   category: 'proposal',
   link: 'https://github.com/dslomov-chromium/typed-objects-es7',
-  exec: function () {/*
-    return typeof StructType === "function";
-  */},
-  res: {
+  subtests : {
+    "StructType" : {
+      exec: function () {/*
+        return typeof TypedObject.StructType === "function";
+      */},
+      res: {
+      }
+    },
+    "ArrayType" : {
+      exec: function () {/*
+        return typeof TypedObject.ArrayType === "function";
+      */},
+      res: {
+      }
+    },
+    "TypedObject.storage" : {
+      exec: function () {/*
+        return typeof TypedObject.storage === "function";
+      */},
+      res: {
+      }
+    },
+    "Any type" : {
+      exec: function () {/*
+        return typeof TypedObject.Any === "function";
+      */},
+      res: {
+      }
+    },
+    "Object type" : {
+      exec: function () {/*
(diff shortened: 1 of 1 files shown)


webbedspace:
Doing some research, I believe that the `TypedObject` namespace isn't part of the proposal yet, but merely [a convenience pioneered independently by Firefox Nightly](https://esdiscuss.org/topic/typed-objects-binary-data-polyfills#content-12).

thomasfosterau:
Hmm - I see that the [structs.js](https://github.com/dherman/structs.js) polyfill uses `TypedObjects` as a namespace, and I'm not sure what Intel's Crosswalkis doing. If the test is actually going to return the right result, it seems a test for three different things is needed.

webbedspace:
Personally I feel that until the basic API of this feature is finally unambiguously expressed in the proposal, there shouldn't really be a test at all, since the feature is otherwise not consistently usable in end-user code.

thomasfosterau:
Perhaps it'd be better to treat the namespaces as a vendor-prefix - the actual API is the same and doesn't look to be particularly unstable.

webbedspace:
The current trend for misnamed features is to leave the test code as-is, mark the result as "No", but attach a footnote reading "Available in [version] as [alternative name]." (See the Array#includes test as an example). I think this is a good approach, as an off-spec name _should_ be considered falling short of usability for developers.

thomasfosterau:
That doesn't really seem right to me - it's not as though the feature is misnamed, it's just that it's within a namespace for now. 

This case is more like Web APIs being vendor prefixed while they are being standardised, not a misnamed out of date version of the spec.

kangax:
I agree that we should probably continue with established way of clarifying misnamed features via footnotes; otherwise, we run into a can of worms similar to what Chrome is doing with enabling some features via "use strict" or what browsers are doing with user-enabled flags, and so on. The bottom line is that the feature is not available as is and as specified.

@ljharb @zloirock what do you think?

zloirock:
Typed objects removed from the proposals list. All repos / gists obsolete. I don't think makes sense add it to the table.

kangax:
Haha, great. Ok, closing.

I'm guessing type values are still on the roadmap?

zloirock:
https://github.com/tc39/ecma262/issues/223

PR #532, opened 2015-05-04
Bring back sticky table header on safari

stipsan:
Firefox still don't do sticky table headers very well, but Safari do.

It makes it so much easier to read the table as the headers are always visible.

What it looks like on Safari 8.0.5 on Yosemite:

animation

### DIFF
--- a/master.css
+++ b/master.css
@@ -15,9 +15,14 @@ table {
     margin-bottom: 2em; width: 100%;
 }
 thead > tr:nth-child(2) {
+    background: white;
     position: -webkit-sticky;
+    position: sticky;
     top: 0px;
     z-index: 1;
+}
+thead > tr:nth-child(2) > th {
+    outline: 2px solid white;
 }
 th {
     font-weight: normal; text-align: center; padding: 1em 0.45em; position: relative; min-width: 40px; font-size: 13px; background: #eee; vertical-align: middle;
--- a/master.css
+++ b/master.css
@@ -14,6 +14,11 @@ body {
 table {
     margin-bottom: 2em; width: 100%;
 }
+thead > tr:nth-child(2) {
+    position: -webkit-sticky;
+    top: 0px;
+    z-index: 1;
+}
 th {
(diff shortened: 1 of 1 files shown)


kangax:
@stipsan thanks!
@webbedspace i recall something about us trying this before but then reverting back? Do you remember by any chance? Was it buggy implementation or smth else?

webbedspace:
`position:sticky` or its vendor-prefix versions don't work for `<tr>` elements in Firefox, hence I withdrew it on that basis.

I'm a bit leery of accepting this as-is because I don't like the look of the transparent cell borders - aesthetically, they should really be solid white for the duration of it being stuck. This is a bit difficult to do for a `<tr>` in CSS, admittedly.

kangax:
I agree about solid white. As for partial support — I'm ok with it. I know
that Chrome is working on it (
https://code.google.com/p/chromium/issues/detail?id=231752) and here's
Mozilla's ticket — https://bugzilla.mozilla.org/show_bug.cgi?id=975644 so
I'm guessing we'll have much wider support soon.

claudepache:
> I'm a bit leery of accepting this as-is because I don't like the look of the transparent cell borders - aesthetically, they should really be solid white for the duration of it being stuck.

For simulating solid white borders, one may hack with the outline of cells:

``` js
thead > tr:nth-child(2) > th {
    outline: 1px solid white;
}
```

There is still a problem with column headers of unstable and obsolete platforms, as their background is semi-transparent:

c

stipsan:
@claudepache I adjusted it to 2px so the bottom outline match the overall cell spacing.

The semi-transparent background is fixed.

@webbedspace here's what it looks like now:
compat

webbedspace:
The top header looks like this in Firefox, which is somewhat unpleasant.

stipsan:
@webbedspace indeed, using box-shadow instead of outline solves it in FF while looking exactly the same in Safari, Chrome etc.

There's a plugin for cross browser compat: https://github.com/filamentgroup/fixed-sticky
But it's not yet fully working with table thead elements yet.

MS is considering pos sticky for Edge, and FF is working on a fix for sticky table elements thus I vendor prefixed those too.

stipsan:
@webbedspace what do you think of it now?

PR #534, opened 2015-05-04
[ES7] Add some more correct tests

zloirock:
- `Object.getOwnPropertyDescriptors`
- `Map#toJSON`
- `Set#toJSON`
- `String#lpad`
- `String#rpad`


kangax:
Nice, thanks!

PR #538, opened 2015-05-06
[ES7] Add some more correct tests

zloirock:
- `Array#includes`
- `Object.observe`

### DIFF
--- a/data-es7.js
+++ b/data-es7.js
@@ -175,7 +175,11 @@ exports.tests = [
   link: 'https://github.com/tc39/Array.prototype.includes/blob/master/spec.md',
   category: 'draft',
   exec: function () {/*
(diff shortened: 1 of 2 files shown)


data-es7.js
```diff
@@ -159,7 +159,14 @@ exports.tests = [
   link: 'http://wiki.ecmascript.org/doku.php?id=harmony:observe',
   category: 'draft',
   exec: function () {/*
-    return typeof Object.observe === 'function';
+    var O = {x: 1};
```
kangax: Let's change `O` to something like `obj` coz it's almost confusing with `0`. Other than this, LGTM, and thanks, as always!

zloirock: Changed :)

PR #547, opened 2015-05-24
Update es7-shim SIMD results; fix a bunch of es7 tests

ljharb:
The `es7-shim` now supports `SIMD` using the official polyfill for the spec. I'm not sure where the three extra methods in `compat-table` came from though - just because Firefox has them doesn't mean they belong there :-)

The `getOwnPropertyDescriptors` test had an unnecessary dependency on `Symbol`, and the `Map#toJSON`/`Set#toJSON` tests don't account for the fact that some engines don't accept an initial argument in their `Map`/`Set` constructors - both fixed :-)

### DIFF
--- a/data-es7.js
+++ b/data-es7.js
@@ -320,7 +320,7 @@ exports.tests = [
   link: 'https://gist.github.com/WebReflection/9353781',
   category: 'strawman',
   exec: function () {/*
-    var B = Symbol('b');
+    var B = typeof Symbol === 'function' ? Symbol('b') : 'b';
     var O = Object.defineProperty({a: 1, [B]: 2}, 'c', {value: 3});
     var D = Object.getOwnPropertyDescriptors(O);
 
--- a/data-es7.js
+++ b/data-es7.js
@@ -738,7 +738,10 @@ exports.tests = [
   category: 'strawman',
   link : 'https://github.com/DavidBruant/Map-Set.prototype.toJSON',
   exec: function(){/*
-    return JSON.stringify(new Map([['a', 'b'], ['c', 'd']])) === '[["a","b"],["c","d"]]';
+    var map = new Map();
+    map.set('a', 'b');
+    map.set('c', 'd');
+    return JSON.stringify(map) === '[["a","b"],["c","d"]]';
   */},
   res: {
     babel:       true,
@@ -773,7 +776,9 @@ exports.tests = [
   category: 'strawman',
   link: 'https://github.com/DavidBruant/Map-Set.prototype.toJSON',
   exec: function(){/*
-    return JSON.stringify(new Set([1, 2, 3, 2, 1])) === '[1,2,3]';
+    var set = new Set();
+    [1, 2, 3, 2, 1].forEach(function (i) { set.add(i); });
+    return JSON.stringify(set) === '[1,2,3]';
   */},
   res: {
     babel:       true,
--- a/data-es7.js
+++ b/data-es7.js
@@ -320,8 +320,10 @@ exports.tests = [
   link: 'https://gist.github.com/WebReflection/9353781',
   category: 'strawman',
   exec: function () {/*
+    var object = {a: 1};
     var B = typeof Symbol === 'function' ? Symbol('b') : 'b';
-    var O = Object.defineProperty({a: 1, [B]: 2}, 'c', {value: 3});
+    object[B] = 2;
+    var O = Object.defineProperty(object, 'c', {value: 3});
     var D = Object.getOwnPropertyDescriptors(O);
 
     return D.a.value === 1 && D.a.enumerable === true && D.a.configurable === true && D.a.writable === true
--- a/data-es7.js
+++ b/data-es7.js
@@ -419,6 +419,7 @@ exports.tests = [
       */},
       res: {
         firefox39:   true,
+        es7shim: true,
       }
     },
     'float32x4' : {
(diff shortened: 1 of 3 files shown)


ljharb:
(Actually I'm getting confirmation on the `Array#includes` issue, please don't merge just yet)

ljharb:
I removed the `Array#includes` change, looks like it's correct as-is.

zloirock:
> both fixed :-)

I think, ES7 tests should be called In ES6 environment.

ljharb:
@zloirock Then you're not testing the thing you're testing, you're testing other things. If you want to do that, then core-js wouldn't be able to pass any tests that depend on `Symbol`, since `typeof Symbol() !== 'symbol'` :-p

zloirock:
@ljharb `Symbol` missed in ES6 environment? :)

ljharb:
Safari doesn't have `Symbol`s. Any tests that automatically fail in any modern browser are invalid. Also, as the compatibility table itself indicates, there's no such thing as "an ES6 environment" - there's still no such thing as "an ES5 environment", even, I just haven't gotten around to adding conformance tests to that page :-)

All `Object.getOwnPropertyDescriptors` needs to ensure is that it returns an object map of the same results that `Object.getOwnPropertyDescriptor` would return for each property. The existence of `Symbol` is orthogonal to that.

zloirock:
BTW, `core-js` passes tests w/o `typeof Symbol() !== 'symbol'` or something else, browsers hasn't yet this features, it's only `es7-shim` problem. Also - computed property in `Object.getOwnPropertyDescriptors` test :)

ljharb:
ah, good call. I'll fix that too.

ljharb:
Updated

data-es7.js
```diff
@@ -320,8 +320,10 @@ exports.tests = [
   link: 'https://gist.github.com/WebReflection/9353781',
   category: 'strawman',
   exec: function () {/*
-    var B = Symbol('b');
-    var O = Object.defineProperty({a: 1, [B]: 2}, 'c', {value: 3});
+    var object = {a: 1};
+    var B = typeof Symbol === 'function' ? Symbol('b') : 'b';
+    object[B] = 2;
+    var O = Object.defineProperty(object, 'c', {value: 3});
```
zloirock, line 326: `O === object`, why produce extra variables? :)

data-es7.js
```diff
@@ -738,7 +763,10 @@ exports.tests = [
   category: 'strawman',
   link : 'https://github.com/DavidBruant/Map-Set.prototype.toJSON',
   exec: function(){/*
-    return JSON.stringify(new Map([['a', 'b'], ['c', 'd']])) === '[["a","b"],["c","d"]]';
+    var map = new Map();
+    map.set('a', 'b');
+    map.set('c', 'd');
+    return JSON.stringify(map) === '[["a","b"],["c","d"]]';
```
zloirock, line 769: You really think any browser will add `{Map, Set}#toJSON` w/o support initialization from iterable? :) Even such test will not work in IE9. Simply add your `es6-shim` as test dependency for `es7-shim`.

ljharb, line 769: It's not about that, since the compatibility table is testing polyfills and shims too. Since `Map`/`Set` _has_ been shipped such that it does not support initializations from iterables, the test should not break in those browsers.

As for adding `es6-shim` as a dependency for the `es7-shim` for the compatibility table, that's not a bad approach, but I'm not sure that's warranted. There's zero reason to keep these tests compact.

data-es7.js
ljharb, line 326: That's what the test was already doing. If you think it makes a difference, I could definitely remove `var O =`, but I don't think it matters.

data-es7.js
webbedspace, line 769: My personal policy for ES7 tests is this: some platforms are implementing ES7 features alongside ES6 - consider that Chrome supports `Object.observe` but not, say, destructuring. Therefore, assumptions about the native ES6 implementation must not be made. Continue to write tests using mainly ES3, with only the minimum of ES5 and ES6 features deployed where strictly necessary, e.g. `JSON.stringify` for a `toJSON` test, or very basic class syntax for the class properties tests.

ljharb, line 769: Thank you, that sounds like exactly the right philosophy. Let me know if there's anything in this PR that needs adjustment, but it seems like it's good to go :-)

ljharb:
@kangax @webbedspace does this need any changes, or could it be merged?

data-es7.js
```diff
@@ -773,7 +801,9 @@ exports.tests = [
   category: 'strawman',
   link: 'https://github.com/DavidBruant/Map-Set.prototype.toJSON',
   exec: function(){/*
-    return JSON.stringify(new Set([1, 2, 3, 2, 1])) === '[1,2,3]';
+    var set = new Set();
+    [1, 2, 3, 2, 1].forEach(function (i) { set.add(i); });
+    return JSON.stringify(set) === '[1,2,3]';
```
webbedspace, line 806: What I'd like to see in a `stringify` test is whether the contents are _recursively_ serialised - for instance, whether an object inside a Set is serialised as well. But, it's not really that important.

ljharb, line 806: To be honest, that's up to the `JSON.stringify` implementation calling `toJSON` - the responsibility of `toJSON` method is solely to return something that `JSON.stringify` can serialize. I could definitely make a set that contains a set, but I don't think that would actually test much.

PR #548, opened 2015-05-24
[core-js] set Object.{values, entries} results

zloirock:
As I see, are added tests for `Object.values` and `Object.entries`. `core-js` has long supported them.

### DIFF
--- a/data-es7.js
+++ b/data-es7.js
@@ -196,6 +196,7 @@ exports.tests = [
     return v instanceof Array && v + '' === "foo,bar,baz";
   */},
   res: {
+    babel:       true,
   }
 },
 {
@@ -213,6 +214,7 @@ exports.tests = [
       && e.length === 3;
   */},
   res: {
+    babel:       true,
   }
 },
 {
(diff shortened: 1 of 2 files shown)


ljharb:
They still don't have a strawman which means they're not even stage 0 yet, so I'm not sure why they're in the table at all?

zloirock:
For this reason I did not add them to the table before. But the tests here... :) In any case, I'm not sure the status of the many features. For example, IIRC, the development of parallel JS has been stopped, or I'm wrong?

ljharb:
My general sense is that anything that's still on the old wiki, and not on github, has no champion and is unlikely to continue in the short term. I'm not certain about parallel JS, or "string padding" for example, and I'm pretty certain that comprehensions are completely dead.

zloirock:
About string padding see here https://github.com/zloirock/core-js/pull/61#issuecomment-97835659 Regarding the comprehensions - IIRC recently seen several speeches were still is mentioned that they will be in ES7, but nothing concrete.

webbedspace:
I pinged @rwaldron about `entries` & `values` back in February and got a [slightly vague answer](https://twitter.com/webbedspace/status/568977414824419331) that didn't seem to say they were off the table yet. Maybe the sitch has changed outside of meeting notes?

ljharb:
I'll get an update at this week's meeting.

zloirock:
[These comments](https://esdiscuss.org/topic/providing-object-iterators-beyond-just-object-keys#content-3) is enough? Anyway, this discussion applies only to the test rather than `core-js` support and this PR.

ljharb:
imo, anything not on the table listed on https://github.com/tc39/ecma262 should not be considered actually on the table for future spec inclusion.

I agree with @zloirock though, this discussion about whether the features should be in the table at all is unrelated to whether babel/core-js supports them.

rwaldron:
@ljharb 

> imo, anything not on the table listed on https://github.com/tc39/ecma262 should not be considered actually on the table for future spec inclusion.

Well, certainly not for compat-table's needs anyway.

PR #550, opened 2015-05-28
Update ES7 proposal processes for individual tests; add pre-strawman

ljharb:
- Update ES7 proposal processes for individual tests
- Add "pre-strawman" for proposals that are not listed on https://github.com/tc39/ecma262/blob/master/stage0.md or https://github.com/tc39/ecma262/blob/master/README.md (per https://github.com/kangax/compat-table/pull/548#issuecomment-106106401)

Anything not yet on the `ecma262` repo shouldn't be considered "required", since it's nowhere close to being included in the spec yet.

### DIFF
--- a/build.js
+++ b/build.js
@@ -372,7 +372,10 @@ function dataToHtml(skeleton, browsers, tests, compiler) {
           !browsers[browserId].needs_annex_b &&
           isOptional(t.category)
       ) {
-        cell.attr('title', "This feature is optional on non-browser platforms.");
+        var msg = t.category === 'annex b'
+          ? 'This feature is optional on non-browser platforms.'
+          : 'This feature is optional on all platforms.';
+        cell.attr('title', msg);
         cell.addClass("not-applicable");
       }
 
--- a/build.js
+++ b/build.js
@@ -31,6 +31,10 @@ var child_process = require('child_process');
 
 var useCompilers = String(process.argv[2]).toLowerCase() === "compilers";
 
+var isOptional = function isOptional(category) {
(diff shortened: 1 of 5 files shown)


webbedspace:
One problem: "annex b" currently produces cells bearing the message "Optional on non-browser platforms" for only non-browser platforms, so extending it to mean "pre-stage-0" is a bit dubious, given that arguably pre-stage-0 should be optional for all platforms, including browsers.

ljharb:
Ah, true. I'll improve that real quick :-)

ljharb:
@webbedspace updated :-D

build.js
```diff
@@ -31,6 +31,10 @@ var child_process = require('child_process');
 
 var useCompilers = String(process.argv[2]).toLowerCase() === "compilers";
 
+var isOptional = function isOptional(category) {
```
kangax, line 34: why not just `function isOptional() { ... }`? :)

ljharb, line 34: Just to avoid hoisting - if you have a strong preference I'll be happy to change it to a declaration.

kangax, line 34: I just always prefer function declarations unless in blocks. Not a big deal, was just curious. LGTM, merging.

ljharb:
Thanks!

zloirock:
Additional category for pre-strawman - fine, but with current background I don't see where `Yes`, where `No` =\

webbedspace:
I pushed a change now.

zloirock:
@webbedspace thanks :)

Issue #551, opened 2015-06-02
Time zone detection


mattjohnsonpint:
The ECMAScript Internationalization API supports the following:

```
Intl.DateTimeFormat().resolvedOptions().timeZone
```
- Some browsers (such as Chrome) implement the spec completely, and return the IANA time zone id, such as `"America/Los_Angeles"`.
- Some browsers return `undefined`, as they don't implement the time zone portion of the spec.
- Some browsers do not have `Intl` at all.

Please add a line to the table indicating which browsers will return a time zone id.  Thanks.

kangax:
A PR would be super welcome!

mattjohnsonpint:
Completed with #552.

PR #552, opened 2015-06-02
Add test for time zone detection

mattjohnsonpint:
Adds a test for time zone detection, per #551.

### DIFF
--- a/data-esintl.js
+++ b/data-esintl.js
@@ -912,6 +912,19 @@ exports.tests = [
         iojs1_1: false,
       },
     },
+    'resolvedOptions().timeZone defaults to the host environment': {
+      link: 'http://www.ecma-international.org/ecma-402/1.0/#sec-6.4',
+      exec: function(){/*
+        try {
+          var tz = Intl.DateTimeFormat().resolvedOptions().timeZone;
+          return tz !== undefined && tz.length > 0;
+        } catch(e) {
+          return false;
+        }
+      */},
+      res: {
+      },
+    },
   },
 },
 {


mattjohnsonpint:
@kangax - Do you generate the test results from somewhere?  Or do you run the tests manually and fill in the results?

webbedspace:
Results are run and entered into the `data-*.js` files manually (along with any footnotes necessary to explain any anomalies).

data-esintl.js
```diff
@@ -912,6 +912,19 @@ exports.tests = [
         iojs1_1: false,
       },
     },
+    'resolvedOptions().timeZone defaults to the host environment': {
+      link: 'http://www.ecma-international.org/ecma-402/1.0/#sec-6.4',
+      exec: function(){/*
+        try {
```
webbedspace, line 918: This `try` block is not required - see README.md.

mattjohnsonpint:
Sorry, but I'm confused still.  Do you have a test environment for all these different browsers and versions? Or do you expect me to add the test results to this PR?  I don't have access to a test bed with so many configurations already set up.

kangax:
We don't have a dedicated test environment -- just a bunch of browsers locally which we check manually (at least that's how I do it, not sure about @webbedspace, @ljharb, and @zloirock)

I'm actually not sure how to deal with PR's without tests. Maybe we should still merge them and then fill in results at some point in the future?

mattjohnsonpint:
I've not used them before, but I know there are services that provide this functionality.  [BrowserStack says it's free for OSS projects](https://www.browserstack.com/javascript-testing-api).  Maybe look into that?

Issue #553, opened 2015-06-03
Add a test for the the correct [[Prototype]] of bound functions


claudepache:
ES6 introduces a breaking change in `Function.prototype.bind`.

In ES5, the [[Prototype]] of `f.bind(null)` was `Function.prototype`. In ES6, it is `f.[[Prototype]]`.

None of the three browsers that I've tested implements the new semantics.

Reference, steps 2 and 8 of [BoundFunctionCreate()](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-boundfunctioncreate) 
Justification of the change: [es:bug3829](https://bugs.ecmascript.org/show_bug.cgi?id=3819#c0)

Here is a proposed test:

``` javascript
function getProtoBound(proto) {
    var f = Object.setPrototypeOf(function() { }, proto)
    var boundF = Function.prototype.bind.call(f, null)
    return Object.getPrototypeOf(boundF)
}

var success = [ Function.prototype, function(){}, [], null ].every(function(proto) {
    return getProtoBound(proto) === proto
})

return success
```

kangax:
Hm, interesting. `getProtoBound` looks reasonable to me. As far as testing multiple [[Prototype]] values — I'd be OK with just one (say, `function(){}`) to see that behavior is implemented, but I guess the more the better (just in case).

kangax:
@claudepache if you could send a PR that would be great or one of us will get to it sometime soon

PR #555, opened 2015-06-03
adding a test for prototype of bound functions

claudepache:
Adding a test for the new initial value of the [[Prototype]] of bound functions, as discussed in Issue #553.

Also, add negative results for tested browsers (didn't test Edge).

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -2915,12 +2915,16 @@ exports.tests = [
 },
 {
   name: 'prototype of bound functions',
-  category: 'functions',
+  category: 'misc',
   significance: 'small',
   link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-boundfunctioncreate',
   exec: function () {/*
     function getProtoBound(proto) {
-      var f = Object.setPrototypeOf(function() { }, proto)
+      var f = function () { }
+      if (Object.setPrototypeOf)
+        Object.setPrototypeOf(f, proto)
+      else
+        f.__proto__ = proto
       var boundF = Function.prototype.bind.call(f, null)
       return Object.getPrototypeOf(boundF)
     }
--- a/data-es6.js
+++ b/data-es6.js
@@ -2914,6 +2914,30 @@ exports.tests = [
   },
 },
 {
+  name: 'prototype of bound functions',
+  category: 'functions',
+  significance: 'small',
+  link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-boundfunctioncreate',
+  exec: function () {/*
+    function getProtoBound(proto) {
+      var f = Object.setPrototypeOf(function() { }, proto)
+      var boundF = Function.prototype.bind.call(f, null)
+      return Object.getPrototypeOf(boundF)
+    }
+
+    return [ Function.prototype, function(){}, [], null ].every(function(proto) {
+      return getProtoBound(proto) === proto
+    })
+  */},
+  res: {
+    ie10:        false,
+    firefox11:   false,
+    chrome:      false,
+    safari51:    false,
+    webkit:      false,
+  }
+},
+{
   name: 'octal and binary literals',
   category: 'syntax',
   significance: 'small',


kangax:
@bterlson Brian, can you give result in Edge?

webbedspace:
IMHO this should be categorised in "misc", where most of the other "ES5 compatibility breaks" are stored. ("Block-level function declaration (in strict mode)" isn't really a compatibility break since it wasn't available at all in ES5). Also unsure whether it should be its own test, or just stashed as a subtest of "miscellaneous".

data-es6.js
```diff
@@ -2914,6 +2914,30 @@ exports.tests = [
   },
 },
 {
+  name: 'prototype of bound functions',
+  category: 'functions',
+  significance: 'small',
+  link: 'https://people.mozilla.org/~jorendorff/es6-draft.html#sec-boundfunctioncreate',
+  exec: function () {/*
+    function getProtoBound(proto) {
+      var f = Object.setPrototypeOf(function() { }, proto)
```
ljharb: fwiw, this test will fail in any engine without `Object.setPrototypeOf` - is it worth creating a helper polyfill like the one in es6-shim, so that browsers with `__proto__` support but not `setPrototypeOf` can still pass this test (when testing shims, for example)?

data-es6.js
```diff
@@ -2914,6 +2914,30 @@ exports.tests = [
...
+      var boundF = Function.prototype.bind.call(f, null)
+      return Object.getPrototypeOf(boundF)
+    }
+
+    return [ Function.prototype, function(){}, [], null ].every(function(proto) {
```
ljharb, line 2932: It'd be nice, only in browsers that support arrow functions, to test an arrow function here too (and similarly, a generator function)

data-es6.js
claudepache: > fwiw, this test will fail in any engine without Object.setPrototypeOf - is it worth creating a helper polyfill like the one in es6-shim, so that browsers with **proto** support but not setPrototypeOf can still pass this test (when testing shims, for example)?

I don't think there is any engine that doesn't support setPrototypeOf but supports the new prototype initialisation of bound functions (which is a very recent change in the spec); therefore I didn’t care.

ljharb: Right, but these tests need to run in older engines that have shimmed behavior. Since `Function#bind` is shimmable, the `es6-shim` will eventually correct this behavior, and would need to be able to pass in IE 9. Otherwise the test will only pass in browsers, never with shims/transpilers.

claudepache: My guess is that `es6-shim` does already shim `Object.setPrototypeOf` now when possible, and won't be able to correct `Function#bind` otherwise. Unless you have a way to correct Function#bind without mutating the prototype?

data-es6.js
claudepache, line 2932: It seems to me that the utility is not worth the trouble of using `try/eval/catch` for the sake of older engines. (I do probably already too much by testing 4 possible prototypes.)

claudepache:
> IMHO this should be categorised in "misc"

I think you are right. Are there opinions about whether to put it as a subtest of miscellaneous or not?

data-es6.js
ljharb: A ha, that's a fair point.

I suppose the only reason to care is if/when a browser fixes `bind` before implementing `setPrototypeOf`.

data-es6.js
ljharb, line 2932: `try`/`Function`/`catch` would be better, but you're right :-)

data-es6.js
webbedspace: My take on this is that one should make as few assumptions about what order platforms implement ES6 features as possible. So, I think this should, in the absence of `Object.setPrototypeOf`, fallback to the `__proto__` setter.

data-es6.js
webbedspace, line 2932: In my opinion, these should be tested, but included as separate subtests of this greater supertest (with the current test being moved to a "basic support" subtest).

claudepache, line 2932: I fear that adding subtests for other types of functions (arrows, generators, subclasses, etc.) would add more distraction than useful information. Unless the tested implementation introduces a strange bug in its modified version of `.bind()`, the result of the subtest for `.bind` on generators would just be the intersection of the result of the base test for `.bind()` and the one for the existence of generators; and a negative result of that subtest could only be interpreted correctly when compared with the existence of generators.

kangax, line 2932: I also think breaking those into subtests would be an overkill. Seems unlikely that one would work and not the other.

webbedspace, line 2932: > Unless the tested implementation introduces a strange bug in its modified version of .bind()

Well, stranger bugs have happened. Consider the number of "instances of iterables" tests I've added to iterator consumer features - if they implemented the spec correctly, retrieving the `[Symbol.iterator]` method from the prototype chain should be no different than retrieving it directly. But, you can never be too sure. (In fact, I'm starting to wonder if I should split them all up into "generator instances" and "plain-object iterables" - [look at this in FF38](http://i.imgur.com/yGkJYPu.png)).

kangax, line 2932: @webbedspace I trust your judgement then. If you think it's possible, let's do it that way.

claudepache, line 2932: The issue with those subtests is that they may fail to give the wanted answer in a readable way.

As a developper, I am interested in: Does the implementation implements the correct semantics for bound functions, whatever their target function is?

Now, if I see the base test succeeding, but a subtest for some type of functions failing, I don't know the answer to my simple question, because I don't know if the subtest fails because of a bug in the implementation of `.bind`, or just because the implementation does not support that type of function.

(Note that there is currently a similar issue with the three last subtests of "__proto__ in object literals".)

I see two solutions, here:
- force a given subtest to pass if the implementation does not support the type of function that the subtest is testing; or,
- in addition to the "Yes"/"No" answers, add the possibility of a "N/A" answer.

ljharb, line 2932: I would love all tests to provide an N/A answer when the thing they depend on (as opposed to the thing itself) is simply not supported yet in the engine. An alternative here would be for the subtest to pass - because arrow functions that aren't present don't fail, for example.

webbedspace, line 2932: I don't see the necessity of the distinction. If you want to know "does the implementation implement the correct semantics for bound functions" then you need to assume generators exist, because the correct semantics for bound functions _require the existence of generators_. If generator functions do not exist, then any semantics involving them are obviously not implemented. (If you instead want to know "does the implementation implement the correct semantics for bound _normal_ functions" then the specific subtest provides complete information.)

I also don't see the necessity of the highly misleading "N/A" answer. If the bound generator functions test says "No", then you can find out whether basic generator support exists by checking the basic generator function test result. Either way, it simply tells the truth: no, the platform does not support feature X used with feature Y. Remember that in order to be ES6 compliant, _the entire feature set_, barring Annex B where noted, must be supported - not implementing every feature that intersects with a feature is leaving that feature incomplete.

data-es6.js
```diff
@@ -2914,6 +2914,34 @@ exports.tests = [
...
+      return getProtoBound(proto) === proto
+    })
+  */},
+  res: {
+    ie10:        false,
+    firefox11:   false,
+    chrome:      false,
+    safari51:    false,
+    webkit:      false,
```
webbedspace, line 2941: Just a note: every platform is assumed `false` unless `true` or something else is given. Therefore, listing any `false` values is redundant.

data-es6.js
ljharb, line 2932: I can see the argument for all of them. I'd say that if an engine doesn't implement, say, generators, then that doesn't mean that "all of the features that intersect with generators" either work, or don't work - it's simply not applicable.

PR #557, opened 2015-06-10
Updated TypeScript compatibility for 1.5 with core-js

rbuckton:
Updated compatibility results for TypeScript 1.5 in tandem with the core-js polyfill.


data-es6.js
```diff
@@ -6,6 +6,24 @@ var temp = {};
 var flag = "flagged";
 var fallthrough = "needs-polyfill-or-native";
 
+var typescript = {
+    compiler: { 
+        val: true, 
+        note_id: "typescript-compiler-or-native", 
+        note_html: "This feature is supported through a compiler error, or when the <code>--target ES6</code> flag is set and a native ES6 host is used."
```
webbedspace: "supported through a compiler error"...? We don't need that level of detail in these descriptions - it's assumed that for any compiler platform, a test that requires an error thrown via `eval()` should represent producing an error when receiving such a line at compile-time.

webbedspace: Actually, let me get this clear: TS itself supports this error _unless_ `--target ES6` is set, in which case it just falls through to the native platform? Hmm. This seems like a counterintuitive notion of "support", and I'd be inclined to change the wording to state it exactly as I have.

zloirock:
Cool :)

data-es6.js
```diff
@@ -5948,6 +5979,7 @@ exports.tests = [
       */},
       res: {
         babel:       { val: false, note_id: "name-configurable", },
+        typescript:  typescript.corejs,
```
zloirock: This and some other tests from this category will work in environment w/o native `Function#name`, but not in actual environments with native. Run tests also in Chrome and FF.

data-es6.js
rbuckton: We would report the static error in both cases during compilation, but you would only get a runtime error if the host is ES6. If you don't need the level of detail I'm fine with just reporting `true`, which is what other compilers/transpilers seem to be doing.

data-es6.js
rbuckton: Fixed.

kangax:
Hm, now that we'll be adding TypeScript + polyfill (in addition to Babel + polyfill), it makes me remember a feature of combining engines that I've been meaning to add for a long time. For example, to see support of Firefox + Babel vs. Edge + Traceur. I'm not yet sure exactly how to make it work UI-wise, but this way we can have polyfill as a separate column as well! Then you'd be able to select, say, just Babel or Babel + core.js or Babel + core.js + Firefox. Granted, browser + compiler combo seems slightly less useful (it's more of a curiosity kind of thing) than compiler + polyfill, but few people asked for it and I think it could make sense to have that.

rbuckton:
In general what might be best is a way to delineate features that require runtime support (e.g. those that can be polyfilled) vs. features that use or derive from new syntactic forms (e.g. those that can be transpiled), though there may be cases where there is overlap.

rbuckton:
@kangax, is there an issue for that currently? I don't anticipate something like that would need to be added to this PR, but it would be worthwhile to capture this part of the discussion in a separate issue.

ljharb:
I would _love_ that, because then I could have a column for each browser + the relevant es*-shim :-)

webbedspace:
I split up a few tests here and there today, so rebase s'il vous plaît.

kangax:
> @kangax, is there an issue for that currently? I don't anticipate something like that would need to be added to this PR, but it would be worthwhile to capture this part of the discussion in a separate issue.

Hm, don't think we have it. Please create and we can continue conversation there.

rbuckton:
I've created #565 to continue that discussion.

rbuckton:
@kangax I've updated my fork with the recent changes.

rbuckton:
@kangax, @webbedspace: Is there anything else you need me to look into, or can this PR now be merged?

kangax:
+1 from me!

kangax:
We'll obviously be doing separation (I talked about) later.

webbedspace:
Sorry, I've been busy recently and haven't been looking at this. Just one more rebase, if you please, and I'll let it ride.

rbuckton:
@webbedspace, Just updated. Thanks!

PR #558, opened 2015-06-11
Update WebKit to r185450

Constellation:
- `RegExp.prototype.flags` is implemented.
- Update the parser: no escaped reserved words as identifiers.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -311,7 +311,7 @@ exports.browsers = {
     obsolete: false
   },
   webkit: {
-    full: 'WebKit r184902',
+    full: 'WebKit r185450',
     short: 'WK',
     unstable: true,
   },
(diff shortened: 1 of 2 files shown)


ljharb:
👍 yay!

ljharb:
This also turns `global`, `multiline`, and `ignoreCase` into accessors, instead of own properties, which is an ES6 change. Perhaps we should add separate `RegExp` compatibility tests for these?

kangax:
Pretty obscure change... You think it's worth it? :)

ljharb:
@kangax without it, `RegExp` can't possibly be truly subclassable, so I'd say it's pretty important :-)

kangax:
@ljharb ahha, fair enough :) well, let's do it then.

PR #559, opened 2015-06-13
Fix some core-js results and remove babel native constructors subclassing

zloirock:
- Added `core-js` results, lost after separation well-known symbols tests:
  - `Symbol.species` getters in all required constructors
  - `Symbol.toStringTag` in `Math` and `JSON`.
- Fixed some other `core-js` results:
  - `core-js` should pass current `Reflect.ownKeys, symbol order` test in all engines.
  - `(new Function).name === "anonymous"` will work in environments w/o native `Function#name` or, for example, in FF, but not in actual V8
- Also, special case for subclassing native constructors was removed from `babel`

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -5996,6 +5996,7 @@ exports.tests = [
           (function(){}).name === '';
       */},
       res: (temp.legacyFunctionNameResults = {
+        babel:       true,
         ejs:         true,
         firefox11:   true,
         edge:        true,
@@ -6021,7 +6022,6 @@ exports.tests = [
         return (new Function).name === "anonymous";
       */},
       res: {
-        babel:       true,
         firefox11:   true,
         edge:        true,
         safari51:    true,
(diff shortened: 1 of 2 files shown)


webbedspace:
Thanks.

data-es6.js
```diff
@@ -4979,7 +4983,7 @@ exports.tests = [
         return result[l-3] === sym1 && result[l-2] === sym2 && result[l-1] === sym3;
       */},
       res: {
-        babel:       { val: false, note_id: "forin-order", note_html: "This uses native for-in enumeration order, rather than the correct order." },
```
ljharb, line 4982: Won't removing this `note_html` without adding it back in elsewhere remove the note for other `forin-order` notes?

zloirock, line 4982: `note_html` for `forin-order` also contains in previous test.

ljharb, line 4982: 👍 thanks for confirming

PR #566, opened 2015-06-16
[ES7] Remove `/` from `RegExp.escape` escaped symbols

zloirock:
Initially, this symbol was present in polyfill from proposal repo and I thought @benjamingr forgot add it to proposal text. See discussion https://github.com/benjamingr/RexExp.escape/issues/12.

### DIFF
--- a/data-es7.js
+++ b/data-es7.js
@@ -878,7 +878,7 @@ exports.tests = [
 },
 {
   name: 'RegExp.escape',
-  category: 'pre-strawman',
+  category: 'strawman',
   link: 'https://github.com/benjamingr/RexExp.escape',
   exec: function(){/*
     return RegExp.escape('Hello, \\^$*+?.()|[]{}!') === 'Hello, \\\\\\^\\$\\*\\+\\?\\.\\(\\)\\|\\[\\]\\{\\}!';
--- a/data-es7.js
+++ b/data-es7.js
@@ -881,7 +881,7 @@ exports.tests = [
   category: 'pre-strawman',
   link: 'https://github.com/benjamingr/RexExp.escape',
   exec: function(){/*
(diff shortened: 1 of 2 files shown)


benjamingr:
I did initially deliberately include it in the polyfill and considered including it in the proposal text.

After empirical research that stated that escaping `"/"` does not actually affect a pattern people use. It was included for people `eval`ing a RegExp instead of using the constructor. However, it turns out usages of this (in the 10k most dependent on NPM packages, 10k most starred NPM packages, tutorials, Stack Overflow, and top websites according to Alexa) are extremely rare. There is more data as well as methodology discussion in the proposal repository.

zloirock:
Also, this proposal added to stage 0 list, https://github.com/tc39/ecma262/blob/master/stage0.md

ljharb:
We should probably also test that the input is stringified - ie, `RegExp.escape({ toString: function () { return str; } }) === RegExp.escape(str)` for any string.

zloirock:
I don't see something like this in tests for other methods, working with strings.

ljharb:
That simply means that the compatibility table tests are incomplete. We may decide it's not worth testing that, but I'd say that if any existing implementations _did_ fail to stringify correctly, that's a convincing reason to test it. What does `corejs` do here?

zloirock:
> What does corejs do here?

Converts object to string, what else? :)

The same useless tests are not necessary until some environment hasn't buggy behavior.

ljharb:
That's a fair point. My only remaining thought is that we wouldn't necessarily know about the buggy behavior unless we have a test asserting it, but I don't think it needs to block this PR of course.

ljharb:
I'll merge this shortly - `es7-shim` now includes this as well, so if you'd mind updating that it'd be appreciated! If not, I'll take care of it after merging, nbd

update: sorry, this will now need a rebase :-/

PR #568, opened 2015-06-23
Tests for bound functions: incorrect use of `isPrototypeOf`

claudepache:
The expression `proto.isPrototypeOf(boundF)` will throw when `Object.prototype` is not in the prototype chain of `proto`.

Anyway, `proto.isPrototypeOf(boundF)` is not what is wanted here, because it checks whether `proto` is somewhere in the prototype chain of `boundF`, not whether `proto` is the direct prototype of `boundF`.

Replacing with the correct expression: `Object.getPrototypeOf(boundF) === proto`.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -3011,7 +3011,7 @@ exports.tests = [
               f.__proto__ = proto;
             } 
             var boundF = Function.prototype.bind.call(f, null);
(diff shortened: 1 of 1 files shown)


ljharb:
Alternatively, `Object.prototype.isPrototypeOf.call(boundF, proto)` would work, but this is better :-)

PR #569, opened 2015-06-25
subtest "prototype of bound functions / subclasses": corrections

claudepache:
Two mistakes corrected in subtest "prototype of bound functions / subclasses":
- reference to `proto` instead of `superclass`;
- `class extends null {}` does not get `null` as its [[Prototype]]; therefore, remove that case.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -3067,11 +3067,10 @@ exports.tests = [
               }
             }
             var boundF = Function.prototype.bind.call(C, null);
-            return Object.getPrototypeOf(boundF) === proto;
+            return Object.getPrototypeOf(boundF) === superclass;
           }
           return correctProtoBound(function(){})
-            && correctProtoBound(Array)
-            && correctProtoBound(null);
+            && correctProtoBound(Array);
       */},
       res: {
       },
--- a/data-es6.js
+++ b/data-es6.js
@@ -3067,10 +3067,11 @@ exports.tests = [
               }
             }
             var boundF = Function.prototype.bind.call(C, null);
-            return Object.getPrototypeOf(boundF) === superclass;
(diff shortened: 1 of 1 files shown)


claudepache:
In general, `class extends superclass {}` has its [[Prototype]] equal to `superclass` when `superclass` is a constructor. But `class extends null {}` has its [[Prototype]] equal to `Function.prototype` instead of `null`. Therefore the test `correctProtoBound(superclass)` is invalid when `superclass` is `null`.

If we want to test the case `superclass = null`, we could replace the line:

``` js
Object.getPrototypeOf(boundF) === superclass
```

with:

``` js
Object.getPrototypeOf(boundF) === superclass != null ? superclass : Function.prototype
```

or:

``` js
Object.getPrototypeOf(boundF) === Object.getPrototypeOf(C)
```

data-es6.js
```diff
@@ -3067,11 +3067,10 @@ exports.tests = [
               }
             }
             var boundF = Function.prototype.bind.call(C, null);
-            return Object.getPrototypeOf(boundF) === proto;
+            return Object.getPrototypeOf(boundF) === superclass;
```
ljharb: This is a good fix, and was clearly a typo.

data-es6.js
```diff
@@ -3067,11 +3067,10 @@ exports.tests = [
...
           }
           return correctProtoBound(function(){})
-            && correctProtoBound(Array)
-            && correctProtoBound(null);
+            && correctProtoBound(Array);
```
ljharb: We do still need to check for a `null` proto. I think your suggestion of `Object.getPrototypeOf(boundF) === Object.getPrototypeOf(C)` is better than removing this test.

PR #570, opened 2015-06-26
Update WebKit to r185990

Constellation:
- destructuring rest element accepts iterables
- arrow function parsing functionality is supported


kangax:
\o/

PR #572, opened 2015-06-28
Added tests for collections, promises and function length

shvaikalesh:
1. `@@species` are missing in `WeakMap` and `WeakSet` constructors. Added the test to check for that: c48968a.
2. Some implementations (V8 prior to 4.3 and some polyfills) do not support iterables in `Promise.all` and `Promise.race`. Added tests to check that: 75bb18b.
3. In ES6 version of specification, contrary to ES5, `length` property of `function` instances is `configurable`. Added test for that: 091df00.
4. In early versions of ES6 spec weak collections prototypes had `clear` methods that were later removed. Added tests for their absence: f08d8f5.
5. Firefox 36 throws on `null` in constructors (5c1d6f3) and does not invoke monkey-patched `set` and `add` methods (56ee56c).

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -8735,6 +8735,28 @@ exports.tests = [
       res: {
       },
     },
+    'function "length" is configurable': {
+      exec: function(){/*
+        var fn = function(a, b) {};
+
+        var desc = Object.getOwnPropertyDescriptor(fn, "length");
+        if (desc.configurable) {
+          Object.defineProperty(fn, "length", { value: 1 });
+          return fn.length === 1;
+        }
+
+        return false;
+      */},
+      res: {
+        babel:       false,
+        firefox38:   true,
+        firefox40:   true,
+        chrome43:    true,
+        chrome45:    true,
+        webkit:      false,
+        iojs:        false,
+      },
+    },
   },
 },
 ];
--- a/data-es6.js
+++ b/data-es6.js
@@ -3596,6 +3596,7 @@ exports.tests = [
       */},
       res: {
         babel:       true,
+        es6shim:     true,
         firefox36:   false,
         firefox37:   true,
         firefox38:   true,
@@ -3951,6 +3952,7 @@ exports.tests = [
       */},
       res: {
         babel:       true,
+        es6shim:     true,
         firefox36:   false,
         firefox37:   true,
         firefox38:   true,
@@ -4306,6 +4308,7 @@ exports.tests = [
       */},
       res: {
         babel:       true,
+        es6shim:     true,
         firefox36:   false,
         firefox37:   true,
         firefox38:   true,
@@ -4421,7 +4424,7 @@ exports.tests = [
     },
     'WeakMap.prototype.clear is missing': {
       exec: function () {/*
-        return !("clear" in WeakMap.prototype);
+        return typeof WeakMap !== "undefined" && !("clear" in WeakMap.prototype);
       */},
       res: {
         firefox38:   false,
@@ -4506,6 +4509,7 @@ exports.tests = [
       */},
       res: {
         babel:       true,
+        es6shim:     true,
         firefox36:   false,
         firefox37:   true,
         firefox38:   true,
@@ -4598,7 +4602,7 @@ exports.tests = [
     },
     'WeakSet.prototype.clear is missing': {
       exec: function () {/*
-        return !("clear" in WeakSet.prototype);
+        return typeof WeakSet !== "undefined" && !("clear" in WeakSet.prototype);
       */},
       res: {
         firefox38:   false,
@@ -5916,6 +5920,7 @@ exports.tests = [
       */},
       res: {
         babel:       true,
+        es6shim:     true,
         typescript:  temp.typescriptFallthrough,
         firefox38:   true,
         firefox40:   true,
@@ -5978,6 +5983,7 @@ exports.tests = [
       */},
       res: {
         babel:       true,
+        es6shim:     true,
         typescript:  temp.typescriptFallthrough,
         firefox38:   true,
         firefox40:   true,
--- a/data-es6.js
+++ b/data-es6.js
@@ -4275,13 +4275,18 @@ exports.tests = [
         iojs:        true,
       },
     },
-    'WeakMap[Symbol.species]': {
+    'WeakMap[Symbol.species] is missing': {
       exec: function () {/*
-        var prop = Object.getOwnPropertyDescriptor(WeakMap, Symbol.species);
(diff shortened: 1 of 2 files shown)


data-es6.js
```diff
@@ -3586,6 +3586,54 @@ exports.tests = [
         iojs:        true,
       },
     },
+    'constructor accepts null': {
```
ljharb: Please add `es6shim: true` to this test, since the `es6-shim`'s `Map` constructor does accept null.

data-es6.js
```diff
@@ -3586,6 +3586,54 @@ exports.tests = [
...
+        firefox38:   true,
+        firefox39:   true,
+        firefox40:   true,
+        chrome43:    true,
+        chrome45:    true,
+        webkit:      true,
+        iojs:        true,
+      },
+    },
+    'constructor invokes set': {
```
ljharb, line 3611: please add `es6shim: true` to this test as well.

data-es6.js
```diff
@@ -3893,6 +3941,54 @@ exports.tests = [
         iojs:        true,
       },
     },
+    'constructor accepts null': {
```
ljharb: Please add `es6shim: true` to this test, since the `es6-shim`'s `Set` constructor does accept null

data-es6.js
```diff
@@ -3893,6 +3941,54 @@ exports.tests = [
...
+        firefox38:   true,
+        firefox39:   true,
+        firefox40:   true,
+        chrome43:    true,
+        chrome45:    true,
+        webkit:      true,
+        iojs:        true,
+      },
+    },
+    'constructor invokes add': {
```
ljharb, line 3967: please add `es6shim: true` to this test as well.

data-es6.js
```diff
@@ -4275,13 +4419,31 @@ exports.tests = [
         iojs:        true,
       },
     },
-    'WeakMap[Symbol.species]': {
+    'WeakMap.prototype.clear is missing': {
       exec: function () {/*
-        var prop = Object.getOwnPropertyDescriptor(WeakMap, Symbol.species);
-        return 'get' in prop && WeakMap[Symbol.species] === WeakMap;
+        return !("clear" in WeakMap.prototype);
```
ljharb: I feel like this test should only return "false" if both `WeakMap` is present, and `WeakMap#clear` is present - in any other case, this security hole doesn't exist. (Same with the similar `WeakSet` tests)

data-es6.js
```diff
@@ -5668,6 +5896,35 @@ exports.tests = [
         iojs:        true,
       },
     },
+    'Promise.all supports iterables': {
```
ljharb: Please add `es6shim: true` to this test.

data-es6.js
```diff
@@ -5701,6 +5958,35 @@ exports.tests = [
         iojs:        true,
       },
     },
+    'Promise.race supports iterables': {
```
ljharb: Please add `es6shim: true` to this test.

data-es6.js
```diff
@@ -5701,6 +5963,36 @@ exports.tests = [
...
+      exec: function () {/*
+        var fulfills = Promise.race(new Set([
```
zloirock, line 5968: `Promise` methods do not depend on `Set`. For  generic iterables  available `__createIterableObject`.

webbedspace:
Previously, we haven't tested for the _absence_ of ES draft methods/properties before, so points 1 and 4 are somewhat extending the compat table's current scope. Moreover, it seems to suggest dozens of other absence tests of dubious relevance: should we check for the absence of `"create" in Symbol`, `"toMethod" in Function.prototype`, or `"Realm" in Reflect`? I'm not really a fan of this idea at present.

ljharb:
@webbedspace Essentially, I'd say it depends on what the thing is testing. If we're testing for a method's presence and compliance, then absence fails it. But, if we're testing for something _negative_ - specifically, known violations and/or bugs - I'd say absence passes it, because the environment lacks the problem (which is all the test is ascertaining).

webbedspace:
There are actually a few tests which check for absences of properties which are explicit prohibitions (such as `prototype` on arrow functions), but I refer in this case to unspecified properties - extensions to the built-ins beyond those in the spec are [allowed except where noted](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-error-handling-and-language-extensions), so even draft methods like `Function#toMethod` are permitted in a conforming ES6 implementation.

ljharb:
In this case, `WeakMap#clear` is not a permitted extension, because a `WeakMap` _must_ not allow any iteration or mutation except when you already have the key. It's a security invariant.

webbedspace:
Well:
- I've just read through [the WeakMap definition again](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-weakmap-objects) and it seems such an invariant against removal of key-value pairs isn't explicitly specified, at least not in the style of other invariants such as [those for proxies](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-invariants-of-the-essential-internal-methods).
- This pull request's test only checks for the absence of a property called "clear", without examining the behaviour of any present "clear" property. If a present `WeakMap#clear` method did nothing when invoked, it would be compliant.

ljharb:
@webbedspace The entire motivation for removing `WeakMap#clear` from the spec was the security invariant. If it's not clearly specified enough I can get a note added to the next edition - suffice it to say, it's prohibited. In this case, the only engines that have a property called "clear" _do_ in fact clear the WeakMap. None exist that have a noop "clear" method.

You could certainly write a test that, if the "clear" method is present, attempts to clear the map, and fails if it is indeed cleared. However, I think that'd be a waste of time until such time as an engine exists that does this.

webbedspace:
I'll write that test anyway.

I'll accept this pullreq now and make all the suggested changes myself.

ljharb:
Sounds good to me

zloirock:
WTF? Merged? `Promise` tests uses `Set`, `es6-shim` `WeakSet` 2/9, `WeakMap` 1/10?

ljharb:
I agree the `Promise` tests shouldn't use `Set`, that was a good call.

But yes, anything that doesn't have `WeakMap#clear` should pass that test, even if it doesn't have `WeakMap`. Same with `WeakSet`. I can make that even more certain by actively deleting those methods when they're detected.

webbedspace:
I disagree. The negative test (absence of `clear`) is designed to check if the existing feature (WeakMap) adheres to the security invariant (by not exposing a given API). If there is no existing feature, then there is no adherence. But, I recall we've been down a similar rhetorical path before...

Issue #498, opened 2015-04-11
Chrome 43 supports classes now


paulirish:
The table indicates `flag` but it's shipped nowadays.

webbedspace:
I don't think so?
![](http://i.imgur.com/YJRHaSZ.png)
(Unless it actually became re-flagged in 44 Canary.)

paulirish:
Hmm my apologies. 
It appears they're all behind strict mode. 

That's different than a flag, though.

kangax:
Hm, let's add a note? But then... what to show as a result? Introduce 3rd type — "Strict" (meaning only works under strict mode)? Feels iffy. On the other hand, it would also simplify `let` tests, which are essentially broken into 2 sets of strict/non-strict right now, and maybe others. There would also be cases of having "Strict + Flag", like TDZ in Babel. Maybe it's not a bad idea. @webbedspace what do you think?

paulirish:
Yeah I'd be super happy with that, especially since a developer can choose for users to run their JS in strict mode whereas they can't with a flag.

PR #574, opened 2015-07-01
Add "use strict" to class tests that are unrelated to strict mode

arv:
For example it makes no sense to fail "Array is subclassable" just.
because V8 does not allow class syntax in sloppy mode.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -593,6 +593,7 @@ exports.tests = [
     },
     'lexical "super" binding': {
       exec: function(){/*
+        "use strict";
         class B {
           qux() {
             return "quux";
@@ -1413,6 +1414,7 @@ exports.tests = [
     },
     'is block-scoped': {
       exec: function () {/*
+        "use strict";
         class C {}
         var c1 = C;
         {
@@ -1431,6 +1433,7 @@ exports.tests = [
     },
     'class expression': {
       exec: function () {/*
+        "use strict";
         return typeof class C {} === "function";
       */},
       res: {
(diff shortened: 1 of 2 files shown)


arv:
@kangax What do you think? Also, I guess I should update the browser results?

ljharb:
I think we might need separate tests, one for strict and one for sloppy.

Chrome absolutely should fail some test if it doesn't work in sloppy mode. Sloppy code isn't going away any time soon :-)

arv:
I kept all the top level class tests as is (sloppy).

But the question if super.prop works as expected inside an arrow function is irrelevant to sloppy mode.

ljharb:
That's quite true. But since an engine exists where the mode affects how it works, it should probably be tested in both modes. If V8 doesn't support something in sloppy mode that the spec says it should, something should fail. It's not something that should be acceptable, or glossed over, that a feature doesn't work by default (since sloppy is default outside of `class` and modules).

webbedspace:
The solution I've now implemented for this is to add a "Strict" result value, which is regarded for percent calculations as equivalent to "Flag" - as discussed previously in #498. (This will also report "strict-only" results in browser tests, which finally allows features with this restriction to be tracked more easily.)

paulirish:
Commit of @webbedspace's solution here: https://github.com/kangax/compat-table/commit/91c2aab8648800c81d0927d71ce4be5052b57d5a

screenshot: 
image

While I agree we can't treat a pass in strict as 100% support, I'm happy the table now communicates the feature is available to devs if they go for strict.  

I think it's probably worth scoring as partial credit instead of 0%... I've opened that as a new issue: #575

arv:
Thanks. Having the tests run in both strict and sloppy seems like the right thing to do.

Issue #507, opened 2015-04-17
Make it easier to select/copy test contents


kangax:

bterlson:
:+1: :+1: :+1:

bzbarsky:
Maybe make it so a click on the question mark in a circle locks the tooltiip in place so it doesn't disappear?

kangax:
@bzbarsky That's a nice idea. Not sure how discoverable it would be (for those who don't know). I guess it's way better than what we have now, even if not perfect solution.

PR #576, opened 2015-07-07
Click to lock the tooltip containing the test code

claudepache:
When clicking on the circled "c", the tooltip containing the test code is locked in place.
That allows to easily select and copy the test code, resolving Issue #507.


hoobdeebla:
very useful. i've had to use devtools to select test code many times before, and this'll make it a lot easier to extract the code. :+1:

kangax:
@claudepache, thanks for taking care of this! Have you tested it on few browsers?

claudepache:
@kangax, tested on mainstream desktop browsers (Safari, FF, Chrome, IE, last stable releases) and on iPad.

PR #580, opened 2015-07-17
Proxy internal calls tests

webbedspace:
[Preview link](https://raw.githack.com/webbedspace/compat-table/proxy/es6/index.html)

This in-progress PR adds tests for a little-acknowledged ES6 API surface: Proxy traps called by ECMAScript internal methods. For instance, a wide variety of internal/built-in methods call "get" on a passed-in object as an incidental part of their algorithm, to access, say, the "length" or "prototype" properties - and the order and parameters of these calls are precisely specified.

For these tests, I only wanted to include:
- Tests which fall outside the realm of "basic support" covered by the basic Proxy tests (for instance, checking that "get" is called by an assignment, the increment operator, a for-in loop head, etc., rather than something less observable, like the == operator).
- Tests which can only be observed by a Proxy (for instance, the Typed Array methods' [[Get]] calls are unobservable because Proxy objects lack the Typed Array internal slots, and thus the methods will immediately reject them).
- Tests for methods which are precisely specified - the behaviour of Array#sort, for instance, is largely implementation-defined.

**This isn't complete yet**, as I haven't entered any support results yet - although the only working Proxy implementations are Firefox, Edge and EchoJS, there are plenty of old Firefox versions (from version 18 to 41) to revisit and record results for. Also, I want to edit the test weighting system so that these could have an even lower weighting (mainly because they are grouped into four "tests" in a way that artificially inflates their weight). However, I await any feedback any of you may have about the tests and their code, especially bugs.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -5224,6 +5224,126 @@ exports.tests = [
   },
 },
 {
+  name: 'Proxy, internal \'get\' calls',
+  category: 'built-ins',
+  significance: 'small',
+  link: 'http://www.ecma-international.org/ecma-262/6.0/#sec-proxy-object-internal-methods-and-internal-slots',
+  subtests: {
+    'Abstract Relational Comparison': {
+      exec: function() {/*
+        // Abstract Relational Comparison -> ToPrimitive -> Get -> [[Get]]
+        var get = [];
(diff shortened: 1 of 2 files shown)


webbedspace:
OK, I believe it's complete now.

PR #582, opened 2015-07-20
split Reflect.ownKeys into base test and key order

claudepache:
Split Reflect.ownKeys into two subtests
- the first one checks that Reflect.ownKeys enumerates the keys, presumably in the same order as for/in.
- the second one checks that the order is correct.

Firefox Nightly (as well as Babel, TypeScript, and es6-shim, according to the table) passes the first subtest, not the second.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -5395,62 +5395,6 @@ exports.tests = [
         edge:        true,
       },
     },
-    'Reflect.ownKeys, key order': {
-      exec: function() {/*
-        var obj = {
-          2:    true,
-          0:    true,
-          1:    true,
-          ' ':  true,
-          9:    true,
-          D:    true,
-          B:    true,
-          '-1': true,
-        };
-        obj.A = true;
-        obj[3] = true;
-        Object.defineProperty(obj, 'C', { value: true, enumerable: true });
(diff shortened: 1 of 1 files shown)


ljharb:
👍 LGTM

evilpie:
Seems to me that moving this to Misc - Own Property Order would be more consistent.

claudepache:
> Seems to me that moving this to Misc - Own Property Order would be more consistent.

OTOH, one might want to keep all tests for `Reflect.ownKeys` together.

But I agree with you, because there is more chance for an implementation to use the same (wrong or correct) order for `Reflect.ownKeys`, `for/in`, etc., than to implement `Reflect.ownKeys` with its own correct order.

If the main contributors agree, I’ll move the "Reflect.ownKeys, key order" and "Reflect.ownKeys, symbol order" tests under "Misc – Own Property Order". (And add a similar test for `Reflect.enumerate`, BTW.)

ljharb:
I think that the tests definitely need to be split. All key iteration tests should have one test that tests that the proper keys are iterated, and a separate one that tests the order of those keys.

I think it does make sense to put the "key order" tests under the "own property order" category.

webbedspace:
In the end I decided to do this manually in 0c2aa9a. Thanks anyway.

PR #592, opened 2015-07-28
[esintl] Fix table headers not match columns

Maks3w:


ljharb:
This will be automatically reverted every time someone runs `npm run build:compilers` - this likely needs to be fixed elsewhere.

PR #600, opened 2015-08-06
CSS - Enforce default background and color

Delapouite:
Without specifying these explicit colors, the site is unreadable for people who have configured their browser to display white on black texts.

### DIFF
--- a/master.css
+++ b/master.css
@@ -5,6 +5,8 @@ html {
   box-sizing: inherit;
 }
 body {
+    background-color: #fff;
+    color: #000;
     font-family: 'Open Sans', sans-serif; margin: 0; font-weight: 300; display: inline-block;
     width: 100%;
 }


kangax:
Out of curiosity — how many sites "break" with this kind of setting, in your experience?

Delapouite:
Hard to tell without stong statistics but I would say roughly 20%.

2 alternatives then : the site is closed source and I have to monkey patch the style through Stylish, or like the present case where I can contribute and fix it once and for all.

ljharb:
What browser settings are you referring to here? When I turn on "invert colors" on my OS, with or without "use grayscale", the site looks fine.

Delapouite:
browser.display.background_color
browser.display.foreground_color

in Firefox about:config

master.css
```diff
@@ -5,6 +5,8 @@ html {
   box-sizing: inherit;
 }
 body {
+    background: #fff;
```
ljharb: Can this be the more granular `background-color`?

ljharb:
OK, i'm not sure who would ever do this besides you, but I see what you mean now.

Delapouite:
Thanks for your suggestion @ljharb I switched the commit from using background to background-color

As for the reason why people tweak their color, I guess it's the same anwser that people prefering dark themes over bright ones in their text editor. In my case it's because I find it better on the eyes at the end of the day.

ljharb:
@Delapouite sure, but i meant, why anyone would do it at the browser level instead of the OS level :-)

Either way this is a small change so I don't see any reason not to merge it. @kangax?

Delapouite:
Because the "OS" concept is meaningless when the two only windows displayed every day by Xmonad are a solarized xterm and Firefox. ;-)

kangax:
Yep, ok with me, merged.

PR #602, opened 2015-08-20
Show columns when hover

ghost:
es6


webbedspace:
I feel like sticky header proposals, such as #532 (but expanded to all browsers), are a much better solution.

PR #604, opened 2015-08-21
Update WebKit to  r188682

Constellation:
- Arrow function has lexical `this` binding feature
- `Reflect.{get, getOwnPropertyDescriptor}`


kangax:
> Arrow function has lexical this binding feature

Omg, yes!

PR #605, opened 2015-08-21
Add tests for await and await *

thomasfosterau:


ljharb:
`await *` is not part of the proposal, and is highly unlikely to make it in. See https://github.com/tc39/ecmascript-asyncawait for the official spec proposal.

data-es7.js
```diff
@@ -337,6 +337,23 @@ exports.tests = [
...
+          return 10 + await Promise.resolve(10);
+        })() instanceof Promise
+      */},
+      res: {
+      }
+    }, 
+    'await * support' : {
+      exec: function () {/*
+      return (async function(){
+        var g = await * Promise.all([Promise.resolve(10), Promise.resolve(1)]);
```
ljharb: At the time when `await*` was part of the proposal, it would have made `Promise.all` implicit - so this isn't correct at any point in the proposal.

thomasfosterau:
Ah right - is it still worth testing for `await` separately to `async function(){}`?

ljharb:
It can't hurt - but until somebody implements one without the other, the separate tests aren't likely to help much :-)

thomasfosterau:
Alright, the `await *` test is removed.

PR #607, opened 2015-08-22
Add tests for Observables

thomasfosterau:


data-es7.js
```diff
@@ -1190,6 +1190,59 @@ exports.tests = [
...
+      */}
+    },
+    'Symbol.observable well known symbol': {
+      exec: function () {/*
+        return typeof Symbol.observable === 'symbol';
+      */}
+    },
+    'Observable.prototype.subscribe': {
+      exec: function () {/*
+        return 'subscribe' in Observable.prototype;
```
ljharb, line 1212: should these check that they're functions, and hasOwnProperty, not just `in`?

data-es7.js
```diff
@@ -1190,6 +1190,59 @@ exports.tests = [
...
+      */}
+    },
+    'Observable.of': {
+      exec: function () {/*
+        return Observable.of(1, 2, 3) instanceof Observable;
+      */}
+    },
+    'Observable.from': {
+      exec: function () {/*
+        return Observable.from([1,2,3,4]) instanceof Observable;
```
ljharb: Should this test generic iterables (as a separate test) as well as arrays?

thomasfosterau: `Observable.of` and `Observable.from` should - I'll add that to the tests. 

The `Symbol.observable` property is a getter that returns the observable itself - is testing for this exact behaviour worth it or is just existance enough (there isn't any precedent for this with Iterators)?

data-es7.js
thomasfosterau, line 1212: Yes they probably should - I'll add a function check as well as check the return type.

data-es7.js
ljharb: I think that eventually we'll want these tests - so it's your call. It might be nice to have the tests be specific - you can't really overtest these things :-)

thomasfosterau: Just added a test for iterables to `Observable.from` - the commit is below.

On second thoughts `Observable.of` just need to use more than one argument (and since it's an Observable, we'd need async tests to ensure the arguments are used).

data-es7.js
```diff
@@ -1190,6 +1190,63 @@ exports.tests = [
...
+        return 'filter' in Observable.prototype && o.filter(function(e){return true}) instanceof Observable;
+      */}
+    },
+    'Observable.prototype.map': {
+      exec: function () {/*
+        var o = new Observable();
+        return 'map' in Observable.prototype && o.map(function(e){return e}) instanceof Observable;
+      */}
+    },
+    'Observable.prototype.@@observable': {
```
webbedspace: Spec-shorthand like `@@` isn't used in the compat-table if there's a userland equivalent form - in this case, that's `Observable.prototype[Symbol.observable]`.

data-es7.js
```diff
@@ -1190,6 +1190,63 @@ exports.tests = [
...
+      */}
+    },
+    'Observable.prototype.subscribe': {
+      exec: function () {/*
+        return 'subscribe' in Observable.prototype;
+      */}
+    },
+    'Observable.prototype.forEach': {
+      exec: function () {/*        
+        var o = new Observable();
```
webbedspace, line 1217: Isn't `new Observable()` with no arguments supposed to throw an error?

ljharb, line 1217: Yes - `If IsCallable(subscriber) is false, throw a TypeError exception.`, step 2 in https://github.com/zenparsing/es-observable#observablesubscriber

We should also verify step 1 - that it throws a `TypeError` when it's not invoked with `new`.

zloirock:
Too unstable for adding detail tests.

ljharb:
That's a fair point as well. It's only stage 1.

zloirock:
Problem rather not in stage. Observable API too often serious changes.

ljharb:
The stage is an indication of this :-) when it moves to stage 2, the API will begin to be more stable.

data-es7.js
thomasfosterau: :+1: changed.

kangax:
Is this ready to be added? Just need to rebase?

zloirock:
``` js
var o = new Observable();
```

not ready.

Issue #593, opened 2015-07-28
Proxy test crashes Edge


bterlson:
A recently added test (81ca7ced3452eb49b26ff82fce41aad1cf1bc0db) includes the following:

``` js
var p = new Proxy([0,0,0,0,0], {});
p.length = 1;
```

This crashes Edge presently (nice find :+1:). While we are working on a fix I wonder if we can disable this test temporarily? Right now edge users are unable to see their compat score :(

webbedspace:
While I personally regret having to add a special case for any platform, no matter how temporary, I can agree that knowingly crashing the browser is rather bad UX regardless.

kangax:
Back in the days we had a module/import test that would crash Chrome, so I also disabled it at some point, with the note. I think this is a fair decision.

bterlson:
Thanks guys. I personally regret shipping with this crash bug so regret all around! :-P

Will let you know once we have a fix out!

zloirock:
@bterlson also
`Reflect.defineProperty(Object.defineProperty({}, 'x', {value: 1}), 'x', {value: 2})` -> `TypeError`, should be `false`, the same in other cases.

`Reflect.setPrototypeOf.length` -> `3`, should be `2` (wtf?)
`Reflect.get.length` -> `3`, should be `2`
`Reflect.set.length` -> `4`, should be `3`

bterlson:
@zloirock Indeed! I think we are tracking all of these already (thanks to test262) but I will double check.

Anything else I should investigate while I'm at it?

PR #610, opened 2015-08-23
[ES6] Extend the Reflect.defineProperty test

zloirock:
MS Edge throws error instead of returning `false`, see https://github.com/kangax/compat-table/issues/593#issuecomment-126067756


ljharb:
@zloirock just to confirm, you're testing these in the latest Edge released with windows 10?

zloirock:
Yes. BTW, I think need do something with Edge releases, 1 column for all - bad idea.

PR #611, opened 2015-08-27
Add site logo & responsive design

williamboman:
Inspired by the [logo.js](https://github.com/voodootikigod/logo.js) JS logo.

Before
ecmascript_5_compatibility_table 2
After
screenshot 2015-09-02 at 19 30 46


kangax:
Heh, nice, thanks :)

+1 from me

williamboman:
I may or may not have used Preview.app for this :hear_no_evil:

ljharb:
LGTM

webbedspace:
Dissenting opinion: I feel favicons should, as a rule, represent the colour scheme of the page itself - and the current compat-table palette doesn't use strong yellow much. I'd be more inclined toward something similar to caniuse.com's favicon.

ljharb:
That's not a rule most sites follow. Favicons match logos; color schemes are their own thing. The yellow is the official JS logo, and I think that fits the best.

kangax:
Well, considering that logo and site color/theme are usually at least somewhat similar, I understand the point @webbedspace is making. Looking at yellow JS logo, I don't associate it with compat tables. So ideally, I'd love to see a separate compat table logo (perhaps yellow(?) ES) which could then be used on a site and in favicon. But considering lack of resources, I'd probably take this over blank favicon for now.

williamboman:
What about adding something like this then?

JS & ECMAScript beside each other might be a bit confusing though (as if it already isn't).

meh

screenshot 2015-08-29 at 17 40 41

kangax:
What if we make the same logo but with ES instead of JS. Feels [a little strange](http://stackoverflow.com/a/1507934/130652) to use JS for ES compat tables.

williamboman:
@kangax I totally agree. I've updated it now. I also included some responsive stuff I did (maybe that should be another PR though). Let me know if you want me to remove the responsive stuff from this PR.

edit: Live preview available here: http://williamboman.github.io/compat-table/es5/

es5/index.html
```diff
@@ -26,20 +27,24 @@
 </head>
 <body class="es5">
   <div id="header">
+    <a href="/" id="logo"><img src="/logo.png" alt="JS compat"></a>
```
williamboman: I just noticed this breaks on `gh-pages`, I'll update them to be relative.

kangax:
The logo looks great to me! @ljharb @webbedspace @zloirock whacha guys think?

zloirock:
LGTM :+1:

williamboman:
I want to point out that I retroactively included responsive design for the nav bar in this PR as well (in case that went unnoticed).

FWIW, the typeface used for the logo is Bitstream Vera Bold.

ljharb:
👍

kangax:
I'm not sure if it's a regression from this PR or was like this for a while but I just noticed that scrolling on iPad is ridiculously slow. As if there's expensive event handler on scroll or something else causing freezes for few seconds. Anyone else see this?

Sent from my iPad

williamboman:
@kangax If it was introduced in this PR it'd most likely be the viewport meta tag. Do you still get lag without it?

PR #612, opened 2015-08-29
[ES6] Add test "JSON.stringify ignores symbols"

zloirock:
[`SerializeJSONObject`](http://www.ecma-international.org/ecma-262/6.0/#sec-serializejsonobject) uses [`EnumerableOwnNames`](http://www.ecma-international.org/ecma-262/6.0/#sec-enumerableownnames):
- `JSON.stringify` should ignore `Symbol` keys in objects

[`SerializeJSONProperty`](http://www.ecma-international.org/ecma-262/6.0/#sec-serializejsonproperty) hasn't special case for symbols. That mean symbols should be converted to `undefined`:
- `JSON.stringify` should ignore `Symbol` values in objects
- `JSON.stringify` should convert to `null` `Symbol` values in arrays

MS Edge `JSON.stringify` converts symbol values to object, for example, `'[{}]'` for the last part of this test.
Traceur fails all parts of this test.
I didn't test it in the last WebKit and Echo JS.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -8586,6 +8586,23 @@ exports.tests = [
         webkit:      true,
       },
     },
+    'JSON.stringify ignores symbols': {
+      exec: function() {/*
+        var object = {foo: Symbol()};
+        object[Symbol()] = 1;
+        var array = [Symbol()];
+        return JSON.stringify(object) === '{}' && JSON.stringify(array) === '[null]';
+      */},
+      res: {
+        babel:       true,
+        typescript:  typescript.corejs,
+        firefox36:   true,
+        chrome35:    flag,
+        chrome38:    true,
+        node:        true,
+        iojs:        true,
+      },
+    },
     'global symbol registry': {
       exec: function() {/*
         var symbol = Symbol.for('foo');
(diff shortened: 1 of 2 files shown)


ljharb:
I don't think it should count as compliance to add `Symbol.prototype.toJSON` - `JSON.stringify` itself needs to be overridden in order for a non-engine to be compliant here.

`Symbol`, by spec, does not have a `toJSON` method at all.

ljharb:
([updated](https://github.com/kangax/compat-table/commit/9c56aca2e0fdc2bf39a7539aa2ad4e7b9b694fe7?w=1) the WebKit results, which handles this correctly)

webbedspace:
@ljharb Uhh... no it doesn't?

ljharb:
Hmm - it passed in my Webkit Nightly. http://trac.webkit.org/changeset/189162 is the relevant changeset.

ljharb:
Ah, I see @zloirock updated the test. Feel free to revert https://github.com/kangax/compat-table/commit/9c56aca2e0fdc2bf39a7539aa2ad4e7b9b694fe7?w=1 but the next nightly release should fix it. If you want to wait, I can reverify within the next day or so and ensure it's up to date.

zloirock:
@ljharb I updated test in `core-js`, not in `compat-table`.

ljharb:
@webbedspace The example in your screenshot shows correct handling - Symbols should be stringified to null; and (not shown in your test) Symbol keys should be omitted.

ljharb:
OK, so I think I forgot that `{[Symbol()]: true, a: Symbol()}` should in fact stringify to `"{}"` - however, as of r189183, WK Nightly also complies with this :-) so although I jumped the gun, it's correct now.

PR #616, opened 2015-09-01
Add Safari 9 data (from public beta 6)

mgol:
Fixes #556 

Public beta 6 has been released yesterday so it's fresh.

PR #617, opened 2015-09-01
Make the "Desktop browsers" column wide enough

mgol:
This regression was introduced when Safari 9 results were added

Refs #616


kangax:
Great, thanks!

PR #619, opened 2015-09-02
Chrome 45 has been released

mgol:


webbedspace:
🚨You didn't update the colspan in `skeleton.html`🚨

mgol:
colspan of which column? What does display incorrectly? I didn't add/remove any browser here.

webbedspace:
Actually, you sorta did - `obsolete` browsers aren't visible by default. Take a look at the current page and see where e.g. the mobile column is now.

(This wouldn't be a problem if this thing automatically computed the colspans at build time, like a useful piece of software.)

mgol:
> Actually, you sorta did - obsolete browsers aren't visible by default.

They're not visible but they're there, that doesn't affect `colspan`.

What's broken according to you? I've been clicking various checkboxes and the table looks fine in all cases.

webbedspace:
> They're not visible but they're there, that doesn't affect `colspan`.

It does, though? All of the cells in the obsolete browser columns have `display:none`, so that entire column is essentially gone, and the top cell is spanning over a column that isn't there anymore, and thus it's 1 greater than what it should be.

Just look at this when you load the page initially:

So you know, the `colspan` is dynamically recalculated and fixed by the JS whenever you adjust a checkbox, so the problem disappears as soon as that action is performed.

mgol:
You're right, d'oh. So it's wrong only on initial load as that's when it's not getting re-computed, right?

PR: #621. I just modified colspan and run `node build.js`; I hope that's all...

PR #620, opened 2015-09-03
index.html: avoid additional redirect from GitHub, update to ES6

williamboman:
- The appended / is needed to avoid a 301 response with a non-HTTPS
  location from GitHub.

Also updated to redirect to ES6.

``` sh
$ curl -I https://kangax.github.io/compat-table/es5 # essentially the same as https://kangax.github.io/compat-table/ in a browser
#.............^
HTTP/1.1 301 Moved Permanently
Location : http://kangax.github.io/compat-table/es5/
#..............^
```


kangax:
:+1:

PR #624, opened 2015-09-05
replacing (function(){}).length with (function(){}).prototype

claudepache:
`(function(){}).length` is now configurable in ES6; replacing it with `(function(){}).prototype`, so that the test "attempting to delete an unconfigurable property should throw a TypeError" remains valid.

### DIFF
--- a/strict-mode/index.html
+++ b/strict-mode/index.html
@@ -213,10 +213,10 @@
       (function(){
         var error;
         try {
-          eval('delete (function(){}).length;');
+          eval('delete (function(){}).prototype;');
         }
         catch (err) { error = err; }
-        print('<span class="test"><code>delete (function(){}).length;<\/code> is a TypeError<\/span>' + colorResult(error instanceof TypeError));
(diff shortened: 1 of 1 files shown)


ljharb:
It might be safer, long-term, to modify this test so it explicitly creates a nonconfigurable property, and then attempts to delete it?

claudepache:
> It might be safer, long-term, to modify this test so it explicitly creates a nonconfigurable property, and then attempts to delete it?

The advantage of the current test (and of all tests in that file), is that it uses only pre-ES5 features (apart the `"use strict"` directive).

claudepache:
Correction of my previous comment: some tests in that file do use ES5 features. (That weakens my argument.)

ljharb:
Since the test is for strict mode, which doesn't exist in a pre-ES5 browser, I don't think that's an important requirement :-)

webbedspace:
It is, though: a partial ES5 implementation may easily have one but not the other. (Granted, none such seem to exist, but that's just coincidence and not something a robust test should assume.)

ljharb:
Fair point.

PR #625, opened 2015-09-08
Add Node.js 4.0.0 & remove io.js

mgol:
Node 4.0.0 doesn't differ a lot from io.js 3.3.0 and io.js is not
developed anymore (it merged with Node.js) so it doesn't seem worth it
to keep its data.

P.S. Current page at http://kangax.github.io/compat-table/es6/ looks broken; the "Desktop browsers" column appears twice, the second one gathers servers & mobile browsers as well... Clicking on checkboxes doesn't fix it. In my PR it looks fine.

P.S.2. I'm a little worried that `node build.js` is not idempotent; running it changes "yes" to "no" in some rows, running it again changes it back. O_o Could someone check what's going on here? @kangax @webbedspace


mzgol:
More concretely, the results for the test `%TypedArray%.prototype.join` are flipping.

ljharb:
I think this is an optimal eventual destination - but people will still be using `io.js` in production for awhile - just like people still use node `0.6` and `0.10` in production. I think we should preserve the results for `io.js`.

mgol:
Hmm, maybe it's just the order of results is randomly changing. Which might be related to the fact that objects don't have an order of iteration over keys specified so keeping it stable is fundamentally incompatible with current data structure. Unless it's sorted in some way but it doesn't seem so.

mgol:
@ljharb People may as well use io.js 2.5.0 for a while (io.js 3.0.0 came out shortly after that) but it's been removed from the site. I don't see how they're different.

If we want to keep older Node/io.js versions in the table, perhaps all should be kept, not just Node 0.12, 4.0 & io.js 3.3.0.

mgol:
io.js is different from Node.js 0.10, though - the latter is still supported and will be supported until October 2016 or so, Node 0.12 will be supported until the end of 2016, Node.js 4 will be supported until 2018.

io.js 3.x.y is no longer supported as far as I understand.

ljharb:
Personally I think we should keep them all, just marked "obsolete". Including 2.5.0.

node 4.0 isn't actually released yet, so io.js 3.x is still 100% supported.

mgol:
Yeah, sorry, I meant the situation after today or a few more days since Node.js 4.0.0 is going to get released really soon (there are already experimental final builds with `process.version === 'v4.0.0'`). I meant this PR to get merged once Node 4 gets released, not necessarily before that.

Who decides what gets kept? I can update this PR to restore io.js 3.3.0 but I don't want to go back and forth, I'd like to hear the final decision first. :)

ljharb:
I don't see any harm in ever keeping any data as "obsolete". It's easy to keep, hard to reconstruct.

mgol:
Yeah, I think you're right. I didn't notice you're a collaborator to this repo. I assume that means your decision is official. :) So I'll restore the data.

mgol:
@ljharb OK, io.js restored. Unfortunately, since io.js & Node.js are in different families I had to put io.js after Node.js 4 as otherwise Node 4 wouldn't pick up Node 0.12 data as a base. Putting io.js between Nodes would require duplicating all Node 0.12 data to Node 4.

Let me know what you think.

mgol:
Node.js 4.0.0 [has been officially released](https://nodejs.org/en/blog/release/v4.0.0/); this PR is ready to go unless there are any further comments.

ljharb:
I'm OK with this 👍 i'll let one of the other contributors merge it tho, just so i'm not the only opinion represented

PR #627, opened 2015-09-10
Add subtest "super() invokes the correct constructor"

claudepache:
This test checks that `super()` is _not_ a synonym of `super.constructor()`.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -2303,6 +2303,7 @@ exports.tests = [
         node4:       strict,
         chrome41:    strict,
       },
+    },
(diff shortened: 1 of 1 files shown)


claudepache:
The results for different environments should be completed. For those I have tested:
- webkit: FAIL, see https://bugs.webkit.org/show_bug.cgi?id=149001
- firefox: does not support super() yet
- chrome: current version: PASS in strict mode (no support in sloppy mode). Should be tested for earlier versions.

I am not able/willing to test other environments.

PR #628, opened 2015-09-10
correct tests for Miscellaneous subclassable

claudepache:
add a check that the prototype chain of instances is correct. Webkit is failing that check

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -10471,12 +10471,13 @@ exports.tests = [
         class C extends Boolean {}
         var c = new C(true);
         return c instanceof Boolean
+          && c instanceof C
           && c == true;
       */},
       res: {
         typescript:  typescript.fallthrough,
-        safari9:     true,
-        webkit:      true,
+        safari9:     false,
+        webkit:      false,
         chrome43:    strict,
         node4:       strict,
       },
@@ -10486,12 +10487,13 @@ exports.tests = [
         class C extends Number {}
         var c = new C(6);
         return c instanceof Number
+          && c instanceof C
           && +c === 6;
       */},
       res: {
         typescript:  typescript.fallthrough,
-        safari9:     true,
-        webkit:      true,
+        safari9:     false,
+        webkit:      false,
         chrome43:    strict,
         node4:       strict,
       },
@@ -10501,14 +10503,15 @@ exports.tests = [
         class C extends String {}
         var c = new C("golly");
         return c instanceof String
+          && c instanceof C
           && c + '' === "golly"
           && c[0] === "g"
           && c.length === 5;
       */},
       res: {
         typescript:  typescript.fallthrough,
-        safari9:     true,
-        webkit:      true,
+        safari9:     false,
+        webkit:      false,
         chrome43:    strict,
         node4:       strict,
       },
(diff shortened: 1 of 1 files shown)


claudepache:
Let's take, for example, the test "Boolean is subclassable". The old test was:

``` js
class C extends Boolean {}
var c = new C(true);
return c instanceof Boolean && c == true;
```

The basic check `c instanceof C` was missing. That maked the test void for implementations (such as webkit) that support class syntax, but for which the Boolean constructor just acts as a factory returning an object of predefined prototype, ignoring `new.target`, as it was the case in ES5.

For Array, RegExp, Function, Promise, this is tested by the subtest "correct prototype chain", (although I think it should be part of the "basic functionality" subtest).

data-es6.js
```diff
@@ -10540,12 +10543,12 @@ exports.tests = [
         set.add(123);
         set.add(123);
 
-        return set.has(123);
+        return set instanceof S && set.has(123);
```
ljharb, line 10546: this should probably check `set instanceof S && s instanceof Set`.

However, to truly be sure it's working properly, `instanceof` is irrelevant - you need to check for internal slots. `return Set.prototype.size.call(set) === 1;` will throw unless it's properly subclassed, ie, the internal slots are set up properly.

I think a similar approach should be taken on all of the builtin subclassing tests.

claudepache, line 10546: > this should probably check set instanceof S && s instanceof Set

`set instanceof Set` follows from `set instanceof S` if the implementation implements `extends` correctly (which should be tested elsewhere in the compat table). Moreover the current test uses `set.add()` and `set.has()` in a manner that assumes that `set` inherits from `Set.prototype`.

> However, to truly be sure it's working properly, instanceof is irrelevant 

On the contrary, `instanceof` is fundamental: It means that when you write `set.someRandomMethod()`, a method named  `"someRandomMethod"` is searched in the right place.

> you need to check for internal slots. return Set.prototype.size.call(set) === 1; will throw unless it's properly subclassed, ie, the internal slots are set up properly.

The current test checks that `set` has `.add()` and `.has()` methods that work as expected. If you are not paranoid, you may assume that these methods are those inherited from `Set.prototype`, thus that `set` supports the methods that were intended to `Set` objects (which, I assume, is what you mean by "the internal slots are set up properly").

ljharb, line 10546: What I mean is that prior to ES6, you could not properly subclass builtins - meaning, internal slots (things users can't control) wouldn't be set up. For example, `String.prototype.valueOf.call(foo)` would always throw, even if `foo` was an instance of `class Foo extends String { }` or the equivalent.

The _sole_ new behavior that ES6 allows is that using `extends`, and calling `super` in the constructor, sets up the internal slots, so these things don't throw.

`instanceof` merely tells you which `[[Prototype]]` will be referenced - that something has `Set.prototype` in its prototype chain **does not** mean it's a `Set`. The only way you can know it's truly a `Set` is if you use one of the builtin methods that throws when internal slots are missing.

I've had lots of debates about "branding" with the committee, and this is the method they all agreed must be used for branding - and as such, it's the only way to accurately test for builtin-subclassability.

claudepache, line 10546: > The only way you can know it's truly a Set is if you use one of the builtin methods that throws when internal slots are missing.

We do agree. Concerning the test, I’ve just claimed that this particular point is already tested through the invocations of `.add()` and `.has()`, unless the implementation is grossly cheating.

ljharb, line 10546: Ah, I see what you mean - because you're not overwriting `add` or `has`, it's falling back to the `Set.prototype` methods. I still feel it'd be more robust doing an explicit `Set.prototype.has.call` etc, but I withdraw any previously implied objection :-) Thanks for bearing with me and explaining!

PR #632, opened 2015-09-12
Re-enable user scaling

williamboman:
[Live Preview](https://williamboman.github.io/compat-table/)


williamboman:
I can't think of a simple way to retain responsive design and not having scroll lag (don't have time to do render profiling). Is it even worth keeping responsive design, or should it be reverted?

PR #640, opened 2015-09-27
Add a "test-" prefix so that things like "Symbol" don't shadow the global Symbol

### LINKED ISSUE #477: IDs that match global names conflict - add a prefix?
`<td id="SIMD">` will shadow the global `SIMD` object. Either the test needs refinement, or the ID needs to not be "SIMD"

ljharb:
I'm not sure how any of these Symbol tests are marked as "passed" for anyone, since clearly nobody verified them in the browser :-)

Fixes #477.

### DIFF
--- a/build.js
+++ b/build.js
@@ -434,7 +434,7 @@ function dataToHtml(skeleton, rawBrowsers, tests, compiler) {
         t.significance === "medium" ? 0.5 : 1)
       .addClass(isOptional(t.category) ? 'optional-feature' : '')
       .append($('<td></td>')
-        .attr('id',id)
+        .attr('id', 'test-' + id)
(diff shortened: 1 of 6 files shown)


ljharb:
I'll merge this in an hour or so unless there's objections.

kangax:
Dear god, how did this get through in the first place... Good catch, thanks!

Related: http://kangax.github.io/domlint/ that I made like... forever ago :)

webbedspace:
Small problem: hyperlinks to specific rows (such as [this](https://kangax.github.io/compat-table/es6/#Symbol)) no longer scroll to them, because the old IDs were being used as URL fragment references. At the least, all the generated `<a>` anchor links need updating too.

ljharb:
aha, thanks - I'll fix that now

ljharb:
k, that should be consistent, and apply the same "test-" prefix to subtest IDs.

PR #647, opened 2015-10-01
Fix laggy scrolling

### LINKED ISSUE #645: Compatibility table is unusable in Safari
The compatibility table is unusably slow in Safari (both 8 and 9). Redraw is so slow, that the page is effectively unusable and cannot be scrolled and having the window in the foreground pegs a CPU.

williamboman:
Prioritize usability over responsive navbar design.

Closes #645.

### DIFF
--- a/master.css
+++ b/master.css
@@ -293,8 +293,7 @@ td[title] {
 
 /* page header */
 #body {
-    padding-left: 1em;
-    overflow-x: auto;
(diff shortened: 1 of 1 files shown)


zloirock:
LGTM. Enough fast even with obsolete browsers.

zloirock:
I don't see any objections, so I merge it.

PR #653, opened 2015-10-10
Check that Error.prototype is not an Error object

claudepache:
Test: "built-in prototypes are not instances".
Add checks for Error and TypeError.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -11808,6 +11808,10 @@ exports.tests = [
         try {
           Date.prototype.valueOf(); return false;
         } catch(e) {}
+        if (Object.prototype.toString.call(Error.prototype)  !== '[object Object]')
+          return false;
+        if (Object.prototype.toString.call(TypeError.prototype) !== '[object Object]')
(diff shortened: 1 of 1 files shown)


data-es6.js
```diff
@@ -11808,6 +11808,10 @@ exports.tests = [
         try {
           Date.prototype.valueOf(); return false;
         } catch(e) {}
+        if (Object.prototype.toString.call(Error.prototype)  !== '[object Object]')
+          return false;
+        if (Object.prototype.toString.call(TypeError.prototype) !== '[object Object]')
```
ljharb: If you're going to check `TypeError`, what about all the other built in error types?

claudepache: I've just picked a random representative of _NativeError_. Do you think it is useful to do a thorough check?

ljharb: We might as well - implementations could easily do `TypeError` but fail to do `RangeError` or `SyntaxError` :-) I think you could just do this in a `forEach`?

PR #658, opened 2015-10-16
Update WebKit to r191173

Constellation:
- Improve TypedArray
- Improve ES6 class and super


kangax:
@Constellation please don't forget to update revision as well!

PR #659, opened 2015-10-17
Update WebKit to r191235

Constellation:
- Implement `String.prototype.normalize`. Now String.prototype methods are complete.


ljharb:
Please update the revision number on line 366 as well.

Constellation:
Rebaseline :)

PR #665, opened 2015-10-19
Detach/attach subtests HTML Nodes and click event delegation

shospodarets:
Instead of CSS show/hide added detach/attach JS mechanism to subtests rows.
Also provided click delegation (instead of adding click even listener to each subtest row now it's one global click listener).

**Results**
Before:
Alt Text
Alt Text

After:
Alt Text
Alt Text

### DIFF
--- a/master.css
+++ b/master.css
@@ -155,6 +155,11 @@ table.one-selected td.selected {
     left: 1px;
     top: 1px;
 }
+.supertest.is-open .folddown {
+    -webkit-transform: rotate(90deg);
+    -ms-transform: rotate(90deg);
+    transform: rotate(90deg);
+}
 .subtest td:first-child {
   padding-left: 1.2em;
   font-style: italic;
@@ -162,6 +167,9 @@ table.one-selected td.selected {
 .subtest {
   font-size: 0.8em;
   display:none;
+}
+.js-applied .subtest {
+    display: table-row;
 }
 .subtest td.yes, .subtest td.no, .subtest td.strict {
   border-left: solid 5px #eee;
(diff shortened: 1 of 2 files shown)


ljharb:
http://caniuse.com/#search=transform doesn't work in IE 8 - how does this gracefully degrade in the older browsers that are indeed used to verify results for the compat table?

Perhaps a better initial step is to use event delegation - ie, bind 1 listener instead of N - and leave the logic in JS?

shospodarets:
Hey @ljharb.
It sounds very funny as that code was added about one year ago by @webbedspace in commit- https://github.com/kangax/compat-table/commit/22c1261c47f96ddef825cac0dd7cd87fa5ab003b
I just moved that code from [JS](https://github.com/malyw/compat-table/commit/36aa269112dbb6c659e8d127194f4d141f6a5157#diff-540574eb492134e4ba7f18f9b863e9d8L99)  to [CSS](https://github.com/malyw/compat-table/commit/36aa269112dbb6c659e8d127194f4d141f6a5157#diff-e6bd4811679b4fe024a3d5fb7632e032R161) as it's not a good practice to apply such styles in JS (separation of concerns).

Not sure about the level of old browser support for this project, but no complains during one year 😏

BTW, event delegation is actually used instead of previous binding clicks to each row.

webbedspace:
`transform`'s support is a non-issue because its only purpose is to rotate the dropdown triangle 90 degrees, which has no bearing on the ability of the table to display the results.

ljharb:
good call 👍

PR #667, opened 2015-10-21
New test for super() invocation in arrow function

claudepache:
There was already a test for `super.foo()` in arrow function, but not for `super()`.

Firefox Nightly currently supports the former, but not the latter. (In fact, arrow functions are disabled in derived constructors.)

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -778,6 +778,29 @@ exports.tests = [
       },
     },
     {
+      name: 'lexical super-constructor binding',
+      exec: function(){/*
+        var received
+
+        class B {
+          constructor (arg) {
+            received = arg
+          }
+        }
+        class C extends B {
+          constructor () {
+            var callSuper = () => super('foo')
+            callSuper()
+          }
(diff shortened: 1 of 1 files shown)


data-es6.js
```diff
@@ -778,6 +778,29 @@ exports.tests = [
...
+          constructor () {
+            var callSuper = () => super('foo')
+            callSuper()
+          }
+        }
+
+	return new C instanceof C && received === 'foo'
+      */},
+      res: {
+      	/* TBD */
```
kangax, line 800: We should really look into making a script that can connect to multiple browsers, run snippet of code, and report results back. Something like http://tutti.tobyho.com/ (which used to work back in the days). Perhaps using @saucelabs or @browserstack API's?

@jdalton I know you've worked with some of these; do you know if it's possible?

jdalton, line 800: With saucelabs you can specify a custom runner with the rest API to handle the payload. Totally doable.

Issue #673, opened 2015-10-26
Data Edge and Firefox switched when sorting on features


Studio384:
When you sort on features, I noticed that Edge and Firefox which have the first 2 collums, switch data. The numbered scores (like 12/15) are correct, however the individual features are switched between the twoo.

sarahkadar:
Not just the first two... all of them...

Studio384:
This is still not fixed.

sarahkadar:
I have a PR waiting for review @kangax :)

PR #678, opened 2015-10-30
Fix bug reported in issue #673

sarahkadar:
Re: issue #673. 

I think this is some legacy thing that got changed one point (the number of columns in a row) and then the sorting broke.


KTamas:
Oh, fun, this both fixes AND breaks the ES7 table. Will have to investigate more.

Issue #681, opened 2015-11-03
Remove string trimming from non-standard tests.


thomasfosterau:
I noticed today that the tests for String#trimLeft and String#trimRight are still in the non-standard compat table as well as being in the ES7 compat table. Given the string trimming methods are on track for ES7 (or are at least proposed), shouldn't they be removed from the non-standard tests?

ljharb:
They're still non-standard until they reach stage 4. They belong in both places until that time.

thomasfosterau:
Shouldn't more features be in the non-standard tests then? String${trimLeft, trimRight} are the same stage as Array#includes for example.

ljharb:
`trimLeft`/`trimRight` are non-standard because they are already implemented everywhere - in fact, the reason for the proposal is to normalize the defacto standard.

`Array#includes` only exists in browsers as a response to the proposal, so it belongs solely in the ES7 table.

PR #682, opened 2015-11-03
there are trimLeft and trimRight on page es7

connorhu:


ljharb:
Per #681, these must remain here until `trimLeft`/`trimRight` are stage 4 in the spec.

PR #686, opened 2015-11-04
Added FF44 to ES6/ES7 tables; Added FF42 to ES7 table, FF41 obsolete

sarahkadar:
For `let` changes see https://developer.mozilla.org/en-US/Firefox/Releases/44#JavaScript


data-es7.js
```diff
@@ -78,6 +78,16 @@ exports.browsers = {
   firefox41: {
     full: 'Firefox',
     short: 'FF 41',
+    obsolete: true,
```
kangax, line 81: Perhaps we need to combine es5, es6, es7 data into 1 file, to avoid having to update browsers' status like this (across multiple tables).

ljharb, line 81: Instead of that, we could just store browsers in a separate requireable file, no?

sarahkadar, line 81: I would lean towards having browsers in a separate file. Editing the giant es5/es6/es7 file always makes me feel like "oh shit I will break something", and separating them would also sort out the mess of missing versions/not using the latest version/etc which is also a mess right now.

sarahkadar, line 81: Oh, I sort of misunderstood. Yes, browsers in a separate file is great. But I think there should be some remedy regarding the mess in the test results.

PR #689, opened 2015-11-12
Mark MS Edge 13 as stable

dstorey:
…and remove IE from its alt text.


zloirock:
@dstorey why you changed **only** compiled html? I'll add it manually.

zloirock:
Removing `Internet Explorer` from the full name discards IE10-11 results for Edge. Need some more changes.

mgol:
See #691 for a followup.

PR #690, opened 2015-11-13
update for toshok/echojs@d8f51ed

toshok:
Lots of fixes for ejs.  Now that I have a separate test runner it should be easier to keep things up to date.


zloirock:
Please, fix linting warnings.

PR #691, opened 2015-11-13
[ES6] Mark Edge 12 as obsolete

mgol:
Edge 12 should have been marked as obsolete, similarly how Chrome & Firefox are
treated.

Refs #689
Refs 9f6da90bc91347c124030b5e12b8e549da53f5f7

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -89,7 +89,8 @@ exports.browsers = {
     family: 'Chakra',
     short: 'Edge 12',
     note_id: 'edge-experimental-flag',
-    note_html: 'Flagged features have to be enabled via "Enable experimental Javascript features" setting under about:flags'
+    note_html: 'Flagged features have to be enabled via "Enable experimental Javascript features" setting under about:flags',
+    obsolete: true,
   },
   edge13: {
     full: 'Microsoft Edge',
--- a/es6/index.html
+++ b/es6/index.html
@@ -115,7 +115,7 @@
 
           <th colspan="6" class="platformtype" id="compiler-header" style="background: #ffdfa4">Compilers/polyfills</th>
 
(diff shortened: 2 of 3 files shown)


zloirock:
I'm not sure about it. On my local machine still Edge 12, update center doesn't see any new updates. Anyway, it's Windows - after years people will have this version. For example, IE10 still not marked as obsolete.

mgol:
I don't think the comparison to IE 10 is fair. IE wasn't a browser with a rapid release schedule, Edge is. Also, on Windows 10, unless you have the Pro version that regular people won't have you can't even stop updates.

> On my local machine still Edge 12, update center doesn't see any new updates.

Maybe they throttle updates over a couple of days (@dstorey, @jacobrossi, @reybango: could you explain?). In any case, I doubt any significant number of people will have Edge 12 in the coming weeks.

mgol:
Also, Edge will now have frequent updates; the previous one was 3 months ago - I'm not sure if in 2 years you want to have 8 Edge versions displayed in the default view...

ljharb:
It'll be easy enough before 2 years to mark older/unused versions obsolete. I think it might be a bit too early to be making claims about Edge and its ability to autoupdate universally.

mgol:
I didn't expect these changes to be controversial but since they apparently are I extracted the test update portion to PR #692.

webbedspace:
Closing in agreement with @ljharb.

jacobrossi:
Sorry for chiming in late here. Actually, for our enterprise customers we're offering the ability for them to run either version N or N-1.  So at the moment that means EdgeHTML 12 and EdgeHTML13.  When EdgeHTML 14 comes out, then EdgeHTML 12 would be considered obsolete.

Issue #680, opened 2015-11-02
Object.observe is dead


mgol:
`Object.observe` is dead: https://mail.mozilla.org/pipermail/es-discuss/2015-November/044684.html

Should it be removed from the es7 list?

ljharb:
Just in case, I think we should wait for the next TC39 meeting, which is in 3 weeks. However, assuming it's withdrawn, we absolutely should remove it then.

mgol:
Sounds good.

## 

Michał Gołębiowski

rambo-panda:
Do not want to remove!

ljharb:
Object.observe is definitively withdrawn.

PR #700, opened 2015-11-30
removed Object.observe

crunchtime-ali:
The `Object.observe()` proposal has been withdrawn by its original author, Adam Klein (https://mail.mozilla.org/pipermail/es-discuss/2015-November/044684.html). Therefore I removed it all references from the ES7 table. #680

### DIFF
--- a/data-non-standard.js
+++ b/data-non-standard.js
@@ -1474,9 +1474,8 @@ exports.tests = [
 {
   name: 'Object.observe',
   link: 'https://arv.github.io/ecmascript-object-observe/',
-  category: 'draft',
   significance: 'large',
-  exec: function () {
+	exec: function () {/*
     var obj = {x: 1};
     Object.observe(obj, function(changes){
       var data = changes[0];
@@ -1485,7 +1484,7 @@ exports.tests = [
       }
     });
     obj.x = 2;
-  },
+  */},
   res: {
     android50: true,
     android51: true,
--- a/data-non-standard.js
+++ b/data-non-standard.js
@@ -1438,6 +1438,30 @@ exports.tests = [
   separator: 'after'
 },
 {
+  name: 'Object.observe',
+  link: 'https://arv.github.io/ecmascript-object-observe/',
+  category: 'draft',
+  significance: 'large',
+  exec: function () {/*
+    var obj = {x: 1};
+    Object.observe(obj, function(changes){
+      var data = changes[0];
+      if(data.name === 'x' && data.type === 'update' && data.oldValue === 1 && data.object.x === 2){
+        asyncTestPassed();
+      }
+    });
+    obj.x = 2;
+  */},
+  res: {
+    chrome33:        true,
+    chrome34:        true,
+    chrome35:        true,
+    chrome37:        true,
+    node:            true,
+    iojs:            true,
+  }
+},
+{
   name: 'Object.prototype.watch',
   link: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/watch',
   exec: function () { return typeof Object.prototype.watch == 'function' },
--- a/data-non-standard.js
+++ b/data-non-standard.js
@@ -20,6 +20,20 @@ exports.browsers = {
     short: 'io.js',
     platformtype: 'engine',
   },
+	node4:  {
+		full: 'Node.js',
+		family: 'Node.js',
+		short: 'Node<br>4.0',
+		platformtype: 'engine',
+		note_id: 'harmony-flag',
+	},
+	node5:  {
+		full: 'Node.js',
+		family: 'Node.js',
+		short: 'Node<br>5.0',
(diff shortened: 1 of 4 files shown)


zloirock:
I'm waiting November TC39 meeting notes. Also, I think, it should be moved to the [non-standard features table](http://kangax.github.io/compat-table/non-standard/).

crunchtime-ali:
Ok, let's wait for the meeting notes.
I'm fine with moving it to non-standard as well.

ljharb:
There's nothing in the notes that will change the outcome here - it's definitely withdrawn. However. I think moving it to nonstandard features is a good thing, since the data will still be useful, and who knows how quickly v8 will remove it.

crunchtime-ali:
I did some research. The browsers, engines and their versions in the non-standards table are very different from the ES*-ones.

None of them does support `Object.observe`.
What do you say to these?
- IE7-11 — no
- Firefox 28 — no
- Safari 3 - 8 — no,
- WebKit r72487 — no
- Konqueror 4.4, 4.9 — no
- Opera 10.10, 10.50-11.10 — no
- Besen — no
- Rhino — no
- Phantom — no
- Android 4.0, 4.1 — no
- io.js 1.0 — no (O.o support landed in 1.04)

What about CH 11+ and OP 15+ ? Chrome only started supporting `Object.observe()` in version 36 and Opera in 23.

ljharb:
@dj-hedgehog I just checked; `io.js` v1.0.0 absolutely also has `Object.observe`, as does node `v0.11` and later.

crunchtime-ali:
Thanks @ljharb. What to make of CH 11+ and OP 15+ ?

ljharb:
Can we just add the relevant browser version points to the nonstandard table? Eventually all the tables should share the same identical list of browser versions, but in the meantime we usually just add what is needed.

ljharb:
There's no need to wait for the notes - https://github.com/tc39/ecma262/commit/c67a000d7a2ff83481b04b58c10bb4e2e44bc7e0#diff-04c6e90faac2675aa89e2176d2eec7d8 is the removal of the proposal.

Please rebase this on the latest master, and move the data to the "non-standard" table (don't forget to `npm run build`) and we'll merge this.

crunchtime-ali:
Done, is it ok like that? I didn't change the old `res`-object in the data.

webbedspace:
One last thing - add browser entries for Chrome 33, Node and io.js, so that their results show up in the table. You can copy these from the es6 file.

zloirock:
And Android 5.0+

zloirock:
@webbedspace also `Object.observe` test (with `asyncTestPassed`) doesn't work in the non-standard table, have no ideas why and have no time to explore it. Any ideas?

crunchtime-ali:
I added Chrome 33, Android 5.0, Android 5.1, Node 4 and Node 5 to the non-standard list. However when generating the table it shows `no` for Object.observe() in Android 5.1 while it should be `yes`.

@zloirock What do you mean by `O.o test doesn't work`?

zloirock:
@dj-hedgehog actual Chromium:
default

crunchtime-ali:
The test-code works for me in Chrome 47.
I will try to debug the problem but am not sure when I will get to it.

data-non-standard.js
```diff
@@ -1438,6 +1472,34 @@ exports.tests = [
   separator: 'after'
 },
 {
+  name: 'Object.observe',
+  link: 'https://arv.github.io/ecmascript-object-observe/',
+  category: 'draft',
```
zloirock: Also, remove this line
default

crunchtime-ali:
Done and I spent an hour trying to find out why the test evaluates to true in the current browser. I compared my build-results to the current online one. They are exactly the same thing.

```
test(function(){try{var asyncTestPassed=global.__asyncPassedFn && __asyncPassedFn("33");try{return Function("asyncTestPassed","\nvar obj = {x: 1};\nObject.observe(obj, function(changes){\n  var data = changes[0];\n  if(data.name === 'x' && data.type === 'update' && data.oldValue === 1 && data.object.x === 2){\n    asyncTestPassed();\n  }\n});\nobj.x = 2;\n  ")(asyncTestPassed)}catch(e){asyncTestPassed=global.__strictAsyncPassedFn && __strictAsyncPassedFn("33");return Function("asyncTestPassed","'use strict';"+"\nvar obj = {x: 1};\nObject.observe(obj, function(changes){\n  var data = changes[0];\n  if(data.name === 'x' && data.type === 'update' && data.oldValue === 1 && data.object.x === 2){\n    asyncTestPassed();\n  }\n});\nobj.x = 2;\n  ")(asyncTestPassed)&&"Strict"}}catch(e){return false;}}());
```

ljharb:
@dj-hedgehog can this be rebased? Object.observe is definitively withdrawn, so I'd like it moved out of the es7 tests.

PR #704, opened 2015-12-07
Update for Chrome 47, 48 and 49

natorion:


ljharb:
What is M47/48/49? Are these v8 designations? This also seems like it could all be rebased down to a single commit.

natorion:
M*\* is the short name for Chrome **.

Rebasing makes sense though. I will squash them.

ljharb:
Thanks! In the future, please force push to the same branch name rather than opening up a second PR (#705) :-)

PR #705, opened 2015-12-07
Update Chrome 47, 48, 49

natorion:


webbedspace:
These Chrome destructuring results are invalid - Ch49 Canary currently supports destructuring in `var/let/const` declarations, but not in assignment statements, thus causing most of the destructuring tests to pass arbitrarily based on whether they involve this. I find it extremely unlikely that such an incomplete implementation would be allowed to propagate to stable versions, so I question their inclusion.

natorion:
Quoting from the page "Please note that some of these tests represent existence, not functionality or full conformance". I simply run the tests and write down the results. Currently destructuring minus assignment is shipped in Chrome 49. If this changes the results can be changed too.

The right thing would be to alter the tests, but this is a different story.

webbedspace:
Thing is, though, this table only lists results for stable releases of browsers. Beta versions are taken as safe assumptions about the support of the next stable release. Development versions like Firefox Nightly are not because several of their features (like classes) have, in the past, been disabled when they were promoted to beta or stable. Chrome 49 is currently only in the Canary stage, and while in the past we've trusted Canary as a prognosticator for the stable release, these nonsensical destructuring results give me reason to want to hold off on including the 49 column until it reaches beta.

ljharb:
Perhaps we should also improve the destructuring tests? "Only including existence not functionality" isn't a requirement, it's just the way the tests historically and commonly are written. The more precise they can be, imo, the better.

kangax:
We can totally remove that note btw. It's been there since the early times when I just started with `typeof .. !== underined` for a good chunk of features (back in the ES5 table). I haven't exactly envisioned us having such an extensive test suite since that's what official tests are for (sputniktests back in the days, official ecma262 suite, etc.) but it sort of happened and now people even use us to develop and measure their implementations (like that recent kinoma thingie). And I guess it's pretty amazing — the more chances/exposure we have of faulty implementations, the better.

sarahkadar:
@kangax heck, Microsoft cited the table in their latest Edge release which is a pretty nice thing :)

rossberg:
@webbedspace, fwiw, Chrome 49 _will_ indeed ship destructuring bindings but not (yet) destructuring assignment. The latter is quite a different construct, with different pattern syntax and substantially different semantics, so there is no strong reason to lump it together.

webbedspace:
I guess the only way to properly capture these results is to duplicate all of the destructuring tests for binding vs. assignment.

natorion:
I updated the commit. Because of the changes in the tests a simple squash is no longer possible.

PR #707, opened 2015-12-15
When we're missing data, put a `?` instead of "No".

ljharb:
This makes it very apparent when nobody's checked a specific engine/feature combo, so that someone knows to take the time to do so.

`?` is still counted as a "No" in the totals.


kangax:
Sure, sounds good to me!

webbedspace:
This is not feasible, actually. For the past year I've used the absence of a result line in `res` objects as a positive "No". This is necessary because otherwise the `data-es6.js` file would be 10x longer due to the need to include explicit `false` results for basically every platform. `false` should now only be used whenever a previous version of a platform had `true` or `flag`, and then withdrew support - or, when `Object.assign` is used to modify an object containing results that a particular subtest is based off. As I've said in the past: the `data` files are for human and machine consumption equally. I'm going to revert this.

If someone has knowingly not checked a certain platform, they should do what the non-standard table does and use `null` as the result for that platform.

ljharb:
@webbedspace the majority of platforms aren't checked by people who add results. What is the harm of having `data-es6` be long?

The default should be "I haven't checked this", since that's what humans do when they (since they, and not machines) enter the results.

ljharb:
I agree that it makes sense for, say, `"firefox30": true` to imply that the feature in future firefox versions, but that's orthogonal to this issue.

I'm not sure why you reverted this when 2 out of the 3 maintainers on this thread think it's a good idea.

ljharb:
Also, if an explicit `null` must be entered to indicate "hasn't been checked", then it's perfectly valid to add every combination of every browser version as `null` to every test, which makes the data file precisely as long. It's simply incorrect to have "No" for something that nobody's checked.

ljharb:
Perhaps instead of "firefox44" etc, we should have a "firefox" key, that points to an object with a "version" value?

webbedspace:
> I'm not sure why you reverted this when 2 out of the 3 maintainers on this thread think it's a good idea.

I reverted it because the live table was currently displaying incorrect information (almost every single "No" had been changed to a "?" because of it) and it seemed unwise to leave it that way until a better solution was reached.

My opinion is that having explicit "?" and implicit "No" is _far_ superior to having implicit "?" and explicit "No" because right now, almost all of the tests on the ES6 table have already been verified for every listed platform. Thus, implicit "?" results will only need to be added for future ES6 tests (which, considering how comprehensive they currently are, would be vanishingly few) and ES7 tests. Moreover, as these tests are verified, the explicit "?" results are removed - the files become smaller, not greater, as time progresses.

(I do acknowledge that in the past I and others have oft been cavalier in adding new tests' results, and leaving certain platforms as "No" until proven "Yes", but that's not a conduct that I wish to continue in the future, especially now that "?" will be displayed on the table instead of a vacant gap.)

ljharb:
@webbedspace fwiw that could have been achieved without a revert simply by changing the default "?" to a "No".

Also I'm highly skeptical of "vanishingly few" because engines still have a litany of ES5 and ES6 bugs (most of which aren't tested by the table).

PR #708, opened 2015-12-16
[ES6] adding a test for iterable arguments objects.

ljharb:
https://bugzilla.mozilla.org/show_bug.cgi?id=1232929

I'll be happy to fill in the undefineds if someone wants to provide me with any of the missing test results.


zloirock:
We have [this test](http://kangax.github.io/compat-table/es6/#test-well-known_symbols_Symbol.iterator,_arguments_object). Possible extend it.

ljharb:
Where? I saw "generic iterable" but nothing specifically for `arguments` under "for..of loops".

zloirock:
_well-known symbols -> `Symbol.iterator`, `arguments` object_. Many over features required iterables, but I don't think makes sense add `arguments` object to each section.

ljharb:
That's a good point. However, `node012` is marked `true` for that test, and it currently fails :-/

I'll close this, and update the results of that test.

Issue #603, opened 2015-08-21
Remove async generators and RegExp.escape test from ES7


thomasfosterau:
The async generators proposal has been withdrawn (see [this commit](https://github.com/tc39/ecma262/commit/6f35ca7aa0415e3992af304987380b1f5cbab191)) and probably needs to be replaced by a test for [Observables](https://github.com/zenparsing/es-observable).

Additionally, it seems as though RegExp.escape has been rejected (see [this commit](https://github.com/tc39/ecma262/commit/16ef4b6512bd5a975bb8dce882279212e14ebe1e)) and should be removed.

ljharb:
I'd like to wait until the replacement proposal for "escaping user input to make a regex" arrives - I'm not as convinced that it was rejected "definitively" as-is by the committee.

As for Observables, that would definitely be a good thing to add, as it's at stage 1.

thomasfosterau:
My impressions from reading [this](https://github.com/rwaldron/tc39-notes/blob/master/es7/2015-07/july-28.md#62-regexpescape) and having a general poke around is that RegExp.escape at least needs to be moved back to the 'pre-strawman' phase. It sounds to me as though it will be replaced by a template string-based proposal.

I'll have a session of cross-referencing the ES7 compat-table with the tc39/ecma262 repo and send in a  pull request in a few hours with Observables included (and whatever else has changed).

ljharb:
You're right, I'll drop the stage now.

Please try to make one PR for each change, rather than bunching them up :-)

thomasfosterau:
Will do!

zloirock:
`RegExp.escape` definitely should be removed from the table. I also leave it in `core-js` as a non-standard feature, but `RegExp.escape` in the table can confuse users and they can start to use it as a "part of ES7".

zloirock:
> I'm not as convinced that it was rejected "definitively" as-is by the committee.

https://github.com/zloirock/core-js/issues/116 it's enough for me

kangax:
Does anyone know what kind of template string -based solution `RegExp.escape` was "replaced" with?

zloirock:
https://github.com/mikesamuel/regexp-make-js , but it's still not a proposal.

ljharb:
fwiw, I still believe that a template string approach isn't the right one, and I hope to revisit RegExp.escape in the future.

kangax:
[...]

PR #717, opened 2016-01-01
[ES7] Cleaning pre-strawman

zloirock:
- Removed `RegExp.escape`, more info and discussion in #603. I also like this method, but it was rejected. This method in the table can confuse users.
- Comprehensions moved to the non-standard table. Looks like it's finally dead. In most places notes changed from the `es7` to the `non-standard` (for example, [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Array_comprehensions)). It was removed from the `babel` 6.

If there are no objections - I will merge it.


ljharb:
Agreed on all points, LGTM

PR #718, opened 2016-01-02
fix floatThead firefox header misalignment + refactor some related code

mkoryak:
main things in this pr:
- fix that ff header misalignment 'bug' by changing table `width:100%` to `min-width:100%`
- always show percentages while scrolling
- refactor some js code make calling floatThead less weird

I tried to not go too crazy with refactoring, but you might feel I did more than this PR would warrant - let me know and ill make 2.


.gitignore
```diff
@@ -3,3 +3,4 @@ es*/compilers/*
 .DS_Store
 .ruby-version
 _site/
+.idea/
```
kangax, line 6: Out of curiosity, what is this for?

master.js
```diff
@@ -44,67 +44,43 @@ $(function() {
...
-    };
-    return fn;
-  })(table.floatThead);
-
-  // Set up the Show Obsolete checkbox
-  $('#show-obsolete, #show-unstable').on('click', function() {
-    setTimeout(function() {
+  if (false) { //set to true to disable floating headers
+    // to fully remove, search for floatThead and remove all
+    $.fn.floatThead = function(){ return this };
```
kangax, line 50: Nice :)

master.js
```diff
@@ -446,7 +424,7 @@ $(function() {
                           : 'engines';
 
     // First, hide the platformtype bar if we're sorting by features.
-    $('.platformtype')[(sortByFeatures || sortByFlaggedFeatures) ? 'hide' : 'show']();
+    $('body')[(noPlatformtypeBar || sortByFeatures || sortByFlaggedFeatures) ? 'addClass' : 'removeClass']('hide-platformtype');
```
kangax, line 427: I know we already had it like this but should probably use `toggleClass` here at some point

kangax:
I love the animated transition on percentages when you scroll! Feels nice, thanks @mkoryak

.gitignore
ljharb, line 6: blocks IntelliJ files - imo it should be in the user's global gitignore rather than an individual project's, but it can't hurt.

kangax, line 6: Gotcha, thanks

mkoryak:
no problem, thanks for using my plugin, though you had no choice, since all the other ones suck ;)

master.js
mkoryak, line 427: ah, I see, toggle class can take a `state`, that would be cleaner

PR #720, opened 2016-01-03
fix the column misalignment bug when highlighting a column

mkoryak:
also refactored the sort function to be a bit more readable.

I tested the sorting against prod on all tables, it matches. You should double check though.

### DIFF
--- a/master.js
+++ b/master.js
@@ -67,7 +67,7 @@ $(function() {
 
 
   table.on("floatThead", function (evt, isFloating, $container) {
-    $container[isFloating ? 'addClass' : 'removeClass']("floating-header");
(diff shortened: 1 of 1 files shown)


master.js
```diff
@@ -267,10 +269,12 @@ $(function() {
       if (!supertest.length) {
         supertest = elem.prev();
       }
-      supertest.click();
+      supertest.triggerHandler('click');
     }
 
     table.addClass('one-selected').insertBefore('#footnotes');
+    win.scrollLeft(left);
+    table.triggerHandler('reflow');
```
mkoryak, line 277: this is the only thing you really need to fix this bug

PR #726, opened 2016-01-14
Intl's resolvedOptions().timeZone results better defined.

paulirish:
Turned some `?`s into definitive `No`s.


zloirock:
Thanks, confirmed.

PR #727, opened 2016-01-14
ES7: FF42 obsolete, added FF45

sarahkadar:
FF45 supports `Object.values` and `Object.entries`

As an aside, the calculated score is now 15%, but it shows 14% for "Current browser". Weird.

### DIFF
--- a/data-es7.js
+++ b/data-es7.js
@@ -106,6 +106,7 @@ exports.browsers = {
   firefox42: {
     full: 'Firefox',
     short: 'FF 42',
+    obsolete: true,
   },
   firefox43: {
     full: 'Firefox',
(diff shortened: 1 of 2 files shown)


zloirock:
> FF45 supports `Object.values` and `Object.entries`

MDN says it will be available only in nightly builds. I'm not sure it should be added. Or need to add a note about it.

sarahkadar:
Ah, my mistake, then. Maybe there should be a separate flag for nightlies?

ljharb:
These functions are still behind a flag in Firefox while they convince themselves it's performant enough. Nightlies are nice in that they offer a preview of what's coming, but if it's not going to land in the stable browser, there's no benefit in displaying it as "supported" (same with all of Chrome's flagged features).

If you can amend your commit to preserve the `false` for these values, we'd be happy to merge in the new version identifier.

sarahkadar:
Commit updated. Left a comment at those `false` flags for future reference.

zloirock:
Thanks.

PR #735, opened 2016-02-01
updated closure to v20160125

oldergod:


zloirock:
Please, fix linting: 51+ (!!!) duplicate keys.

oldergod:
Sorry, I tried to automate the fillings but did not go that well. Should be good now.

zloirock:
Stop... I checked just syntax. Why do you add standard library features? Looks like need revert it.

oldergod:
I am not following, which features are you talking about?

zloirock:
Typed Arrays, Map, Set, etc...

oldergod:
Because Closure did add many polyfills recently : https://github.com/google/closure-compiler/tree/master/src/com/google/javascript/jscomp/js/es6

zloirock:
But definitely not typed arrays. [You should run tests in an ES5-only engine, for example, IE9.](https://github.com/kangax/compat-table#in-order-to-test-compilers) Reverted. Please, fix the results.

oldergod:
Alright, I'll look into that. thank you.

zloirock:
BTW looks like closure test page broken - compile error for all tests.

oldergod:
If I'm not mistaken, closure test page has never been used and this test suite has been built instead https://github.com/teppeis/closure-compiler-es6-compat-table

PR #736, opened 2016-02-02
Chrome 48 does not support Symbol.toStringTag

claudepache:
Symbol.toStringTag seems to be supported by Chrome since v49 only


zloirock:
Yep, just flagged feature. Thanks.

PR #741, opened 2016-02-04
Update Closure Compiler to v20160201

teppeis:
- Improve Spread Operator
- Improve Iterator and Generator
- Partial support for Symbol
- and some fixes

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -47,7 +47,7 @@ exports.browsers = {
     platformtype: 'compiler',
   },
   closure: {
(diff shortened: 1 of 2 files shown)


zloirock:
@teppeis recently [some polyfills](https://github.com/google/closure-compiler/tree/master/src/com/google/javascript/jscomp/js/es6) were added to Closure. What about them?

teppeis:
@zloirock yes, but the almost polyfills can not be enabled with CLI options in this version.

teppeis:
thanks!

PR #742, opened 2016-02-06
Make JSX obsolete

teppeis:
react-tools and JSTransform was deprecated and switched to Babel last year.
- [Deprecating JSTransform and react-tools | React](https://facebook.github.io/react/blog/2015/06/12/deprecating-jstransform-and-react-tools.html)

So please make it obsolete and hide.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -55,7 +55,7 @@ exports.browsers = {
   jsx: {
     full: 'JSX',
     short: 'JSX',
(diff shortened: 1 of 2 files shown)


zloirock:
Agree. In this case, [in the ESNext table](http://kangax.github.io/compat-table/esnext/) - too.

teppeis:
thanks :smile_cat:

PR #749, opened 2016-03-03
Updated Chrome 50 and 51

natorion:
Updates based on 50.0.2657.0 dev


zloirock:
Strange, in the canary 50 before updating to 51 without a flag I saw 91%.

data-es6.js
```diff
@@ -13910,6 +13931,7 @@ exports.tests = [
       */},
       res: {
         xs6:         false,
+        chrome50:    true,
```
zloirock: For this Chrome 50 build - `yes`, but not for the current 51 canary build:
default
Please, disable it for Chrome 51.

littledan: Thanks @zloirock, I'll look into it. I know Chrome 48 threw some exceptions that broke some sites, and Chrome 49 will revert this in a way that the spec does not describe, but I did not know of a change here.

zloirock: @littledan looks like `RegExp#source` related this problem https://github.com/tc39/ecma262/pull/263

littledan: Yes, and this "fix" actually is backported to 49. Unfortunately, I'm not confident that the web can support as strict RegExp semantics as TC39 called for. My plan is to collect data and revert the workaround once I have more confidence in the web compatibility of these changes.

natorion: Good catch. I will fix the data.

natorion:
Fixed the data.

PR #752, opened 2016-03-03
Update to WebKit r197500

Constellation:


data-es6.js
```diff
@@ -4799,6 +4811,7 @@ exports.tests = [
         firefox44:true,
         xs6:      true,
         ejs:      true,
+        webkit:      true,
```
zloirock: Duplicate key. Please, fix it.

Constellation: Thx, fixing.

Constellation:
Updated :)

PR #758, opened 2016-03-13
[ES6] Add tests for const for-loop interation scope

### LINKED ISSUE #571: Test `const ... in ...`
ES6 supports the following construct:

``` js
var keys = [];
for (const i in {a: 2, b: 3}) {
    setTimeout(function () {
        keys.push(i);
    });
}
console.assert(keys.indexOf('a') > -1);
console.assert(keys.indexOf('b') > -1);
```

This seems to be the only situation where `const` is allowed without explicit initializer so it might be worth checking it.

Chrome 43 passes the test, IE11 & Edge don't.

P.S. A cleaner test would use `for-of` but that's yet another feature I didn't want to mix in.

mgol:
Fixes #571


data-es6.js
```diff
@@ -1059,6 +1082,28 @@ exports.tests = [
       },
     },
     {
+      name: 'for-loop iteration scope',
```
ljharb: can you add explicit false's to engines that are known to fail, rather than leaving them blank? specifically, v8 iirc fails this

ljharb: (also this title should be `for-in loop`, not `for-loop`)

data-es6.js
```diff
@@ -1190,6 +1261,31 @@ exports.tests = [
         jxa:         true,
       },
     },
+    {
+      name: 'for-loop iteration scope (strict mode)',
```
ljharb: for-in loop

data-es6.js
mgol: I copied this from the `let` section & just changed `let` to `const` (after removing parts that don't apply to `const`). Should I apply those changes in both places?

mgol: V8 doesn't fail this, though, as you can see from the `chrome49` entry. :) AFAIK only Firefox & Safari do (on the browser side) and it's already fixed in WebKit Nightly.

ljharb: ah, perhaps it's fixed in chrome 49. it fails in chrome 48 :-)

either way tho, it'd be ideal to a) say "for-in loop" on every for-in loop, and b) include explicit `false`s for every known failure

mgol: I added the renaming in a separace commit. As for `false` entries it seems more involved as I don't see a single `false` entry related to a browser without a future `true` or `flag` entry. It seems a separate script-generated commit that would fill in all those missing entries would be best here. At least for web browsers it should be quite safe as their scores are checked by many people so someone would notice a mistake.

mgol:
@ljharb Do you have any further remarks?

ljharb:
No, this looks good (needs a rebase tho)

mgol:
Rebased.

PR #765, opened 2016-03-20
remove unused variable 'x'

thefourtheye:


zloirock:
1. Usage this parameter - detection basic support default parameters, without it this test pass engines without support default parameters.
2. Simple updating HTML makes no sense because it's generated from `data-es6.js`.

thefourtheye:
@zloirock Oh, right. Got it. Thanks.

PR #769, opened 2016-03-28
[ES6] Update Closure Compiler to v20160315

### LINKED ISSUE #747: Update Closure Compiler
I suspect compat-table is testing closure compiler with the default options without using externs to rewrite the polyfills (which is equivalent to babel-polyfill), that's why you see all the built-ins and built-ins extensions not supported in closure compiler.

@dimvar and @shicks, can you guys help @teppeis to level up the results of closure compiler?

teppeis:
Enable polyfill with `--rewrite_polyfills`.
Fix #747.


zloirock:
:+1:

PR #770, opened 2016-03-29
fix test 460

ghost:
test 460 was using the bitwise `&=` operator, resulting in `1` instead of `true`, and did not return the `passed` variable


ghost:
This change makes Chrome pass the test where previously it was failing.  I'm not sure how to test all the browsers and update the results.

zloirock:
> test 460 was using the bitwise `&=` operator, resulting in `1` instead of `true`

It's not principal.

> did not return the `passed` variable

Thanks.

PR #777, opened 2016-04-12
Edge14 updates

dstorey:


data-es6.js
```diff
@@ -1689,6 +1691,7 @@ exports.tests = [
         chrome49:    true,
         webkit:      true,
         edge12:      flag,
+		edge14:      true,
```
ljharb: all these tabs need to be spaces instead

dstorey: there must be mixed tab/spaces, as I copy pasted from the ES6 file to make sure…I'll update.

ljharb: if you could just globally replace any tabs that are in there in this PR, that would be great :-) thanks!

PR #780, opened 2016-04-15
Fixed a bad footnote reference

phistuck:


webbedspace:
This commit needs to patch `data-esnext.js`, not the compiled HTML.

phistuck:
Oh.
Well, it looks like it already has a note for that, but it is not rendered for some reason.

chicoxyzzy:
proper fix https://github.com/kangax/compat-table/pull/836

PR #782, opened 2016-04-20
Added Safari Technology Preview for ES6/ESNext/INTL/non-standard

sarahkadar:
I'm guessing there is no point for ES5?


lin7sh:
🤘This is so awesome! 🤘

zloirock:
For other browsers we have 2-3 unstable versions, so adding the second unstable version for Safari makes sense.

> I'm guessing there is no point for ES5?

Looks like Safari and last WebKit has some difference in the ES5 table. If Safari TP has a difference with previous Safari versions it should be added.

Also, I think, should be added release version.

sarahkadar:
@ziorock "Also, I think, should be added release version."

I don't understand what you mean. 

I'll add ES5.

sarahkadar:
There was something fishy with webkit, the es5 file missed a bunch of values even though the HTML had it. I can't test webkit myself so I took the values from the HTML and put it back to the data-es5.js file. 

(sidenote: anyone knows why the f$%# webkit fails on my MBP running 10.11.4? I download it, put it in /Applications, launch it and it launches regular safari with the safari engine instead. Every time. Safari TP works without issues.)

zloirock:
> I don't understand what you mean.

[Release 1, release 2, release 100500.](https://developer.apple.com/safari/technology-preview/release-notes/) For example, for WebKit specified `r199524`.

sarahkadar:
@zloirock done

zloirock:
Thanks.

PR #783, opened 2016-04-23
Use try/catch blocks so thrown TypeErrors do not give false test results

jonknapp:
When running the tests in Node a TypeError is thrown and the result is never able to return correctly. This allows the errors to be captured and a successful result returned.

### DIFF
--- a/data-es5.js
+++ b/data-es5.js
@@ -1050,9 +1050,16 @@ exports.tests = [
   {
     name: 'undefined',
     exec: function () {/*
-      undefined = 12345;
-      var result = typeof undefined == 'undefined';
-      undefined = void 0;
+      var result = false;
+      try {
+        undefined = 12345;
+        result = typeof undefined == 'undefined';
+        undefined = void 0;
+      } catch (error) {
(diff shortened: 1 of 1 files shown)


zloirock:
Emm?
default

jonknapp:
I guess I should give some context to the issue I encountered. I'm working on a website to display the js syntax compatibility for specific versions of Node running on AWS Lambda. Instead of recreating a suite of tests, I utilized the `data-*` files that are in this repository, though I had to write my own testing harness around them for collecting results.

I am running the test methods in a function wrapper through `eval` with `try/catch` blocks to check for errors that would occur because of parsing issues, etc.  When ran in that condition, a `TypeErrror` is thrown which does not allow the function to return a result and it appears to fail.

zloirock:
`TypeError` here is a correct behavior for the strict mode, in those tests tested behavior in the sloppy mode, so you should run those tests in the sloppy mode. Use `Function` constructor instead of `eval`.

jonknapp:
Thanks for the advice. Feel free to close out this request if it's not beneficial to the test suite.

PR #784, opened 2016-04-28
[ES6] Update WebKit to r200136

Constellation:
Only the test for RegExp.prototype.test for Proxy 'get' call is failing :)


zloirock:
As I wrote, please, update WebKit results for other tables if it's required.

sarahkadar:
fixed in #787

PR #787, opened 2016-04-28
Webkit r200167 (April 28, 2016), all tables updated

sarahkadar:


zloirock:
The difference in ES6 results with @Constellation results looks strange. I haven't at hand any Apple device. Can someone check it?

sarahkadar:
@zloirock i've tested it with his revision (r200136) just now and indeed, that version confirms his results. Looks like they are in the process in finishing up things and in some builds they work, some they don't...

Just now an even newer webkit revision came out (r200184). In that version, "Regexp is subclassable" is 4/4 again, but "Proxy internal get calls" is down to 34/36 instead of 35/36.

So... I guess this will keep fluctuating for a while longer. I don't think it's worth for me to change the results once again for the 6-hours-older build, because who knows what will change tomorrow... we can only ever get slightly imperfect results in the table, until we have like a webkit-buildbot-compat-table-tester or something :)

zloirock:
Ok. Thanks.

PR #789, opened 2016-04-30
Add test: Intl.DateTimeFormat should allow IANA timezone names to be used

Daniel15:
Version 1.0 of the Intl spec (ECMA-402) specified that only UTC needed to be supported and all other timezones were optional. Version 2.0 of the spec modified this to say that IANA timezone names **must** be supported. Some browsers are still using the 1.0 version of this spec.

References:
- Old spec, 12.1.1.1 InitializeDateTimeFormat, step 16c: http://www.ecma-international.org/ecma-402/1.0/#sec-12.1.1.1
  
  > If tz is not "UTC", then throw a RangeError exception.
- New spec, 12.1.1 InitializeDateTimeFormat, steps 19c and 19d: http://www.ecma-international.org/ecma-402/2.0/#sec-InitializeDateTimeFormat
  
  > If the result of IsValidTimeZoneName(tz) is false, then Throw a RangeError exception.
  > Let tz be CanonicalizeTimeZoneName(tz).
  > 
  > The IsValidTimeZoneName abstract operation verifies that the timeZone argument (which must be a String value) represents a valid Zone or Link name of the IANA Time Zone Database.
- Firefox bug report: https://bugzilla.mozilla.org/show_bug.cgi?id=1266290

To be verified:
- Edge 12 support. It _should_ be supported now (as per https://github.com/Microsoft/ChakraCore/pull/19), but I can't figure out how to actually get a copy of Edge 12 :stuck_out_tongue: 
- Webkit site only has Mac OS X builds, so I couldn't test Webkit. Likewise, Safari is Mac-only. Would appreciate if someone could test these for me.

### DIFF
--- a/data-esintl.js
+++ b/data-esintl.js
@@ -707,6 +707,33 @@ exports.tests = [
         ios7: false,
         ios9: false
       }
+    },
+    {
+      name: 'accepts IANA timezone names',
+      exec: function() {/*
+        try {
+          new Intl.DateTimeFormat('en-US', {
+            timeZone: 'Australia/Sydney',
+            timeZoneName: 'long'
+          }).format();
+          return true;
+        } catch (e) {
+          return false;
+        }
+      */},
+      res: {
+        ie9: false,
+        ie10: false,
+        ie11: false,
+        ie11: false,
+        edge: false,
+        edge14: true,
+        firefox29: false, // Firefox bug #1266290
+        chrome24: true,
+        chrome29: true,
+        node12: true,
+        node4: true,
+      }
     }
   ],
 },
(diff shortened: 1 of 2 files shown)


sarahkadar:
@Daniel15 

Safari 9.1: no
Safari TP3: yes
Webkit r200167: yes

sarahkadar:
For edge 12, you can get ISOs for the original win10 release from Microsoft, at least I can get it (as a partner with ActionPack) and install that but it's a a bit cumbersome to go through all that.

zloirock:
Thanks.

PR #792, opened 2016-05-06
[ES6] Upgrade WebKit to r200505

Constellation:
Finally, achieve 100% compatibility.


Constellation:
@saambarati

zloirock:
...and we have may be unstable, but winner :D

PR #816, opened 2016-06-15
Remove invalid test for `Date.parse`

jugglinmike:
The behavior of `Date.parse` for formats not recognized by the
specification are implementation-dependent. Implementers are free to
extend it in any manny they wish; returning a value other than `NaN` is
therefore not a spec violation.

ES5 includes the following text in its definition of `Date.parse` [1]:

> The function first attempts to parse the format of the String
> according to the rules called out in Date Time String Format
> (15.9.1.15). If the String does not conform to that format the
> function may fall back to any implementation-specific heuristics or
> implementation-specific date formats. Unrecognizable Strings or dates
> containing illegal element values in the format String shall cause
> Date.parse to return NaN.

ES2015 contains very similar instructions [2]:

> The function first attempts to parse the format of the String
> according to the rules (including extended years) called out in Date
> Time String Format (20.3.1.16). If the String does not conform to that
> format the function may fall back to any implementation-specific
> heuristics or implementation-specific date formats. Unrecognizable
> Strings or dates containing illegal element values in the format
> String shall cause Date.parse to return NaN.

[1] https://es5.github.io/#x15.9.4.2
[2] http://www.ecma-international.org/ecma-262/6.0/#sec-date.parse

### DIFF
--- a/es5/index.html
+++ b/es5/index.html
@@ -2235,56 +2235,6 @@ try {
 <td class="yes" data-browser="android44">Yes</td>
 <td class="yes" data-browser="ios78">Yes</td>
 </tr>
-<tr class="subtest" data-parent="Miscellaneous" id="test-Miscellaneous_Date.parse_produces_NaN_for_invalid_dates"><td><span><a class="anchor" href="#test-Miscellaneous_Date.parse_produces_NaN_for_invalid_dates">&#xA7;</a>Date.parse produces NaN for invalid dates</span><script data-source="function () {
-var brokenOnFirefox = !isNaN(Date.parse(&apos;2012-04-04T24:00:00.500Z&apos;));
-var brokenOnIE10 = !isNaN(Date.parse(&apos;2012-12-31T24:01:00.000Z&apos;));
-var brokenOnChrome = !isNaN(Date.parse(&apos;2011-02-29T12:00:00.000Z&apos;));
-return !brokenOnFirefox &amp;&amp; !brokenOnIE10 &amp;&amp; !brokenOnChrome;
-    }">test(
-function () {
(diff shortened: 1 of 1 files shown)


ljharb:
This data would need to be removed from `data-es5.js`, not just from the generated file.

However, you can see by the data that it in fact is consistent enough across browsers that it could be said to be a defacto standard, the spec text just hasn't caught up yet (this is true for most "implementation dependent" things in the spec).

I do not want this test removed - I think it provides useful information. The compat table isn't test262, and need not have such a strict guideline for inclusion of tests.

jugglinmike:
Got it. I was thinking this test was included from a misinterpretation of the
spec.

I think I got tripped up on the word "compatibility." It's also strange to see
this test labeled specifically under "ES5," as though the behavior were
somehow related to that publication.

There is definitely value in having visibility into details like this, but
maybe these tests belong in a separate tab. That would allow for a more
consistent interpretation of "failures", and it might also make it easier to
accept additional contributions along these lines (for example, the precision
of built-in Math values/methods).

Anyway, this patch is clearly not the solution, so I'll close the pull request.

littledan:
@ljharb I don't think this is a de facto standard. The fact is that there is large disagreement among implementations about what the "legacy" non-ISO 8601 Date.parse semantics should be. It was an explicit decision to not put them in ES2015. Fortunately, work is now ongoing to find a standard semantic.  You can see some progress in https://bugzilla.mozilla.org/show_bug.cgi?id=1274354 . I'm not sure what the "non-standard" tab is intending to measure, but returning NaN for cases like this is neither a de jure nor de facto standard/dependable behavior.

PR #817, opened 2016-06-15
Missing "}" in tests "S.p.{starts,ends}With throws on RegExp"

claudepache:


ljharb:
whoops, thanks! Can you also `npm run build` to regenerate the ES6 html?

PR #818, opened 2016-06-16
ES5: update note about strict getters in FF

claudepache:
The bug mentioned in that note was fixed in FF46. References:
- https://developer.mozilla.org/en-US/Firefox/Releases/46
- https://bugzilla.mozilla.org/show_bug.cgi?id=603201


zloirock:
So maybe makes sense to add FF46 without this note to this table?

claudepache:
Superseded by #841

PR #821, opened 2016-06-24
Update data es-next for Safari TP and WebKit Nightly

chicoxyzzy:


zloirock:
What's with features, removed in your PR? I can't check it right now, but it looks strange.

PR #790, opened 2016-05-03
[ES6] Update Chrome 52 results

chicoxyzzy:
Version 52.0.2723.0 canary (64-bit) now supports TCO without flags

PR #823, opened 2016-06-25
Safari10

chicoxyzzy:
- add Safari 10
- fix es-next features support for Safari TP and WebKit Nightly


data-esnext.js
```diff
@@ -305,6 +305,11 @@ exports.browsers = {
     short: 'SF 9',
     obsolete: false,
   },
+  safari10: {
+    full: 'Safari 10',
+    short: 'SF 10',
+    obsolete: false,
```
zloirock: `unstable: true`

zloirock:
I don't see all es6 diff, but results are incorrect. For example, you enabled TCO for Chrome 52, but it's available only with the flag - #790

zloirock:
Also, can you avoid merge commits in PRs in the future? It complicates checking and pollutes the repo.

chicoxyzzy:
commit with chrome was mistakenly taken from another branch (it was in my [`gh-pages`](https://github.com/kangax/compat-table/pull/790) from which I created new branch). Removed it back.

> Also, can you avoid merge commits in PRs in the future? It complicates checking and pollutes the repo.

hmm what if I squash all commits in this PR?

chicoxyzzy:
> hmm what if I squash all commits in this PR?

realized it's bad idea

ljharb:
Rebasing down to the proper number of commits such that each commit represents an atomic, conceptual change, is a great idea :-)

chicoxyzzy:
fixed

Issue #824, opened 2016-06-26
Mark old browsers as obsolete


chicoxyzzy:
According to compat-table
- Firefox has 3 stable versions: 38, 45, 46
- Safari has 5 stable versions: 6.1 / 7, 7.1 / 8, 9

Can we mark Firefox 38 and Safari 6.1 / 7 as obsolete?

zloirock:
FF38 is ERS, so not now. I'm not sure how supported Safari.

UPD: Stop... Yep, [looks like FF38 support ended some days ago](https://wiki.mozilla.org/RapidRelease/Calendar). And FF46 too :)

chicoxyzzy:
So we should mark FF38, FF46 and Safari 6.1 / 7 as `obsolete`, right? I want to file a PR

zloirock:
FF - yes, but about Safari see #740.

chicoxyzzy:
AFAIK lower version of OS X which Apple still supports is OS X 10.9 Mavericks. Bundled Safari version is 7. Since September 18, 2014 there are security updates only for Safari 7.1 branch.

chicoxyzzy:
Also latest officially supported version of iOS is 9

chicoxyzzy:
should we close this issue now?

PR #825, opened 2016-06-26
Update obsolete browsers

chicoxyzzy:
We still need to reach consensus before merging this.

Previous discussion #824


zloirock:
FF47 should be marked as stable.

Issue #822, opened 2016-06-24
ES2016 / ES7 tab


chicoxyzzy:
ES2016 was released recently. Do we need to add new tab and remove ES2016 features from `next` tab?

ljharb:
Each year from 2016 on is going to be relatively small - in N years we'd end up with N more tabs. Perhaps we should have one tab for "stage 4 things" - which would now contain ES2016 (+ misc), and ES2017 (+ misc), and then another for "proposals" with everything else?

chicoxyzzy:
> stage 4 things

IMO "Post-Harmony" sounds better

ljharb:
Technically proposals are post-Harmony too. That name isn't particularly meaningful.

chicoxyzzy:
According to [TC39 Process](https://tc39.github.io/process-document/) ES2016 features are not technically "stage 4 things" as well

ljharb:
@chicoxyzzy that document does not in any way speak to ES2016 features, nor does it say that a feature leaves stage 4 when it is included in a standard revision. Regardless, the term "stage 4 things" was meant to be a broad term to describe, in general terms, finished features.

chicoxyzzy:
I think we can close this because of #827

PR #827, opened 2016-06-28
move es2016+ and stage 4 things to new tab

chicoxyzzy:
Previous discussion https://github.com/kangax/compat-table/issues/822


chicoxyzzy:
I've choose name "Accepted" for new tab but I'm open to any other suggestions

data-esnext-accepted.js
```diff
@@ -0,0 +1,1243 @@
+// exports browsers and tests
+// new browsers should only be added once they have at least a single 'true' result.
+exports.name = 'ES Next Accepted';
```
ljharb: I think a more accurate name here would be `ES 2016+` or `Finished ECMAScript Proposals`, per https://tc39.github.io/process-document/ - "Accepted" isn't a term that's usually used.

chicoxyzzy: @ljharb I think `ES2016+` is fine

webbedspace:
Extra favour: move the `exports.browsers` object into a separate file `require`d by both esnext.js files. Will save us all in the long run.

chicoxyzzy:
Should I do something else here?

PR #832, opened 2016-07-10
Update Closure Compiler to v20160619

teppeis:


data-es6.js
```diff
@@ -16662,6 +16665,7 @@ exports.tests = [
         return set instanceof S && set.has(123);
       */},
       res: {
+        closure:     true,
```
zloirock, line 16668: That looks strange. Can you show me `closure` output for `class S extends Set {}`?

teppeis, line 16668: @zloirock 

``` js
module.exports = function() {
  var a = function(a) {
    $jscomp.Set.apply(this, arguments);
  };
  $jscomp.inherits(a, $jscomp.Set);
  var b = new a;
  b.add(123);
  b.add(123);
  return b instanceof a && b.has(123);
};
```

and 600 lines of runtime code.

teppeis, line 16668: This is full output: https://gist.github.com/teppeis/2911c795a56e8f82e95b62ce18219d55

teppeis, line 16668: Closure Compiler doesn't support "Class/Map/Set: constructor requires new" now.
So `$jscomp.Set.apply(this, arguments)` doesn't throw an error.

zloirock, line 16668: Hmm, interesting solution. So closure forced uses polyfills for builtins subclassing... Looks like need extend tests for that.

teppeis:
thanks! 😸

PR #835, opened 2016-07-16
update Chrome es2017 support

chicoxyzzy:
`Proxy "ownKeys" handler, duplicate keys for non-extensible targets` is marked as supported

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -853,6 +853,7 @@ exports.tests = [
      return Object.getOwnPropertyNames(P) + '' === "a,a,b,b";
      */},
     res: {
+      chrome51: true,
       safari10: true,
       safaritp: true,
       webkit:   true,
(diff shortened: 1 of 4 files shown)


chicoxyzzy:
I've also marked Chrome 51 as stable (and 50 as obsolete)

chicoxyzzy:
rebased to resolve conflicts

data-es2016plus.js
```diff
@@ -853,6 +853,8 @@ exports.tests = [
      return Object.getOwnPropertyNames(P) + '' === "a,a,b,b";
      */},
     res: {
+      chrome51: true,
+      chrome52: true,
```
mgol: Setting `chrome51: true` should be enough, shouldn't it?

chicoxyzzy: yep. fixed. thank you

PR #838, opened 2016-07-20
update Safari TP to release 9

chicoxyzzy:
https://developer.apple.com/safari/technology-preview/release-notes/

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -175,6 +175,7 @@ exports.tests = [
           chrome47:        true,
           edge14:          true,
           firefox43:       true,
+          safaritp:        true,
         }
       },
       {
@@ -233,6 +234,7 @@ exports.tests = [
           chrome47:        true,
           edge14:          true,
           firefox43:       true,
+          safaritp:        true,
         }
       },
     ],
@@ -787,6 +789,7 @@ exports.tests = [
          */},
         res: {
           firefox49:   true,
+          safaritp:    true,
         }
       },
       {
@@ -805,6 +808,7 @@ exports.tests = [
          */},
         res: {
           firefox49:   true,
+          safaritp:    true,
         }
       }
     ]
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -175,7 +175,6 @@ exports.tests = [
           chrome47:        true,
           edge14:          true,
           firefox43:       true,
-          safaritp:        true,
         }
       },
       {
--- a/data-es6.js
+++ b/data-es6.js
@@ -13935,7 +13935,7 @@ exports.tests = [
         firefox49:   true,
         xs6:         true,
         jxa:         true,
-        safaritp:    false,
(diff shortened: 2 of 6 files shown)


data-es2016plus.js
```diff
@@ -175,6 +175,7 @@ exports.tests = [
           chrome47:        true,
           edge14:          true,
           firefox43:       true,
+          safaritp:        true,
```
zloirock: Duplicate key.

chicoxyzzy: yes it was mistakenly added. removed this one. thank you

chicoxyzzy: @zloirock noticed you've point to wrong line. Mistakenly added twice was `Array.prototype.includes` feature. This one (`%TypedArray%.prototype.includes`) was added in TP 9 so this one is not duplicate

mkurz:
Why is there a "SF 10" which reaches "100%" and "SF TP" which reaches "only" "99%"? Isn't "SF TP" = "SF 10" - because SF 10 hasn't been released yet?

chicoxyzzy:
@mkurz SF TP is not SF 10. Both are unstable versions

data-es2016plus.js
zloirock: [`jshint` не устроила эта строчка :)](https://travis-ci.org/kangax/compat-table/builds/146200034)

chicoxyzzy:
btw looks like TP got `Symbol.isConcatSpreadable` support so I should update es6 table

mkurz:
@chicoxyzzy So what's the difference between the two? I assume SF10 will end up as the final Safari 10 release. But what about SF TP?

data-es2016plus.js
zloirock: > 236

Hmm... I added this comment to 178 and I checked it. Looks strange.

chicoxyzzy:
@mkurz Safari 10 is shipping with macOS Sierra developer and public betas. Safari TP is available under both macOS Sierra and OS X El Capitan and it is kind of nightlies versions of Safari which are based on WebKit Nightlies (which are available for OS X El Capitan only).
So usually new SF 10 build is based on latest SF TP which is based on some of recent versions of WK Nightly. WK updates more frequently then SF TP which updates more frequently than SF 10

mkurz:
Thank you for the explanation. I was confused because SF 10 scored 100% but SF TP "only" 99% (before this pull request). I would have expected it the other way round - but probably in the TP "nightly" a feature got removed or disabled which threw it back to 99%.

chicoxyzzy:
@mkurz previous TP was released earlier than SF 10. Todays TP 9 got `Symbol.isConcatSpreadable` support. I didn't check es6 tab because I mistakenly thought all SF browsers already got 100% es6 support .Thank you for pointing out on that! :)

mkurz:
@chicoxyzzy This makes totally sense of course :smile:

PR #839, opened 2016-07-20
chrome 52 is stable now

chicoxyzzy:


zloirock:
What about ES6 table?

chicoxyzzy:
You are right! I'm too focused on esnext :) Should be more attentive :P

PR #841, opened 2016-07-21
es5: updates for IE, Edge, Firefox

claudepache:
- <del>Mark IE 9 as obsolete.</del>
- Add Edge. As I can't test Edge 12, I've included Edge 13+ only.
- Morph a footnote about a bug of FF into a subtest for strict mode.
- Add FF 46+, that fixed the aforementioned bug.

### DIFF
--- a/data-es5.js
+++ b/data-es5.js
@@ -28,12 +28,18 @@ exports.browsers = {
     full: 'Internet Explorer 9',
     family: 'Chakra',
     short: 'IE 9',
-    obsolete: false
+    obsolete: true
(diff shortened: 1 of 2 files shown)


claudepache:
Note that FF 21-45 is not yet obsolete because of ESR.

zloirock:
The same like in the `non-standard` table - IE9 - the last IE for Vista and it's still supported.

claudepache:
Corrected. I was too optimistic (or not pessimistic enough).

Issue #844, opened 2016-07-22
Yes/no features are scored as "Large Features" even when weights differ


bterlson:
As a result, block-level function declarations (for example) counts for the same points as spread or the like. The bug is in master.js where yesResults is a count of the number of yesses and totalResults is initialized to this value. Instead, yesResults should be weighted count based on the yes features' size.

bterlson:
I also believe flagged "yes" results are not added to flagged score.

bterlson:
I also believe flagged tally results are incorrect due to the non-flagged tally results getting added in. As in, it assumes the flagged tally is the sum of data-tally and data-flagged-tally but this doesn't appear to be the case.

PR #846, opened 2016-07-23
Fix various bugs in score calculations

bterlson:
This implementation fixes all of the issues in #844 and seems to be slightly faster to boot. Most scores have changed slightly (eg. Edge 92-95%, Chrome 98-97%, FF 49 92-93%, Node 6 93-92%, for example), but I believe this is much more accurate (eg. Node 6's flagged score is no longer 110% :-P). Please take a look!

### DIFF
--- a/master.js
+++ b/master.js
@@ -361,27 +361,49 @@ $(function() {
     else {
       name = currentBrowserSelector;
     }
-    var results = table.find('tr:not([class*=test]):not(.optional-feature) td:not(.not-applicable)' + name);
-    var yesResults = results.filter('.yes').length;
-    var totalResults = results.length;
-    /*
-        Add annex b results
-    */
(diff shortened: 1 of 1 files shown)


master.js
```diff
@@ -361,27 +361,49 @@ $(function() {
...
+    var flaggedPoints = 0;
+    var totalPoints = 0;
+
+    var rows = table.find('tbody tr:not([class=subtest]):not([class=category])');
+    rows.each(function() {
+      var row = $(this);
+      var points = Number(row.attr('significance') || 1);
+      var cell = row.children('td' + name);
+
+      if (row.hasClass('not-applicable')) {
```
ljharb: i think you can make this even faster by adding `:not(.not-applicable)` to the selector on line 369?

bterlson: Actually that's a bug :-P Should be cell.hasClass (not-applicable is applied to cells). Fixing now, good catch!!

Issue #852, opened 2016-07-26
no assignments allowed in for-in head test needs updating


bterlson:
As of https://github.com/tc39/ecma262/pull/614, assignments are only an error in strict mode but works appropriately in sloppy. Probably both should be tested.

chicoxyzzy:
I'll create PR

claudepache:
Fixed by #853...

PR #853, opened 2016-07-26
add test for-in initializers test for sloppy mode

chicoxyzzy:
#852

please review


data-es6.js
```diff
@@ -17103,14 +17103,47 @@ exports.tests = [
...
+        }
+      */},
+      res: {
         ejs:         true,
         babel:       true,
         closure:     true,
         typescript:  true,
-        xs6:         true,
-        jxa:         true,
+        chrome52:    true,
```
chicoxyzzy: this is tested on 52.0.2743.82

chicoxyzzy:
cc @littledan @DanielRosenwasser @teppeis @bterlson

bterlson:
Results look correct for Edge, & thanks for the ping :)

littledan:
LGTM

claudepache:
The test "assignments allowed in for-in head in sloppy mode" should be moved in the "2017 annex b" section of the "2016+" table.

The test "no assignments allowed in for-in head in strict mode" could be adorned with a footnote (look for similar existing footnotes):

> The 2015 version of the specification removed completely the for-in initializer. It was re-added in sloppy mode in the 2017 version.

data-es6.js
```diff
@@ -17103,14 +17103,47 @@ exports.tests = [
         }
       */},
```
claudepache: The test should be:
> for (var i = 0 in {}) {}
> return i === 0;

claudepache: The test should be:

for (var i = 0 in {}) {}
return i === 0;

chicoxyzzy:
@claudepache Thank you! I'll fix it soon

teppeis:
@chicoxyzzy correct for closure-compiler, thanks!

chicoxyzzy:
Moved to ES2017 Annex B.3.6.
FYI it's not merged to spec yet

chicoxyzzy:
squashed into 1 commit

esnext/index.html
```diff
@@ -62,7 +62,7 @@
   <div id="body">
 
     <div class="warning">
-      This table shows proposals which have not yet been included in the current ECMAScript standard, but
```
chicoxyzzy: hmm... this one was built from some previous commit's `esnext/skeleton.html`

esnext/index.html
```diff
@@ -62,7 +62,7 @@
...
+      This table shows features which have not yet been included in the current ECMAScript standard, but
```
chicoxyzzy: hmm... this one was built from some previous commit's `esnext/skeleton.html`

chicoxyzzy:
PR to track https://github.com/tc39/ecma262/pull/614

data-es2016plus.js
```diff
@@ -908,6 +908,31 @@ exports.tests = [
...
+     return i === 0;
+     */},
+    res: {
+      tr:          true,
+      ie10:        true,
+      edge12:      true,
+      firefox35:   true,
+      chrome30:    true,
+      safari51:    true,
+      safari9:     true,
```
claudepache: false in recent version(s) such as safari9, true in oldest versions. More precisely, at one point of time, probably not sooner than March 2014, the side-effect of the initialiser was removed from JSC. See also https://mail.mozilla.org/pipermail/es-discuss/2014-March/036709.html

data-es6.js
```diff
@@ -17107,10 +17108,11 @@ exports.tests = [
         babel:       true,
         closure:     true,
         typescript:  true,
-        xs6:         true,
-        jxa:         true,
+        chrome52:    true,
         safaritp:    true,
```
claudepache, line 17075: Also true in safari9, don't know about safari71_8, most probably false in previous versions.

data-es2016plus.js
```diff
@@ -908,6 +908,31 @@ exports.tests = [
...
+    significance: 'tiny',
+    exec: function(){/*
+     for (var i = 0 in {}) {}
+     return i === 0;
+     */},
+    res: {
+      tr:          true,
+      ie10:        true,
+      edge12:      true,
+      firefox35:   true,
```
claudepache: false in firefox40 and later, true in all previous versions. Same situation as Safari: the side-effect of the initialiser was removed.

chicoxyzzy:
Thank you @claudepache! I think I've fixed all inaccuracies now

chicoxyzzy:
TS won't fix it https://github.com/Microsoft/TypeScript/issues/9978

data-es2016plus.js
```diff
@@ -908,6 +908,32 @@ exports.tests = [
     res: {
     },
   },
+  {
+    name: 'Assignments allowed in for-in head in sloppy mode',
```
webbedspace: Nitpick, but: generally tests are only capitalised when they begin with capitalised identifiers (like `Array` or `System`). Also I think I prefer the unambiguous "non-strict" to the more colloquial "sloppy".

chicoxyzzy: Thank you. Fixed

chicoxyzzy:
It was [merged](https://github.com/tc39/ecma262/pull/614) into spec. I'll update branch soon.

chicoxyzzy:
updated

chicoxyzzy:
rebased again :P

anything else should be done here?

esnext-browsers.js
```diff
@@ -117,6 +117,11 @@ module.exports = {
       short: 'FF 39',
       obsolete: true,
   },
+  firefox40: {
+    full: 'Firefox',
+      short: 'FF 40',
+      obsolete: true,
+  },
```
chicoxyzzy, line 129: I need it here https://github.com/kangax/compat-table/pull/853/files#diff-be5c570f292ab8dcff4e5c9e9a7cff9dR1052

data-es6.js
```diff
@@ -17154,15 +17154,16 @@ exports.tests = [
...
+       'use strict';
+       try {
+         eval('for (var i = 0 in {}) {}');
+       }
+       catch(e) {
+         return true;
+       }
+       */},
       res: {
         ejs:         true,
```
claudepache, line 17069: That test is passed for recent versions of Chrome and Safari. More precisely:

``` js
chrome45: true,
safari9: true,
```

chicoxyzzy:
Updated.
Since Sierra beta 5 (Aug 9) Safari throws for for-in initializers in strict mode.

claudepache:
> Since Sierra beta 5 (Aug 9) Safari throws for for-in initializers in strict mode.

It does already so in current stable version (9). I’m wondering why positive results of Safari ≤ 9 does not automatically carry over into Safari 10?

chicoxyzzy:
@claudepache good catch. This is because their `fullname` property doesn't match. I'll fix it now

chicoxyzzy:
I think it's all correct now

chicoxyzzy:
rebased and bumped to latest WebKit

chicoxyzzy:
It's already hard to rebase on `gh-pages` sometimes

data-es2016plus.js
```diff
@@ -175,7 +175,6 @@ exports.tests = [
           es7shim:         true,
           typescript:      typescript.corejs,
           safari9:         true,
-          safari10:        true,
```
webbedspace, line 178: What's with all these `safari10` deletions?

zloirock, line 178: `safari10` is `true` because of `safari9: true`. However, I don't know why it's a part of this PR instead of WebKit-related PRs.

webbedspace, line 178: Ah, OK, looks fine then.

PR #859, opened 2016-08-03
Update FF48 support

### LINKED ISSUE #856: Firefox 48 is out, and has better compatibility than reported
Firefox 48 is out today, and my tests on Win7 show better compatibility on ECMA6 (91% on mine vs 89% on "FF48") and ECMA2016 (45% on mine vs 39% on "FF48") than reported on the results table.  The current build of FF49 (49.0a2 from 2016-08-01) also has better results on ECMA2016 (47% on mine vs 41% on "FF49") than reported on the results table.
Maybe a rerun of the tests on FF48, FF49 and FF50 (when available) is necessary.

claudepache:
Fixes #856


claudepache:
(Also updated FF49 support for one test.)

Issue #785, opened 2016-04-28
Is raw data available?


opcodewriter:
Is raw data somehow available, by a REST API or just as a JSON file how caniuse.com does it: https://github.com/fyrd/caniuse

sarahkadar:
take a look at `data-[es5|es6|esnext|intl|non-standard].js`.

chicoxyzzy:
I think it's a good idea to have raw data in another repo or in separate json files. I'd like to make a web page that will show progress of features in ECMAScript process. I need some data which compat table already has. For example, I want to show that feature needs to be implemented in 2 browsers to move from stage 3 to stage 4 and in it's already available under flag in one of them. So at least two project will share same data and we can maintain it together.

zoontek:
+1 for putting raw data in another repo.
Babel could actually use this data in a customized preset using browserlist.
Discussion has started here: https://github.com/babel/babel/pull/3476

chicoxyzzy:
Babel announced that they

> ...also will be working on getting a target/env (autoprefixer) preset soon

https://babeljs.io/blog/2016/08/24/6.14.0

So yes. @kangax @webbedspace @zloirock @ljharb what do you think about creating something similar to https://github.com/fyrd/caniuse?

ljharb:
I think the first start would be moving the data to be in a consumable format within this repo - moving it into a separate repo and defining the API can come later.

chicoxyzzy:
previous discussion https://github.com/kangax/compat-table/issues/711

chicoxyzzy:
I like proposal from #711. We can have raw data and test suit separated.

chicoxyzzy:
Duplicate of #711. Let's discuss raw data there

PR #862, opened 2016-08-03
Consolidating Firefox versions with identical results on ES intl

afmenez:
This patch consolidates Firefox versions 29...35 with "36+", which all have identical results.  This makes the resulting HTML both shorter and faster to process, without any loss of information.

### DIFF
--- a/data-esintl.js
+++ b/data-esintl.js
@@ -41,42 +41,7 @@ exports.browsers = {
   },
   firefox29: {
     full: 'Firefox',
-    short: 'FF 29',
-    obsolete: true
-  },
-  firefox30: {
-    full: 'Firefox',
-    short: 'FF 30',
-    obsolete: true
-  },
-  firefox31: {
-    full: 'Firefox',
-    short: 'FF 31',
-    obsolete: true
-  },
-  firefox32: {
-    full: 'Firefox',
-    short: 'FF 32',
-    obsolete: true
-  },
-  firefox33: {
-    full: 'Firefox',
-    short: 'FF 33',
-    obsolete: true
-  },
-  firefox34: {
-    full: 'Firefox',
-    short: 'FF 34',
-    obsolete: true
-  },
-  firefox35: {
-    full: 'Firefox',
-    short: 'FF 35',
-    obsolete: true
-  },
-  firefox36: {
-    full: 'Firefox',
-    short: 'FF 36+'
+    short: 'FF 29+',
   },
   chrome22: {
     full: 'Chrome 22',
--- a/esintl/index.html
+++ b/esintl/index.html
@@ -126,14 +126,7 @@
 <th class="platform edge desktop" data-browser="edge"><a href="#edge" class="browser-name"><abbr title="Microsoft Edge">Edge 12-13</abbr></a></th>
 <th class="platform edge14 desktop" data-browser="edge14"><a href="#edge14" class="browser-name"><abbr title="Microsoft Edge">Edge 14</abbr></a></th>
 <th class="platform firefox2 desktop obsolete" data-browser="firefox2"><a href="#firefox2" class="browser-name"><abbr title="Firefox">FF 2-28</abbr></a></th>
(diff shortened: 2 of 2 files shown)


ljharb:
Personally I prefer to keep all versions explicit and separate, although in this case, since no tests were using those browsers, I think it's fine to consolidate them.

claudepache:
:+1:  for me: because  the tables in their current form are too wide.

ljharb:
The obsolete versions should be hidden by default though, no?

afmenez:
Sure, but they add to the download size and processing time, even when hidden.

zloirock:
LGTM

ljharb:
@afmenez i don't think a K or two, in a gzipped response, will make an appreciable difference :-)

afmenez:
@ljharb, you are probably right, but processing time still persists.
But I am most worried about consistency, as in why versions 29..35 are mentioned individually while versions 16..28 and 36..48 are all merged together?
If you prefer all versions explicit and separate, shouldn't "FF 36+" and "FF 16-28" be replaced by every individual version too? ;)

ljharb:
Indeed, I'd prefer that :-)

afmenez:
That was rhetorical question. :)
You can't be seriously suggesting that we add another 24 identical columns to the table.

mgol:
The performance of various tables is already quite bad if you enable showing obsolete versions. I wouldn't un-merge columns in the data until there is some pre-processing in the repo that checks for identical results amongst subsequent versions of the same browser and merges such columns in the tables. Otherwise it will just become completely unusable after some time.

ljharb:
I think that kind of automatic preprocessing would be great! In other words, I'd want the data to contain an explicit entry for each engine-version combo that was human-verified, and then, when building the HTML, condense that so that any consecutive ranges were auto-collapsed.

zloirock:
cc @kangax @webbedspace

chicoxyzzy:
Nice idea. Also that'll be cool if we can move that explicit data to another repo as I suggested in https://github.com/kangax/compat-table/issues/785#issuecomment-233796019

afmenez:
While this auto-column-merge is not implemented, I still believe this patch manually merging identical columns is a progress.

afmenez:
Rebased to keep up with the progress upstream.

afmenez:
So, this is a simple patch and seems to have enough support.  What is missing to have it merged?

chicoxyzzy:
We want to keep browsers data so I'll close this

afmenez:
@chicoxyzzy, this patch does not remove any data, just merges identical columns.  Please review the patch and look at the comments above.

chicoxyzzy:
@afmenez as mention before in https://github.com/kangax/compat-table/pull/862#issuecomment-237380831 we want to have all versions explicitly.

afmenez:
That comment from @ljharb actually agrees with the patch, as do many others on this thread from @claudepache, @zloirock and @mgol.  What I understand from all the comments is that until we have the automatic column merge implemented, we should do the manual merge to avoid the clutter.

chicoxyzzy:
What we want is to make raw data available. I just didn't have time to move all browsers data to a separate module yet. We need data about all versions with additional data (like date of release) to make it possible to be used by Babel (kinda like autoprefixer) so we actually _need_ all data. If we want to merge data in compat-table we need to make it in stage of compilation to HTML not in source data.

zloirock:
@chicoxyzzy I don't see any problems for this here.

chicoxyzzy:
I do :(

zloirock:
@chicoxyzzy could you specify it?
image

chicoxyzzy:
I've already tried to explain before but seems like my language or explanation skills are not strong enough. Anyway all needed information seems to should be processed by hands in future so I'm not really opposed to merge this for now (it will be a little bit harder to make changes I want to propose though)

afmenez:
I don't see why it would be harder.  If I understand it correctly, to make your proposal we will need a centralized definition of browsers, and this one will be dropped anyway.

PR #863, opened 2016-08-03
Adding Firefox ESR versions and fixing their EOL dates

afmenez:
Adding "ESR" to corresponding Firefox versions and fixing their EOL dates.  Since almost all these versions are already obsolete, this patch is mostly documental.


zloirock:
LGTM

afmenez:
I have now actually installed all the ESR versions and run all tests.  I updated the data tables accordingly.

PR #870, opened 2016-08-08
Adding results for Firefox 4..9

afmenez:
Adding/updating results for Firefox 4 to Firefox 9.


chicoxyzzy:
Just noticed that there is no such browser as `firefox4` in `esnext-browsers.js`. The lower version of FF there is 10

chicoxyzzy:
Potentially this breaks previous valid results. @zloirock we should add FF4+ or revert this commit

chicoxyzzy:
Confirmed. This breaks previous valid results for FF10+

chicoxyzzy:
I've filed PR to fix this #873

afmenez:
My bad, I didn't check all the outputs.  Sorry for that.

Issue #866, opened 2016-08-06
Should Node 5 be marked as obsolete?


chicoxyzzy:
Because it's lifetime ended according to https://github.com/nodejs/LTS

zloirock:
It should be marked as obsolete only at LTS end.

ljharb:
Node 5 never entered LTS, so actually yes, as soon as v6 came out, v5 should have been marked obsolete. The same will apply to node 7 - when node 8 comes out, node 7 is instantly obsolete.

However, node 4 remains NOT obsolete, since it's still LTS, until April 2018, and v6 won't be obsolete until April 2019.

chicoxyzzy:
@ljharb yes I know about Node 4. OK. I'll add PR soon if nobody else will take it

PR #871, opened 2016-08-10
mark Node 5 as obsolete

chicoxyzzy:
consensus reached in #866


chicoxyzzy:
FYI it's safe to merge. It doesn't break any stats

PR #872, opened 2016-08-10
Update Firefox 50 ES2017 annex B support

chicoxyzzy:
Firefox 50.0a2 (2016-08-10)


chicoxyzzy:
https://github.com/kangax/compat-table/pull/870#issuecomment-238840563 should be fixed before merging this PR (and probably this PR should be updated after that)

chicoxyzzy:
Actually this was broken by #870

data-es2016plus.js
```diff
@@ -614,6 +614,7 @@ exports.tests = [
         typescript:  typescript.corejs,
         ie11:        true,
         firefox4:    true,
+        firefox50:   true,
```
claudepache, line 617: "firefox4: true" should have meant "true for firefox4 and later".

However, I guess that, in the ES2016+ table, "firefox4" is not in the list of browsers: so that "firefox4: true" is ineffective.

Before commit 7dae8b07d1bb3d4184f2df67f1b1af09f69c7f32, that line was "firefox10: true", that did work.

There are two solutions:
- either add "firefox4" to the list of browsers for the ES2016+ table;
- or revert partially the incriminated commit by resetting "firefox10: true" in data-es2016plus (2 occurrences).

chicoxyzzy, line 617: fix is already here https://github.com/kangax/compat-table/pull/873

PR #874, opened 2016-08-12
[ES2016plus] Make the "IE 10 or older" definition clearer

mgol:


zloirock:
Why do you think it's a typo?

mgol:
Oh, does it mean "IE 10 or older?" With other columns sometimes using ranges it sounded to me as an unfinished "IE10 to something newer" range...

zloirock:
Yep. Feel free propose something better readable.

mgol:
I don't have better ideas, the `n-` shortcut is popular. It might be that I've seen specifically `IE10-11` before and just connected those two, maybe I'd understand it immediately otherwise. I'll close this PR then, sorry for the fuss.

ljharb:
How about `IE <= 10` Or `IE < 11`?

mgol:
That would work. :) With spaces?


## 

Michał Gołębiowski

ljharb:
Probably with spaces, for readability?

mgol:
PR updated with the changes proposed by @ljharb.

ljharb:
LGTM

PR #877, opened 2016-08-15
Adding results for Firefox 2..3.6

afmenez:
Adding/updating results for Firefox 2 to Firefox 3.6.


data-es2016plus.js
```diff
@@ -734,8 +734,10 @@ exports.tests = [
          && Object.prototype.__lookupGetter__.call(obj, "qux") === undefined
          && Object.prototype.__lookupGetter__.call(obj, "baz") === undefined;
          */},
-        res: temp.basicDefineGetterResults,
-      },
+        res: Object.assign({}, temp.basicDefineGetterResults, {
+          firefox2:   true,
```
zloirock, line 740: Funny. FF2-3.x missed in `basicDefineGetterResults` because hasn't support `Object.getOwnPropertyDescriptor`. But it's ES8, so I don't think that a fallback for ES3 engines makes sense.

data-es2016plus.js
```diff
@@ -734,8 +734,10 @@ exports.tests = [
...
+        }),
+	  },
```
zloirock: Indent.

claudepache:
@afmenez It is very dubious that FF2 supports some of the "strict mode" subfeatures, since strict mode wasn’t invented at that time.

Looking closer to those subtests passed by FF2, it appears that they are in fact flawed. (Except the last one, which tests against too much strictness.)

claudepache:
> Looking closer to those subtests passed by FF2, it appears that they are in fact flawed.

See #878

afmenez:
Nice, I'll rerun the tests after #878 gets merged, then I'll update this patch.

afmenez:
Other suspicious issues with the ES6 test results:
- "destructuring, assignment" results are unusually high for Firefox 3;
- "block-level function declaration[14]" passing for Firefox 2 & 3;
- "misc/accessors aren't constructors" passing for Firefox 2 & 3;

There may be more problems on these tests, or maybe we need to add notes on the results.

data-es2016plus.js
webbedspace, line 740: Object.getOwnPropertyDescriptor is ES5… Object.getOwnPropertyDescriptor**_s**_ is ES2017. As for basicDefineGetterResults… since it's kinda not possible to verify that the getters installed by FF2 have all of the features observable with Object.getOwnPropertyDescriptor, I'm not sure how this should go.

claudepache:
> "destructuring, assignment" results are unusually high for Firefox 3;

I think that destructuring was pionneered by one or more browsers before it was considered for standardisation, so that old browsers might support it partly.

> "block-level function declaration[14]" passing for Firefox 2 & 3;

This is because those browsers don’t support strict mode, so that they apply some browser-defined sloppy-mode semantics which happens to pass the particular test case. If we care, we could either
- first check that strict mode is supported – although there _might_ be some browser that supports strict mode partially, and could produce a false positive –; or
- find a testcase that happens to fail for each of the various incompatible sloppy-mode semantics that old browsers used to have.

> "misc/accessors aren't constructors" passing for Firefox 2 & 3;

Indeed, that test is incorrectly passed because those browsers do not support Object.getOwnPropertyDescriptor and/or getters in object literals.

afmenez:
What I found strange is not that FF2 & FF3 pass these tests, but that they do and FF4 doesn't.

webbedspace:
The block-level function semantics were [deliberately changed in FF4](http://whereswalden.com/2011/01/24/new-es5-strict-mode-requirement-function-statements-not-at-top-level-of-a-program-or-function-are-prohibited/).

claudepache:
For block-level function declaration, see #879. By analysing the results of Walden's post mentioned by @webbedspace , it seems to me that pre-ES5 browsers should now fail the improved test.

afmenez:
After #878, most strict mode tests now fail for Firefox 2 & 3, except for "function expressions with matching name and argument are valid", which is still passing.

claudepache:
> except for "function expressions with matching name and argument are valid", which is still passing

@afmenez Yes, that test is for checking that implementations are not too strict (a former bug of WebKit); so it's expected that old browsers pass it.

afmenez:
After #879, Firefox 2 & 3 fail the test, as expected.

afmenez:
I have run all the tests again, and almost all the suspicious results are fixed.  The only one I am still not sure is the "destructuring, assignment" group in ES6, where results are higher for Firefox 3 than for Firefox 4 (14/24 vs 9/24).  If the tests are fine, maybe a note explaining the results is in order, but I am not sure what to write.

claudepache:
Confirmed that there seems to have been a strange regression between FF3 and future versions w.r.t. destructuring.

> If the tests are fine, maybe a note explaining the results are in order, but I am not sure what to write.

I’d say, no note, or maybe just some `/* sic */` comments in the source code. Anyway, nobody cares about the ES6 support of FF2-3; so you could just not include those browsers.

afmenez:
OK, so I believe the patch is ready to merge.

ljharb:
It's likely not a regression - destructuring wasn't a thing when FF3 came out, so they probably decided to change their support for a nonstandard feature.

claudepache:
@ljharb I see no rational reason but bug for the following behaviour change:

``` js
(function() { var { a: a } = { a: 1 }; return a; })()         // 1 in FF3.0, 1 in FF3.6
(function() { var { a }    = { a: 1 }; return a; })()         // 1 in FF3.0, 1 in FF3.6
(function() { var a;  ( { a: a } = { a: 1 } ); return a; })() // 1 in FF3.0, 1 in FF3.6
(function() { var a;  ( { a }    = { a: 1 } ); return a; })() // 1 in FF3.0, undefined in FF3.6
```

afmenez:
It was not very easy to run the tests on Firefox 2 & 3, they are slow and I had to click through A LOT of warnings on each run.  But I believe it was worth it since 3 issues were exposed and fixed.  I now wonder what is missing to have this patch merged now.

PR #881, opened 2016-08-16
Chrome 54 update

chicoxyzzy:
Chrome 54 supports Object.values, Object.entries and Object.getOwnPropertyDescriptors without flags now


esnext-browsers.js
```diff
@@ -284,6 +284,18 @@ module.exports = {
       short: 'CH 52',
       note_id: 'experimental-flag',
   },
+  chrome53: {
```
afmenez, line 287: Is Chrome 53 really needed here?  There's no other mention to this version in the test results, I don't see the point of the extra column.

chicoxyzzy, line 287: I don't see any benefits of having Chrome 52-53 as well. More than that someone may need to mention `chrome53` in table in any time later. Also it will be stable soon.

afmenez, line 287: It could be added then, if necessary, avoiding the cost (size and speed) of adding it earlier "just in case".

chicoxyzzy, line 287: I don't like idea of squashing 52 and 53 into one

ljharb, line 287: Agreed.

We should not be optimizing for column counts in the data itself. If we want columns to be minimized, let's write build/display logic to handle that - the data should always be as granular as possible.

afmenez, line 287: @chicoxyzzy, this is not what I have suggested, but simply removing the 53 column while there's no actual need for it.  As I understand, you then suggested also removing 52, and I agree.

afmenez, line 287: @ljharb, I have to disagree.  There's no data for Chrome 52 or Chrome 53.  You do not create data granularity by creating empty columns of data.
It seems reasonable to remove columns that have identical data, which is not always obvious due to the way the data is defined.  But I see no point in knowingly adding empty columns and then process them out because they are empty.

ljharb: Sorry if I missed the context :-) I agree that empty columns don't serve a purpose, but I'd rather the build process remove empty columns rather than make it harder for someone to add chrome 53-specific data later.

esnext-browsers.js
```diff
@@ -284,6 +284,18 @@ module.exports = {
...
+    full: 'Chrome',
+      short: 'CH 53',
+      note_id: 'experimental-flag',
+      unstable: true,
+  },
+  chrome54: {
+    full: 'Chrome',
+      short: 'CH 54',
```
zloirock: Indent.

chicoxyzzy: @zloirock I made it the same as elsewhere in this file

zloirock: Looks like I didn't check this file after extraction browsers. Need to fix it, maybe in separate PR.

chicoxyzzy: I can fix identation here if you want but I think it's better to create new PR. I can create it after this one will get merged

chicoxyzzy:
BTW estimated release date for Chrome 53 is September 6 and we still don't have it in table

esnext-browsers.js
chicoxyzzy: fixed this here

PR #885, opened 2016-08-17
Update Safari TP to release 12

chicoxyzzy:

PR #886, opened 2016-08-17
Update es2016plus version according latest webkit r204559

gskachkov:
- Object.values/Object.entries
- nested rest destructuring, declarations
- nested rest destructuring, parameters
- class extends null

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -128,6 +128,7 @@ exports.tests = [
       firefox47:  true,
       chrome51:   flag,
       edge14:     true,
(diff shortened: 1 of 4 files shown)


ljharb:
Are these from webkit nightly?

chicoxyzzy:
Actually I've made PR with these changes for WK Nightly and Safari TP 11 one hour before https://github.com/kangax/compat-table/pull/885 :)

gskachkov:
Yep, but I found out that PR those changed already there.

PR #888, opened 2016-08-19
add proposed Math extensions

dnalborczyk:
stage 1 proposal for Math extensions.

tests are based on ES6 "Number properties" and "Math methods", significance: small.

https://github.com/rwaldron/proposal-math-extensions

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1686,6 +1686,56 @@ exports.tests = [
     }
   ]
 },
+{
+  name: 'Math extensions',
+  category: 'proposal (stage 1)',
+  significance: 'small',
+  link: 'https://github.com/rwaldron/proposal-math-extensions',
+  subtests: [
+    {
+      name: 'Math.clamp',
+      exec: function() {/*
+        return typeof Math.clamp === "function";
+      */},
+      res: {},
+    },
+    {
+      name: 'Math.scale',
+      exec: function() {/*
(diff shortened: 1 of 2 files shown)


data-esnext.js
```diff
@@ -1686,6 +1686,56 @@ exports.tests = [
...
+      name: 'Math.degrees',
+      exec: function() {/*
+        return typeof Math.degrees === "function";
+      */},
+      res: {},
+    },
+    {
+      name: 'Math.RAD_PER_DEG',
+      exec: function() {/*
+        return typeof Math.RAD_PER_DEG === "number";
```
ljharb, line 1726: these should check the values as well, not just the types.

data-esnext.js
```diff
@@ -1686,6 +1686,56 @@ exports.tests = [
...
+      name: 'Math.radians',
+      exec: function() {/*
+        return typeof Math.radians === "function";
+      */},
+      res: {},
+    },
+    {
+      name: 'Math.degrees',
+      exec: function() {/*
+        return typeof Math.degrees === "function";
```
ljharb, line 1719: we should also invoke each of these for some edge cases and example values, and not just test presence.

zloirock:
https://github.com/kangax/compat-table/commit/e87b4db995ca3e4ed134671908e32c36799f3f68

PR #889, opened 2016-08-21
add tests for 'template literal revision'

dnalborczyk:
'template literal revision' (stage 3)

tests are based on ES6 template literal and the draft, significance: small.

https://github.com/tc39/proposal-template-literal-revision

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1686,6 +1686,26 @@ exports.tests = [
     }
   ]
 },
+{
+  name: 'template literal revision',
+  category: 'candidate (stage 3)',
+  significance: 'small',
+  link: 'https://github.com/tc39/proposal-template-literal-revision',
+  exec: function() {/*
+    function fn(strings, a, b) {
+      return typeof strings[0] === "undefined" &&
+        typeof strings[1] === "undefined" &&
+        typeof strings[2] === "undefined" &&
+        strings.raw[0] === "\\unicode " &&
+        strings.raw[1] === " \\xerxes " &&
+        strings.raw[2] === " \\u{55}" &&
+        a === 123 &&
+        b === 456;
+    }
+    return fn`\unicode ${123} \xerxes ${456} \u{55}`;
+  */},
+  res: {},
+}
 ];
 
 //Shift annex B features to the bottom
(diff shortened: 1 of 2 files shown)


afmenez:
Besides the merge conflict (which seems trivial to fix), is there another issue holding this back?

ljharb:
The original fork has been deleted, so it's not trivial to fix this PR, sadly. @dnalborczyk, do you have any interest in completing this PR?

afmenez:
I have applied the patch to a local branch, and after updating some fields in the header it seems to be working.  The problem is that the test is failing on Firefox 53 & 54, which claim to have this implemented.

These are the two lines that fail on Firefox:
```
        typeof strings[1] === "undefined" &&
        typeof strings[2] === "undefined" &&
```
If I remove these two lines, the test passes on Firefox 53 and fails on Firefox 51, as expected.  I don't know enough to check if the problem is in the test or in Firefox implementation.

afmenez:
There's a new commit in Firefox 55 Nightly ([bugzilla #1108941](https://bugzilla.mozilla.org/show_bug.cgi?id=1108941)), but the problem stated above persists.

ljharb:
@disnet, could you weigh in on this particular test?

data-esnext.js
```diff
@@ -1686,6 +1686,26 @@ exports.tests = [
...
+{
+  name: 'template literal revision',
+  category: 'candidate (stage 3)',
+  significance: 'small',
+  link: 'https://github.com/tc39/proposal-template-literal-revision',
+  exec: function() {/*
+    function fn(strings, a, b) {
+      return typeof strings[0] === "undefined" &&
+        typeof strings[1] === "undefined" &&
+        typeof strings[2] === "undefined" &&
```
disnet, line 1698: `\u{55}` is a legal escape sequence, both before and after this proposal so you shouldn't be asserting it is undefined

afmenez:
Great.  Should I remove the 2 tests and submit a new PR?

afmenez:
I created PR #1046 with an updated patch.

data-esnext.js
afmenez: Do you mean none of the parameters should be checked for undefined, just [0] (which make the test pass on Firefox), or just [0] and [1] (leaving out just the `\u{55} `)?

Issue #861, opened 2016-08-03
What is the policy for retiring a browser to "obsolete : very"?


afmenez:
There seems to be an excessive number of Firefox and Chrome versions active, which make the download bigger and the page slower.  What is the policy for retiring a browser with "obsolete : very"?

claudepache:
The usage of the only contributor caring about seems to be: "all but the last 10", see: https://github.com/kangax/compat-table/commit/4802cb08171ae4e50a660aa58c35bd10906f72e6, https://github.com/kangax/compat-table/commit/43f10efd69bff5405e1dec17930a24799ca5333f. That represents 13-14 months old, which, in my opinion, is not too much.

afmenez:
We currently have 14 Chrome and 10 Firefox obsolete versions (I removed 2 last week).  I believe this is too much.  Both browsers follow the rapid release with automatic updates policy, so older versions are not that important (as they are for IE for example, where we have only one obsolete version listed).

I suggest we keep the following:
- For Firefox keep 6 obsolete version, and at least one should be an ESR;
- For Chrome keep 6 obsolete versions;

This would remove 12 columns from the tables, making the much lighter.  Maybe even usable on mobile browsers.

afmenez:
Or we can go with browser usage.  If we retire browsers with less than 0.1% usage (according to http://caniuse.com), we get:
- Firefox ESR 38 + Firefox >=43;
- Chrome >= 42;
- IE >= 8;
- Safari >= 8;

Seems reasonable?

claudepache:
Not commenting about the idea of using browser usage, but about how you compute the threshold Firefox 43 / Chrome 42.

You say: Firefox N, for each N between 2 and 42 (excluding 38):  less than 0.1%, not significant.

I say: Firefox 2-42 (excluding 38):  0.02% + 0.04% + ... + 0.07% = 1.11%, significant.

Or, in other words, I am more interested into the _total_ number of users I leave behind.

chicoxyzzy:
Right now we have inconsistency between different tabs 😞

ljharb:
@chicoxyzzy let's make all of the tabs consistent ASAP - any browser that's not obsolete, or not "very" obsolete, in _any_ tab, should be not so in all of them.

chicoxyzzy:
@ljharb than we will end up rendering all browsers starting from IE 7, FF 2, Chrome 5, SF 3.2, etc. because they all will become at least obsolete.

[...]

PR #891, opened 2016-08-24
Updating results for Firefox 18 to Firefox 23

afmenez:
Updating results for Firefox 18 to Firefox 23.


data-non-standard.js
```diff
@@ -1218,7 +1212,7 @@ exports.tests = [
     firefox12: false,
     firefox16: true,
     firefox17: false,
-    firefox20: true,
+    firefox20: false,
     firefox28: false,
```
chicoxyzzy, line 1257: this is useless because your PR removes Firefox 28. Don't sure if it's good idea though

afmenez: I am not sure what you mean.  This is fixing the result for Firefox 20..27, which was wrong.
No browsers were removed, but I did consolidate identical columns of the table.

chicoxyzzy: Firefox 28 was removed by you from browser list for `non-standard` tab. Look at diff starting at line 79. There is only `firefox20` (which includes version from 20 to 29) and `firefox28` anymore in your branch

chicoxyzzy: so this particular line (`firefox28: false,`) does nothing

chicoxyzzy: in fact there's a lot of `firefox28` mentions in this file

afmenez, line 1257: That's what I said, I removed the column but not the data.  Do you mean you want the columns back?  I am starting to feel optimizing identical columns is a lost battle.

chicoxyzzy: What I'm trying to say is that in your branch

```
firefox20: false,
firefox28: false,
```

is the same as

```
firefox20: false,
```

because there is no data under key `firefox28` in browsers list, it is represented under `firefox20` in your branch. So `firefox28: false,` is useless and should be deleted as well as other `firefox28` mentions in this file.

By the way I  hope this PR will not be merged. Those browsers are  **very** obsolete and they don't get rendered in table anyway. The data about particular browsers may still be useful, for example for cases described in https://github.com/kangax/compat-table/issues/785

afmenez: I see now what you mean.  I tried to remove redundant data in a previous patch, and had to revert it because I was throwing data away.  Apparently having these 2 redundant lines mean that both browsers were actually tested.  There's a lot of this in this file.  I kept both because I didn't wat to have my patch refused again because of the optimization.

These browsers ARE rendered in this table if you check the "Obsolete Browsers" checkbox.  For the other tables, the data is there and is wrong, so I believe it is worth fixing even if they are not displayed by the default renderer.

chicoxyzzy: > Apparently having these 2 redundant lines mean that both browsers were actually tested.
> 
> I kept both because I didn't wat to have my patch refused again because of the optimization.

All browsers which have identical `full` field will nest data from previous marked version's value for each test and it's normal in this repo. Actually you loose data when you delete browser from browser list and not loose any useful data when you delete repeating values in tables (and especially you loose nothing when you delete data for browser not represented in browser list).

> These browsers ARE rendered in this table if you check the "Obsolete Browsers" checkbox. For the other tables, the data is there and is wrong, so I believe it is worth fixing even if they are not displayed by the default renderer.

There are no Firefox browsers with versions lower then 31 get rendered if you check obsolete browsers checkbox on. This is because of `obsolete: very,` fields in very old versions. There is nothing about very obsolete browsers in DOM.
So this is not really an optimisation.

afmenez: > All browsers which have identical full field will nest data from previous marked version's value for each test and it's normal in this repo. Actually you loose data when you delete browser from browser list and not loose any useful data when you delete repeating values in tables (and especially you loose nothing when you delete data for browser not represented in browser list).

I agree with you, and I said it above.  I'll do a separate patch removing redundant entries like this, please help me merge it then.

> There are no Firefox browsers with versions lower then 31 get rendered if you check obsolete browsers checkbox on. This is because of obsolete: very, fields in very old versions. There is nothing about very obsolete browsers in DOM.  So this is not really an optimisation.

I understand how this works, and I have marked some versions as "very obselete" myself.  I even have an open issue suggesting we mark more versions as "very obsolete".  But for non-standard features table, which is the one we are talking about here (data-non-standard.js), no browsers are marked as "very obsolete".

chicoxyzzy: > But for non-standard features table, which is the one we are talking about here (data-non-standard.js), no browsers are marked as "very obsolete".

Wow. Didn't notice. They should match obsolete browsers for other tables IMHO. Did someone had any counterarguments?

afmenez, line 1257: > Wow. Didn't notice. They should match obsolete browsers for other tables IMHO. Did someone had any counterarguments?

What we should have is a single browser definition place, like in
"esnext-browsers.js", and remove all other form inside the data files.  I
an willing to do that in a separate patch, but only after this pull queue
is emptied to avoid all the merge issues.

afmenez:
So, this is another straightforward patch and all opposition was addressed. What is missing to have it merged?

chicoxyzzy:
This needs rebase. It LGTM but I'd prefer someone else to review it (I think @ljharb may disagree with merging FF versions in `non-standard` tab for example)

afmenez:
I don't have access to my main machine now, I'll rebase it tomorrow.  If the column merge is the only thing holding this PR back, I'll revert it when I do the update.
I'll also do another PR regarding FF49, which was released today.

data-non-standard.js
```diff
@@ -87,14 +87,8 @@ exports.browsers = {
...
   firefox20: {
-    full: 'Firefox 20-27',
-    short: 'FF 20-27',
-    family: 'SpiderMonkey',
-    obsolete: true
-  },
-  firefox28: {
-    full: 'Firefox 28-29',
-    short: 'FF 28-29',
+    full: 'Firefox 20-29',
```
ljharb, line 90: It would be nice to keep all of the explicit `firefox28` results, but i don't think it needs to block the PR.

chicoxyzzy:
I'd prefer to merge this before merging all browsers descriptions into one module (as per discussed in https://github.com/kangax/compat-table/issues/861)

data-non-standard.js
afmenez: All the data is intact, this only merges the visualization.

afmenez:
Rebased and ready to merge.

chicoxyzzy:
Thank you

PR #892, opened 2016-08-24
Update WebKit to r204915

gskachkov:
- Implemented "Assignments allowed in for-in head in non-strict mode"

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -472,7 +472,7 @@ exports.browsers = {
     unstable: true
   },
   webkit: {
-    full: 'WebKit r204466 (August 15, 2016)',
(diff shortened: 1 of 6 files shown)


data-es6.js
```diff
@@ -472,7 +472,7 @@ exports.browsers = {
     unstable: true
   },
   webkit: {
-    full: 'WebKit r204466 (August 15, 2016)',
+    full: 'WebKit r204915 (August 24, 2016)',
```
zloirock, line 475: Also in `esnext-browsers.js`.

gskachkov: Done

Issue #830, opened 2016-07-07
test262 conformance as an line item the compat table


littledan:
Test262 is another test suite for JavaScript. It provides a number of very detailed set of around 20k tests and aims to test every line in the specification, going into some details which the current compat table omits. However, there is not currently a place where test262 scores are surfaced, and test262 does not include categorization into features, weighting, or a pretty website.

For me as an implementer, test262 has been extremely useful in identifying non-compliant parts of V8 so that we can fix them. Although many of the bugs don't reflect something that seems likely for a user to run into, it feels to me like a good tool for making a JavaScript implementation into a reliable, consistent surface.

Maybe it would be useful for others as well to see in a comparison context. I'm wondering whether it might make sense to have one item in the compat table here be for the entire test262 score. It could be factored into the overall score, or left to the side for independent reference. Thoughts?

ljharb:
I love this idea. This table is mostly checking presence and mostly not checking compliance, and compliance matters far more money than presence.

kangax:
This is a great idea! The fact that test262 compliance is not listed anywhere on a web for any browser seems like an unfortunate omission. We've pushed the table far enough, to the point where some engines score 100%. It's only fair to expose even more detailed metrics. I imagine a separate table would be most convenient for implementors? Something with breakdown by test group, to see what exactly is failing?

ljharb:
If there is a way to _automatically_ convert commandline test262 results to a web-consumable format, that would be ideal imo, because then we could programmatically show a breakdown with collapsible details etc.

natorion:
There is a web test runner for test262. It produces html and json as a result. You can have a look at it at http://v8.github.io/test262/website/default.html. The runner unfortunately is quite buggy (did not work in Safari the last time I checked).

littledan:
Yes, there's significant technical work needed to make this work, which @jugglinmike, @natorion and I have looked into to some extent. If anyone is interested in implementation along these lines, I could help you get started; I won't have the time to do this technical work myself at the moment.

bterlson:
I support this idea, although I still hope someone will contribute a better website to ECMA like the old test262.ecmascript.org website. I think the hardest problem with this is going to be the UI - there is a lot of data in test262 and presenting it in a useful way is hard.

FWIW, I have just released a new version of https://github.com/bterlson/test262-harness which supports running test262 in browsers from the command line.

chicoxyzzy:
This also might be useful
[...]

PR #896, opened 2016-08-30
Handle raw data generation

zoontek:
As Babel community starts talking about an autoprefixer-like preset, we need to extract the raw data in other format that HTML.

It's a first and basic step, maybe we should split in several repos someday (?)
- compat-table/generation (containing JavaScript tests)
- compat-table/raw-data (JSON generation using compat-table/generation)
- compat-table/website (HTML generation using compat-table/generation)

It's just an idea of course.

Refs:
https://github.com/kangax/compat-table/issues/785
https://github.com/babel/babel/pull/3476


build.js
```diff
@@ -356,24 +356,42 @@ function handle(options) {
...
+  var html_result = replaceAndIndent(html, [
     ["<!-- NAME -->", [options.name]],
     ["<!-- URL -->", [options.name.link(options.url)]],
     ["<!-- POLYFILLS -->", !options.polyfills ? [] : options.polyfills.map(function(e) {
       return '<script>' + fs.readFileSync(__dirname + path.sep + e, 'utf-8').replace(/<(?=\/script>)/g,'\\u003c') + '</script>\n';
     })],
   ]).replace(/\t/g, '  ');
+  var json_result = JSON.stringify(options.tests, null, 2);
+
+  var html_target_file = __dirname + path.sep + options.html_target_file;
```
ljharb: Let's use path.join here, instead of having to manually use path.sep

ljharb:
I like this approach, but our data has functions in it, and JSON does not support functions.

zoontek:
@ljharb Are data functions the ones used to generate the "Current browser" column?

chicoxyzzy:
@ljharb why should browser data include functions? Most likely we can replace them by strings or numbers.

ljharb:
@chicoxyzzy for the raw list of browsers, JSON is totally fine. I'm talking about the actual tests, which require the code to use for the test, which are functions. Converting them to strings isn't the right direction imo, as already they rely too heavily on `eval`.

chicoxyzzy:
@zoontek indeed. your PR is missing browser tests. Of corse this repo can be splitted some day but I think we should move step by step with fully functioning intermediate changes.

chicoxyzzy:
Another thoughts for future steps:
- browser lists should be merged for all browsers
- browser lists shouldn't include browser ranges
- browsers with higher versions shouldn't nest data from previous versions implicitly

chicoxyzzy:
Revisiting this PR.

I think it's good but we should add fields which link test to actual test body (or even better to tests in https://github.com/tc39/test262 in the future)

chicoxyzzy:
ref: https://github.com/kangax/compat-table/issues/830

chicoxyzzy:
This PR is outdated. I'm going to solve this problem in another way. Thank you for working on this!

PR #907, opened 2016-09-16
Updating results for Firefox 32 to Firefox 38

afmenez:
Updating results for Firefox 32 to Firefox 38.


data-es6.js
```diff
@@ -6722,7 +6722,7 @@ exports.tests = [
       */},
       res: Object.assign({}, temp.basicSet, {
           firefox13: false,
-          firefox32: true,
```
ljharb, line 6725: does this mean that firefox 32-36 does _not_ support this feature?

afmenez: Yep.

PR #908, opened 2016-09-19
Add FF25 to browsers list

### LINKED ISSUE #906: Feature displaying incorrectly on es2016+ table
On the es2016+ table, the feature named 2016 "misc/Proxy, "enumerate" handler removed" has the following data:
    res: {
      firefox24: true,
      firefox25: false,
      firefox47: true,
      chrome50: true,
      safari10: true,
      safaritp: true,
      webkit: true,
    },

I believe this means that the feature is not working on Firefox 25 to Firefox 46, but it is showing as working for FF31 to FF46 on the generated HTML table.  Can this a bug on the code that builds the table?

chicoxyzzy:
Fixes #906

### DIFF
--- a/esnext-browsers.js
+++ b/esnext-browsers.js
@@ -107,6 +107,11 @@ module.exports = {
     short: 'FF 24<br> ESR',
     obsolete: very // ESR (EOL on October 14, 2014)
   },
+  firefox25: {
+    full: 'Firefox',
+    short: 'FF 25',
+    obsolete: very,
+  },
   firefox27: {
     full: 'Firefox',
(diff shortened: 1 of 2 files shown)


esnext-browsers.js
```diff
@@ -222,7 +227,7 @@ module.exports = {
   chrome34: {
     full: 'Chrome',
     short: 'CH 34',
-    obsolete: very,
+    obssolete: very,
```
afmenez: I believe this was a mistake.

chicoxyzzy: yes. thank you!

PR #909, opened 2016-09-20
corrections for edge14 in es6 and es2016+

claudepache:
According to testing on v14.14393
- es6: doesn't support iterator closing;
- es6: doesn't support Symbol.hasInstance, Symbol.isConcatSpreadable;
- es2016: supports catching generator throw() by inner generator;
- es2017: support of async functions still behind flag;
- es2017: supports Proxy internal calls for **define{G,S}etter**;


afmenez:
LGTM.

chicoxyzzy:
Thank you!

Issue #910, opened 2016-09-20
WebKit Nightly inconsistency in different tables


chicoxyzzy:
WebKit is marked as stable (with comment that it's always up-to-date) in `Intl` table and as unstable in all other tables. We should make things consistent.

In my opinion we should mark it as unstable because Nightly builds are supposed to be unstable. I'm going to create PR after Safari 10 will be released along with Sierra today and mark WK as unstable there. Please leave a comment here if you have an opposite opinion.

PR #911, opened 2016-09-20
mark Safari 10 as stable, update Safari TP to 13

chicoxyzzy:
this PR also includes fix for https://github.com/kangax/compat-table/issues/910

### DIFF
--- a/data-es5.js
+++ b/data-es5.js
@@ -121,9 +121,15 @@ exports.browsers = {
     obsolete: true
   },
   safari6: {
-    full: 'Safari 6.0, Safari 7.0, Safari 7.1, Safari 8, Safari 9',
+    full: 'Safari 6.0, Safari 7.0, Safari 7.1, Safari 8',
     family: 'JavaScriptCore',
-    short: 'SF 6-9',
+    short: 'SF 6-8',
+    obsolete: true
+  },
+  safari9: {
+    full: 'Safari 9',
+    family: 'JavaScriptCore',
+    short: 'SF 9',
     obsolete: false
   },
   safari10: {
@@ -133,7 +139,7 @@ exports.browsers = {
     obsolete: false
   },
   safaritp: {
-    full: 'Safari Technology Preview Release 12',
+    full: 'Safari Technology Preview Release 13',
     family: 'JavaScriptCore',
     short: 'SF TP',
     unstable: true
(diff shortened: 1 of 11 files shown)


kangax:
@chicoxyzzy hey, just added you to the repo! thanks for all the help :)

data-es6.js
```diff
@@ -508,19 +508,20 @@ exports.browsers = {
   safari71_8: {
     full: 'Safari',
     short: 'SF 7.1,<br>SF 8',
-    obsolete: false
+    obsolete: true
   },
   safari9: {
     full: 'Safari',
     short: 'SF 9',
+    obsolete: false,
```
afmenez, line 516: If Safari 10 was released, shouldn't Safari 9 be obsolete now?

data-es5.js
```diff
@@ -121,9 +121,15 @@ exports.browsers = {
...
     family: 'JavaScriptCore',
-    short: 'SF 6-9',
+    short: 'SF 6-8',
+    obsolete: true
+  },
+  safari9: {
+    full: 'Safari 9',
+    family: 'JavaScriptCore',
+    short: 'SF 9',
     obsolete: false
```
afmenez: If Safari 10 was released, shouldn't Safari 9 be obsolete now?

data-esintl.js
```diff
@@ -113,18 +113,28 @@ exports.browsers = {
...
     full: 'Safari',
-    short: 'SF 7.1,<br>SF 8',
-    obsolete: false
+    short: 'SF 7.1-8',
+    obsolete: true,
+  },
+  safari9: {
+    full: 'Safari',
+    short: 'SF 9',
+    obsolete: false,
```
afmenez: If Safari 10 was released, shouldn't Safari 9 be obsolete now?

mkurz:
iOS10 is stable now as well

data-es6.js
chicoxyzzy: We mark 2 latest versions of Safari as stable. This line was added just to make it easier to change it to `true` later

chicoxyzzy:
@mkurz that's true. We should update all iOS in all tabs. Mobile Safari data is outdated everywhere and it's a good idea for another PR. It should be very easy. All iOS data should correspond current Safari data. Do you want to take a look? :)

PR #913, opened 2016-09-21
[ES6] Add Node.js 6.5 data, fixes the Node 0.12 EOL date

### LINKED ISSUE #901: Update Node 6 results.
According to http://node.green/ Node.js 6.5.0 has the following percentages:
- ES2015 (ES6): 99% (99% w/ flags)
- ES2016: 76% (100% w/ flags)
- ES2017: 33% (41% w/ flags)

The following are supported through a flag:
- ES6: tail call optimization
- 2016: exponentiation operator
- 2017: Object.values
- 2017: Object.entries
- 2017: Object.getOwnPropertyDescriptors (basic support)

It would also be nice if the tooltip for Node 6 indicated which version because there's a difference in the level of support between 6.4.0 and 6.5.0.

mgol:
Fixes #901

Two remarks:
1. I'm not sure if `Node 6.5+` is the best moniker as it suggests it covers all future Node versions as well. Ideally we'd have a text that covers `>=6.5 <7` in semver terms.
2. I run the test via `node .` as well and it failed some `Promise.all`/`Promise.race` tests. This is surprising considering that Node 6.5 is based on the same V8 as Chrome 51 which is passing those tests. Also, I run one of the failing tests in Node 6.5 manually and it passed. Are those async tests handled correctly in Node?


chicoxyzzy:
1. Let's just name it `Node 6.5` for now. We are discussing moving all browsers and other environments to a single module (https://github.com/kangax/compat-table/issues/861#issuecomment-248463955) so we'll probably end up separating all versions anyway. Also missing version is better then wrong results IMO.
2. Hmm... This sounds really strange. We should check Node testing. But the changes in PR looks good to me

mgol:
PR updated.

chicoxyzzy:
Thank you!

PR #914, opened 2016-09-21
PTC is available in Chrome 55 under flag only

chicoxyzzy:


mgol:
In light of this, couldn't Chrome 53+ be collapsed again? I don't expect the ES6 table to change in the foreseeable future in the Chrome section and this repo has been usually collapsing such sections to make it for easier reading.

chicoxyzzy:
@mgol Chrome 55 doesn't pass `Symbol.toStringTag affects existing built-ins` test so the results are not the same for Chrome 53-54 and Chrome 55. Also we should collapse columns programmatically and not by merging browser version. But this will be done after we merge all browsers descriptions into one module

Issue #828, opened 2016-07-03
Use new ECMAScript spec website?


impressivewebs:
Do you think it would be beneficial to use the new ECMA spec for the links in your compat tables?

https://tc39.github.io/ecma262/

For example, instead of:

http://www.ecma-international.org/ecma-262/6.0/#sec-generator-function-definitions

You can link to:

https://tc39.github.io/ecma262/#sec-generator-function-definitions

Etc...

Just a thought since the new ECMA spec is much easier to search/navigate/use.

ljharb:
The new one should be linked somewhere prominent (since it's a living spec), but the snapshots on ecma-international.org are both important, and the thing that the tables generally match against.

chicoxyzzy:
This also means we should update `2016+` tab using snapshots of 2016 and 2017 spec

bterlson:
ES2016 is snapshotted here: https://github.com/tc39/ecma262/2016.

chicoxyzzy:
Does it have any differences with www.ecma-international.org/ecma-262/7.0/ ?

chicoxyzzy:
I think we reached a consensus here. We'll use snapshots. So I'll close this issue. Please feel free to re-open it if you think we missed something.

bterlson:
@chicoxyzzy No difference.

PR #918, opened 2016-09-22
Use ES2016 snapshot links

chicoxyzzy:
As per discussion from #828

Please check misc links in this PR because I'm not 100% all of them are ok.

cc @bterlson

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -50,7 +50,7 @@ exports.tests = [
     name: 'exponentiation (**) operator',
     category: '2016 features',
     significance: 'small',
-    link: 'https://github.com/rwaldron/exponentiation-operator',
+    link: 'http://www.ecma-international.org/ecma-262/7.0/index.html#sec-exp-operator',
     subtests: [
       {
         name: 'basic support',
@@ -163,7 +163,7 @@ exports.tests = [
   },
   {
     name: 'Array.prototype.includes',
-    link: 'https://tc39.github.io/ecma262/#sec-array.prototype.includes',
+    link: 'http://www.ecma-international.org/ecma-262/7.0/index.html#sec-array.prototype.includes',
     category: '2016 features',
     significance: 'small',
     subtests: [
@@ -443,7 +443,7 @@ exports.tests = [
     name: 'generator functions can\'t be used with "new"',
     category: '2016 misc',
     significance: 'tiny',
-    link: 'https://github.com/rwaldron/tc39-notes/blob/master/es7/2015-07/july-28.md#67-new--generatorfunction',
+    link: 'http://www.ecma-international.org/ecma-262/7.0/index.html#sec-createdynamicfunction',
     exec: function(){/*
      function * generator() {
      yield 3;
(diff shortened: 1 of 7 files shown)


data-es2016plus.js
```diff
@@ -440,7 +440,7 @@ exports.tests = [
     name: 'generator functions can\'t be used with "new"',
     category: '2016 misc',
     significance: 'tiny',
-    link: 'https://github.com/rwaldron/tc39-notes/blob/master/es7/2015-07/july-28.md#67-new--generatorfunction',
+    link: 'http://www.ecma-international.org/ecma-262/7.0/index.html#sec-createdynamicfunction',
```
ljharb: This link seems useful to keep somewhere

data-es2016plus.js
```diff
@@ -464,7 +464,7 @@ exports.tests = [
     name: 'generator throw() caught by inner generator',
     category: '2016 misc',
     significance: 'tiny',
-    link: 'https://github.com/tc39/ecma262/issues/293',
+    link: 'http://www.ecma-international.org/ecma-262/7.0/index.html#sec-generatorfunction-objects',
```
ljharb: Also this

data-es2016plus.js
```diff
@@ -496,7 +496,7 @@ exports.tests = [
     name: 'strict fn w/ non-strict non-simple params is error',
     category: '2016 misc',
     significance: 'tiny',
-    link: 'https://github.com/rwaldron/tc39-notes/blob/master/es7/2015-07/july-29.md#611-the-scope-of-use-strict-with-respect-to-destructuring-in-parameter-lists',
+    link: 'http://www.ecma-international.org/ecma-262/7.0/index.html#sec-functiondeclarationinstantiation',
```
ljharb: Etc

ljharb (changes requested):
I like using snapshots, but we'd lose a lot of important information by erasing the links to meeting notes and github issues. Not sure what the best solution is.

chicoxyzzy:
@ljharb should I add another field for old links or just add a comment or revert to old links?

ljharb:
I think all of those are good options - maybe we should change "link" to be an object of "link name" to "link url", and then we can display them all in the UI?

bterlson:
There are two concepts here: links to spec text, and links to supporting info. Spec text should probably have a "spec:" key, other links can go into a links array and be displayed as superscript footnotes or something?

chicoxyzzy:
Updated

zloirock:
LGTM

zloirock:
However, I think it would be better to make notes better readable, not just links with 2 words.

zloirock:
"Meeting notes" - from what meeting? "GitHub issue" - which issue from which repo? Etc.

chicoxyzzy:
OK I'll try to add more meaningful notes

chicoxyzzy:
Please let me know if I can improve footnotes even more

zloirock:
LGTM

zloirock:
Inconsistency

```
[9] TC39 meeting notes from 28 July, 2015.
[10] TC39 meeting notes from July 28, 2015.
```

chicoxyzzy:
thank you. fixed

chicoxyzzy:
I'm still not sure that current spelling in footnotes is good enough. @ljharb @bterlson can you re-review please?

ljharb (approved):

PR #920, opened 2016-09-23
Updating results for Firefox 39 to Firefox 44

afmenez:
Updating results for Firefox 39 to Firefox 44.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -272,7 +272,7 @@ exports.browsers = {
   firefox37: {
     full: 'Firefox',
     short: 'FF 37',
-    obsolete: very,
+    obsolete: very
   },
   firefox38: {
     full: 'Firefox',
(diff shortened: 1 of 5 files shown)


data-es6.js
```diff
@@ -272,7 +272,7 @@ exports.browsers = {
   firefox37: {
     full: 'Firefox',
     short: 'FF 37',
-    obsolete: very,
```
ljharb, line 275: why remove the trailing commas?

afmenez: It was not on purpose, I removed these lines during my testing and copied back from FF36 when I finished.  But the comma is not necessary on the last item, and all other versions above this don't have it.

ljharb: We should be consistent; and they should probably all have it to keep diffs small.

afmenez: Done.

ljharb (approved):
Thanks, LGTM

PR #925, opened 2016-10-02
Added iOS 10 and moved iOS 8 and below to obsolete state

devoto13:
1. Added data for iOS 10, same as Safari 10 (data verified on my mobile phone). Looks like [adoption rate](https://mixpanel.com/trends/#report/ios_10/from_date:-29,report_unit:day,to_date:0) is already significant enough to have iOS 10 added.
2. Moved iOS 7-8 to obsolete state. It's a bit controversial, since iOS 8 has about 9% usage according to Apple's statistics, but I think it is fine since data is not lost anyway. And other sources report [much](https://mixpanel.com/trends/#report/ios_9) [lower](https://david-smith.org/iosversionstats/) usage for iOS 8.
3. Fixed support data for Intl for Safari 10 and iOS 10.

What do you think?


zloirock:
I'm not sure about the second point. The rest LGTM.

devoto13:
@zloirock I reverted marking iOS 8-9 as obsolete. After further thinking I agree it was not very good idea.

zloirock:
👍

PR #927, opened 2016-10-04
Remove obsolete FF Nightly footnote

chicoxyzzy:


zloirock:
It can be useful for someone.

chicoxyzzy:
@zloirock in that case we should keep all Nightlies browsers versions data (along with betas, auroras, TPs, canaries, etc) not only FF 45 and 48 Nightlies

chicoxyzzy:
also we shouldn't keep this data in footnote anyway IMO

data-es2016plus.js
```diff
@@ -63,7 +56,6 @@ exports.tests = [
           typescript: true,
           edge13: flag,
           edge14: true,
-          firefox48: firefox.nightly,
```
afmenez, line 66: This is implemented and is enabled by default only on nightly builds.  I believe the correct here would be to change firefox48 to "flag" and add a line with firefox52: nightly.

data-es2016plus.js
```diff
@@ -81,7 +73,6 @@ exports.tests = [
...
```
afmenez, line 84: This is implemented and is enabled by default only on nightly builds.  I believe the correct here would be to change firefox48 to "flag" and add a line with firefox52: nightly.

data-es2016plus.js
```diff
@@ -124,7 +115,6 @@ exports.tests = [
       babel: true,
       es7shim: true,
       typescript: typescript.corejs,
-      firefox45: firefox.nightly,
```
afmenez, line 127: I believe the correct here would be to change firefox45 to "flag", since it is what is implemented.

data-es2016plus.js
```diff
@@ -152,7 +142,6 @@ exports.tests = [
...
```
afmenez, line 155: I believe the correct here would be to change firefox45 to "flag", since it is what is implemented.

afmenez (changes requested):
I agree that the current implementation isn't good, but I believe your patch isn't perfect yet.

chicoxyzzy:
@afmenez changing `firefox.nightly` to `flag` seems incorrect. adding ff 52 to footnotes doesn't make any sense too

afmenez:
As I understand, these features are implemented but disabled by default, which is exactly what the flag tag means.
What is currently not nice is that if I open the test with the nightly version, I get different results from the ones listed, because the test says the features work, but the result table shows only the note.  Marking the features with the flag tag will improve this, because there will be a green corner on each result.  But ideally, the results would show as true for nightly to match with the test.

chicoxyzzy:
What I see here is that we have wrong results for FF 52 (this should be fix in another PR):


What I see here is that O.values and O.entries are available in Nightly and this is exactly what FF 52 tests results show


We can't mark feature as available under flag because this is not true for stable versions.
We should keep all unstable versions results instead of footnotes text if we want this information to be meaningful.

chicoxyzzy:
I don't agree with @zloirock that this info is useful, this info is hard to understand.

ljharb:
Useful and hard to understand aren't mutually exclusive. If it's hard to understand we should certainly make it easier, but having the data present is useful.

chicoxyzzy:
In my opinion it's reasonably useful and easy to understand if we will provide data only for last nightly version. I have no idea why does one need to use some nightly build of FF 45 and know if it supports exponential operator. Noone builds web applications for FF 45 Nightly. It's enough to know that _last_ Nightly support a feature.

chicoxyzzy:
What if only some builds of FF 45 Nightly support a feature and others don't? Should we keep builds info? I can agree that this might be useful info for someone but I can't agree such complexity worth it.

ljharb:
I think we only need the latest nightly info for any given major version - iow we don't need builds info.

chicoxyzzy:
> I think we only need the latest nightly info for any given major version

don't sure that footnotes is a good place for this type of information

chicoxyzzy:
Ok, I give up this time

PR #932, opened 2016-10-06
fill in missing node entries

hzoo:
It's green on the website right now so I'm not sure how it's there? http://node.green/ has it good for 6.7 and tested in 6.3 repl

More curious why it's not in the data but shows up as green

Actually it looks like it's missing a lot of data for node even though it's marked as green
## DONE

```
ENV(node): arrow functions/correct precedence
ENV(node): const/is block-scoped
ENV(node): const/cannot be in statements
ENV(node): const/redefining a const is an error
ENV(node): const/for loop statement scope
ENV(node): const/for-in loop iteration scope
ENV(node): const/for-of loop iteration scope
ENV(node): const/temporal dead zone
ENV(node): const/for-of loop iteration scope (strict mode)
ENV(node): let/basic support
ENV(node): let/is block-scoped
ENV(node): let/cannot be in statements
ENV(node): let/for loop statement scope
ENV(node): let/temporal dead zone
ENV(node): let/for/for-in loop iteration scope
// all of destructuring
ENV(node): destructuring, declarations
ENV(node): destructuring, assignment
ENV(node): destructuring, parameters
ENV(node): for..of loops/iterator closing, break
ENV(node): for..of loops/iterator closing, throw
ENV(node): function "name" property/new Function
ENV(node): function "name" property/variables (function)
ENV(node): function "name" property/object methods (function)
ENV(node): function "name" property/accessor properties
ENV(node): function "name" property/symbol-keyed methods
ENV(node): function "name" property/variables (class)
ENV(node): function "name" property/object methods (class)
ENV(node): default function parameters/basic functionality
ENV(node): default function parameters/explicit undefined defers to the default
ENV(node): default function parameters/defaults can refer to previous params
ENV(node): default function parameters/arguments object interaction
ENV(node): default function parameters/temporal dead zone
ENV(node): default function parameters/separate scope
ENV(node): default function parameters/new Function() support
ENV(node): rest parameters/basic functionality
ENV(node): rest parameters/function 'length' property
ENV(node): rest parameters/arguments object interaction
ENV(node): rest parameters/can't be used in setters
ENV(node): rest parameters/new Function() support
ENV(node): RegExp "y" and "u" flags / "y" flag
ENV(node): RegExp "y" and "u" flags / "y" flag, lastIndex
ENV(node): RegExp "y" and "u" flags / "u" flag
ENV(node): RegExp "y" and "u" flags / "u" flag, Unicode code point escapes
ENV(node): RegExp "y" and "u" flags / "u" flag, case folding
ENV(node): generators/can't use "this" with new
ENV(node): generators/%GeneratorPrototype%.return
ENV(node): generators/yield *, iterator closing
ENV(node): generators/yield *, iterator closing via throw()
```


data-es6.js
```diff
@@ -1005,6 +1005,7 @@ exports.tests = [
         safari10: true,
         webkit: true,
         chrome47: true,
+        node65: true,
```
ljharb: this test passes for me in node 6.0.0 - should this be `node6`?

hzoo: yeah you're right

hzoo:
maybe the code I have is incorrect but getting missing data on node for the below

ljharb:
If you're willing to fill in the missing data, that would be great :-)

hzoo:
Sounds like a good call for help 😛

data-es6.js
```diff
@@ -1524,6 +1532,8 @@ exports.tests = [
         firefox51: true,
         chrome40: flag,
         chrome41: true,
+        node012: flag,
```
hzoo, line 1535: This one was missed?

screen shot 2016-10-06 at 5 58 13 pm

Looks fine with `nvm use 4 && node .`

hzoo:
Ok did it all for node 😛

PR #933, opened 2016-10-06
Mark IE 9 and 10 as obsolete in all tabs

### LINKED ISSUE #861: What is the policy for retiring a browser to "obsolete : very"?

chicoxyzzy:
fixes https://github.com/kangax/compat-table/issues/861#issuecomment-252120515

### DIFF
--- a/data-es5.js
+++ b/data-es5.js
@@ -28,12 +28,18 @@ exports.browsers = {
     full: 'Internet Explorer 9',
     family: 'Chakra',
     short: 'IE 9',
-    obsolete: false
+    obsolete: true
   },
   ie10: {
-    full: 'Internet Explorer 10, 11',
+    full: 'Internet Explorer 10',
     family: 'Chakra',
-    short: 'IE 10-11',
+    short: 'IE 10',
+    obsolete: true
+  },
+  ie11: {
+    full: 'Internet Explorer 11',
+    family: 'Chakra',
+    short: 'IE 11',
     obsolete: false
   },
   edge13: {
--- a/data-esintl.js
+++ b/data-esintl.js
@@ -8,13 +8,13 @@ exports.browsers = {
     full: 'Internet Explorer',
     family: 'Chakra',
     short: 'IE 9',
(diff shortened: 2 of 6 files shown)


afmenez (changes requested):
Nice job, but you missed the Intl and Non-Standard tabs...

afmenez:
LGTM.

afmenez (approved):

zloirock:
MS still support IE maximal for supported platforms. For Vista it's IE9.

chicoxyzzy:
@zloirock that's true but why does IE 9 and 10 marked as obsolete in es6 and esnext-browsers?

zloirock:
I don't know.

afmenez:
It seems IE9 and IE10 are still supported after all:
http://www.welivesecurity.com/2016/01/12/microsoft-ends-support-for-old-internet-explorer-versions/

chicoxyzzy:
My main goal here is to make things consistent. Let's decide do we want to mark IE 9-10 as obsolete or stable.

afmenez:
It is not clear to me what "stable" means, But IE9 and IE10 are still supported on some platforms, and are actually the latest available version for them.  So it seems that for now they are stable indeed.

chicoxyzzy:
Looks like @webbedspace introduced changes for IE 10 (https://github.com/kangax/compat-table/commit/e085bc440252a449d033c3d32f4a721520f10956)

chicoxyzzy:
did @jdalton said that IE 9-10 is unsupported? https://github.com/babel/babel-preset-env/issues/15

jdalton:
Ya, IE9 is an odd one. While IE10 is dropped, IE9 technically still lingers for the Vista SP2 EOL (April 2017). However, the usage of Vista is so low that our guidance to devs is to consider it dropped already.

PR #934, opened 2016-10-10
Updating ES2016+ results for Firefox 52 (Nightly)

afmenez:
Updating ES2016+ results for Firefox 52 (Nightly).  It now has full exponentiation (**) support and allows trailing commas in function syntax.


chicoxyzzy:
Thank you

PR #936, opened 2016-10-13
Babel supports async iterations as of 6.16.0

danez:
https://github.com/babel/babel/releases/tag/v6.16.0


zloirock:
BTW please check results before adding - one of those tests contains a typo (`i++` instead of `++i`), so it's not passed. Sure, it's an issue of the test case, I'll fix it.

PR #951, opened 2016-11-08
update Chrome 56 results

chicoxyzzy:


data-es5.js
```diff
@@ -1628,7 +1633,21 @@ exports.tests = [
...
+      webkit: true,
+      chrome13: true,
+      chrome56: false,
+      opera12: true,
+      besen: true,
+      phantom: true,
+      ejs: true,
+      ios78: true,
+      android41: true,
+    },
```
chicoxyzzy, line 1650: this test is kind of obsolete because of `arguments.caller removed` test in ES2017

PR #955, opened 2016-11-10
Updating Atomics and SharedArrayBuffer support for Firefox 51 and 52

afmenez:
Updating Atomics and SharedArrayBuffer support for Firefox 51 Developer and Firefox 52 Nightly.

PR #956, opened 2016-11-11
Fix Shared Memory results for FF

chicoxyzzy:


afmenez:
The data on #955 was correct.  I'll clarify there.

afmenez:
Copying the answer here, just in case someone looks only at this thread.

What "firefox.nightly" means is that the feature is behind #ifdefs, and is compiled in only on Nightly builds. The feature will get out of the binary once this version is promoted to Developer (aka Aurora, aka Alpha).
What "firefox.developer" means is that the feature is behind #ifdefs, and is compiled in only on Nightly and Developer builds. The feature will get out of the binary once this version is promoted to Beta.

Both features you mentioned are new in Firefox 52, they have never been available in Firefox 51. If you look at the table, both features show "No" for FF51, and "No[7]" for FF52. Firefox 52 is currently the Nightly build, but will be promoted to Developer next week, and the features will still be available. They will not be anymore once Firefox 52 is promoted to Beta, ~7 weeks from now.

I'll try to make the notes clearer on my next patch.

chicoxyzzy:
Ok. Thank you

PR #960, opened 2016-11-13
Merge es6 browsers into esnext browsers

chicoxyzzy:
All results seem to be right except Konqueror, EchoJS, XS6 and JXA on 2016+ and next tabs. Should I filter them in `data-es2016plus.js` and `data-esnext.js` or add some property to browser data which indicates which tabs should render theirs test results?

cc @ljharb @zloirock @webbedspace


chicoxyzzy:
Also `esnext-browsers.js` should be renamed. Any ideas?

chicoxyzzy:
cc @hzoo this probably will affect `babel-preset-env` but I believe it is useful change for Babel

data-esnext.js
```diff
@@ -1,4 +1,4 @@
-var browsers = require('./esnext-browsers');
+exports.browsers = require('./esnext-browsers');
```
ljharb: re-exporting them seems kind of useless

ljharb (changes requested):
Maybe just `browsers`, and if we need to categorize them for things like "next", "6", "5", etc, we can do that in the data?

data-esnext.js
chicoxyzzy: @ljharb `build.js` expects `browsers` property to exist. In this PR I merged only es6 and esnext browsers. Browsers for es5, Intl and non-standard are still separate arrays. I agree that we should just import `browsers.js` into `build.js` but I think this could be done on some of next steps

ljharb: Makes sense.

chicoxyzzy:
Just realised `browsers` probably not a best name because we have transpilers, polyfills and other environments there

ljharb:
@chicoxyzzy good call - `environments`?

chicoxyzzy:
- [x] rename `browsers.js` to `environments.js` 
- [x] add filtering based on `test_suites` array property which contains tab names where environment should be presented (all tabs if there is no `test_suites` property)

build.js
```diff
@@ -35,18 +36,32 @@ var isOptional = function isOptional(category) {
   return (category || '').indexOf('annex b')>-1 || category === 'pre-strawman' || category === 'strawman (stage 0)';
 };
 
+var byTestSuite = function(suite) {
+  return function(browser) {
+    return ('test_suites' in browser) ? browser.test_suites.indexOf(suite)>-1 : true;
```
ljharb: instead of just checking `in`, could we use https://npmjs.com/has, and/or check that `Array.isArray`?

ljharb (changes requested):

build.js
chicoxyzzy: I think `Array.isArray` is ok here

chicoxyzzy:
@ljharb should I introduce any other changes to pass review?

chicoxyzzy:
@zloirock @webbedspace please take a look if you'll have time. Thank you! :)

hzoo:
Yeah if you want to PR env afterwards 👍  or I can after this is merged

chicoxyzzy:
@hzoo I'll file a PR

build.js
```diff
@@ -35,18 +36,32 @@ var isOptional = function isOptional(category) {
   return (category || '').indexOf('annex b')>-1 || category === 'pre-strawman' || category === 'strawman (stage 0)';
 };
 
+var byTestSuite = function(suite) {
+  return function(browser) {
+    return Array.isArray(browser.test_suites) ? browser.test_suites.indexOf(suite)>-1 : true;
```
ljharb, line 41: I think you mean >= here, not >-?

ljharb (changes requested):

build.js
chicoxyzzy: No, I mean exactly `>-1` as pre-ES2016 alternative to `Array.prototype.includes`, using same codestyle as anywhere else in this file

ljharb: Oh right sorry, duh. The lack of spaces around the operator threw me :-) I'll clean that up later, the spaces should be there.

ljharb (approved):
LGTM, but it should probably get more eyes

build.js
chicoxyzzy: we should migrate to ESLint with `eslint-config-airbnb` :)

zloirock:
LGTM.

chicoxyzzy:
Please check last 2 commits:
- first one fixes naming difference for Node.js 0.12 in different tabs
- second one makes it possible to override results from `equals` property (it's needed to override stage 3 `global` for Node.js 4 keeping other test results equal to Chrome 44) + optimises build removing second for-in loop run

chicoxyzzy:
I've triple checked all results and changes and compared each new test to old one and I believe it's all OK now

PR #961, opened 2016-11-15
Promote Firefox 50 to stable

mgol:


environments.js
```diff
@@ -317,11 +317,11 @@ module.exports = {
   firefox49: {
     full: 'Firefox',
     short: 'FF 49',
+    obsolete: true,
   },
   firefox50: {
     full: 'Firefox',
     short: 'FF 50 Beta',
```
zloirock: Please, change it.

chicoxyzzy:
LGTM

chicoxyzzy:
@mgol can you add FF Nightly 53 please? for now it has same results as FF 52 so there is no need to add any test results only browser.

Issue #740, opened 2016-02-03
create an official definition of obsolete


graingert:

graingert:
Is it based on current vendor support? Is it based on usage patterns?

Does only premium enterprise support count as obsolete?

kangax:
Thanks, good idea. We should have done this earlier. @webbedspace should know more on what we're doing now.

ljharb:
Some reasonable options, from which we can select multiple:
- not the latest version
- not officially supported by the engine vendor
- not in wide usage

graingert:
What about not officially supported vs enterprise only support?

graingert:
What about not officially supported vs only enterprise (paid) support?

ljharb:
Any form of support imo qualifies as "officially supported", because it means there will exist some users who are using it with those expectations.

webbedspace:
I don't have any strong opinions about defining non-obsolescence. The heuristics we've been using up to now have been these:
- The current stable version for auto-updating rapid-release browsers (Firefox and Chrome) with exception for Firefox Extended Support Release.
- Official MS support for the IE/Edge line.
- Apple support of particular OS X / iOS versions for the Safari line (though this may currently need updating).

There's no real heuristic for Node, however. Moreover, the ES5 and non-standard tables have been largely neglected w/r/t updating their browsers.

ljharb:
(fwiw node still has official support of 0.10 and 0.12 through 2016)

samsch:
As a user of the table, it would be useful if the previous stable versions were not dropped so quickly. For example, right now, Chrome 49 was just released, and is the current stable, but going by StatCounter (daily count), it doesn't even have as many users as Chrome 47, and 48 has barely started to decline. Firefox also suffers from this, as Firefox 45 is just released, but 44 is still the main version in use.

Perhaps a few weeks of overlap?

This isn't directly related, but it would also be nice to have the beta and alpha (Dev Edition) for Firefox back on the chart.

chicoxyzzy:
We should document this and add to `README.md` / `CONTRIBUTION.md`

graingert:
@chicoxyzzy a decision should be made first.

maybe remove the term "obsolete" and use "supported" and "unsupported"

where supported is clarified with "nb: IE support refers to support on Microsoft Windows desktop editions"

chicoxyzzy:
> maybe remove the term "obsolete" and use "supported" and "unsupported"

I agree. We should have `supported` and `release date` instead

[...]

PR #966, opened 2016-11-21
Updating "obsolete:very" for Chrome & Firefox

afmenez:
Adding release dates for Chrome & Firefox, marking as "obsolete: very" the versions that became obsolete more than one year ago.


afmenez:
Got Chrome dates from here:
https://en.wikipedia.org/wiki/Google_Chrome_version_history

And Firefox dates from here:
https://en.wikipedia.org/wiki/Firefox_version_history

chicoxyzzy:
LGTM

chicoxyzzy:
Maybe we should add release date to browser properties where possible? I think this info can be useful for projects like `babel-preset-env`, especially that `obsolete: very` doesn't really makes any sense for raw data. Not sure about property name and it's format. IMO format it should be human-readable and sortable `YYYY-MM-DD`. Thoughts?

chicoxyzzy:
Actually we should use release dates too (instead of `obsolete: very`) to hide super old browsers in resulting HTML

chicoxyzzy:
Please follow https://github.com/kangax/compat-table/issues/740#issuecomment-262230316

afmenez:
My proposal is to add two new fields:

released: YYYY-MM-DD,
retired: YYYY-MM-DD

Then the build script would follow these rules:

- If "obsoleted:very" or retired>1 year old, don't show the browser;
- If "obsoleted:true" or retired is in the past, the browser is obsolete;
- If "unstable:true" or released is in the future, the browser is unstable;

I can do the first part (add the dates for Firefox and Chrome), but I need help with the second (changing the script).

graingert:
@afmenez there's no reasonable definition of obsolete. Vendor supported or unsupported is something with a concrete definition.

> If "obsoleted:very" or retired>1 year old, don't show the browser;

each browser has different definitions on what they consider unsupported. compat-table shouldn't seek to define it for them with a specific time.

afmenez:
@graingert , I believe you read my message too fast.  I never tried to define what obsolete is, I specifically included I field to define a date by browser.   The line you are commenting proposes a way to define when a browser is irrelevant and should not appear on the site anymore.  I proposed 1 year because it is similar to what is already on the data today, and is easy to script with the data already present (or proposed).

ljharb:
Including date information is an excellent idea.

Programmatically using date information to determine obsoleteness is a very bad idea imo.

chicoxyzzy:
@ljharb we can use date information not for obsoleteness but for what we call `obsolete: very`. Just to hide very old browsers in HTML (before we have some filters in UI)

afmenez:
@ljharb , my proposal explicitly include a field for the date when the browser was declared obsolete (retired: YYYY-MM-DD), so we are not problematically determining anything.

ljharb:
I don't think "very" applies based on time either, but rather based on what new versions have replaced it, or how technologically out of date.

@afmenez i think explicit dates for each status makes perfect sense. That would also enable us, on the ES5 tab, to have a different concept of "obsolete" than on the ES6 tab.

afmenez:
Great, can I proceed with the fields I suggested?  I mean, I'll add the data now for informational purposes, and we can decide if/what to do with them later.

It would become like this:

```
  firefox31: {
    full: 'Firefox',
    family: 'SpiderMonkey',
    short: 'FF 31<br> ESR',
    released: 2014-07-22,
    retired: 2015-08-11,
    obsolete: very,
  }

  firefox45: {
    full: 'Firefox',
    family: 'SpiderMonkey',
    short: 'FF 45 ESR',
    released: 2016-03-08,
    retired: 2017-06-13,
    obsolete: false,
  },

  firefox52: {
    full: 'Firefox',
    family: 'SpiderMonkey',
    short: 'FF 52 Aurora',
    released: 2017-03-07,
    unstable: true,
    obsolete: false, // ESR (EOL on 2018)
  },
```

I would like to get some support because it is a lot of work, and I don't want to waste my time doing it and getting refused later.

afmenez:
Maybe "release:" and "retire:" would be better, since we can have dates in the future.

chicoxyzzy:
`release` and `retire` are OK IMO. Also I think we can change `build.js` later in another PR, after some further discussion in #740. So just adding new fields are enough for this PR

afmenez:
Got IE dates from here:
https://en.wikipedia.org/wiki/Internet_Explorer_version_history#Release_history_for_desktop_Windows_OS_version

And Edge dates from here:
https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history

chicoxyzzy:
Can you please rebase? It should be as easy as taking my changes from `gh-pages` and re-run `build` task?

chicoxyzzy:
I'm going to merge this now. Let's discuss what we mean by 'obsolete' and what should be obsolete and what not in #740

Issue #965, opened 2016-11-21
Raw data


chicoxyzzy:
### Current progress
- [x] remove results computations (with `Object.assign` and/or temporary variables) from test results #971
- [ ] move tests outside tests results (blocks converting test results to json)
- [ ] convert tests results and `data-common` to json
  - [ ] test results
  - [x] data-common https://github.com/kangax/compat-table/commit/6ee311a1f7b6e459e9527c0eb69eae1661efff34
  - [x] environments https://github.com/kangax/compat-table/commit/d107121153ceac188bd72cb26757e24cf590a8c0

chicoxyzzy:
Also #614 should be solved easier with second solution

chicoxyzzy:
I'm almost there #971

chicoxyzzy:
- [x] remove results computations (with `Object.assign`) from test results #971
- [ ] move tests outside tests results
- [ ] convert tests results and `data-common` to json

chicoxyzzy:
Don't have much time until end of the week. Further steps seem to be trivial. I'll finish rest of this soon.

PR #971, opened 2016-11-24
Raw data - Part 1

chicoxyzzy:
See #965


chicoxyzzy:
Next step is to move tests outside test results

PR #975, opened 2016-12-02
Chrome 55 is stable now

chicoxyzzy:


chicoxyzzy:
oops mistyped 56 instead of 55 :(

PR #977, opened 2016-12-06
[WIP] Add raw data generation

chicoxyzzy:
I found it much harder to build all data from JSON files (as I wanted in #965) than generating raw data from current js data files. Having js files allows to have autocomplete in editors when using shared data from 'data-common.json'. Also removing tests is not so trivial and it doesn't brings any significant profits. 

Also we need to register npm name and setup CI for automated package version bumping and publishing.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,14 @@
+node_modules/*
+es*/compilers/*
+.DS_Store
+.ruby-version
+_site/
+.idea/
+es5/
+es6/
+es7/
+es2016plus/
+esnext/
+esintl/
+non-standard/
+strict-mode/
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
     "type": "git",
     "url": "git://github.com/kangax/compat-table.git"
   },
-  "private": true,
   "dependencies": {
     "babel-core": "latest",
     "babel-polyfill": "latest",
@@ -39,6 +38,7 @@
   "scripts": {
     "lint": "jshint .",
     "build": "node build.js",
+    "gen-data": "node generate-raw-data.js",
     "pretest": "npm run lint",
     "build:compilers": "node build.js compilers",
     "git:clean": "if type git > /dev/null 2>&1; then git stash save --keep-index; fi",
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ es*/compilers/*
 .ruby-version
 _site/
 .idea/
+raw-data/
--- a/generate-raw-data.js
+++ b/generate-raw-data.js
@@ -1,0 +1,28 @@
+var fs = require('fs');
+var path = require('path');
+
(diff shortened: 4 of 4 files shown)


chicoxyzzy:
Oh I forgot to generate results for ranges and `equals`

package.json
```diff
@@ -10,6 +10,7 @@
     "type": "git",
     "url": "git://github.com/kangax/compat-table.git"
   },
+  "files": ["raw-data"],
```
ljharb: let's please not use the "files" key - this is a dangerous way to exclude things from npm. In addition, there's no need to exclude anything imo - https://unpkg.com/compat-table@0.0.1/ already has a bunch of other stuff.

ljharb (changes requested):
Why not gitignore `raw-data`, since this is published to npm?

.npmignore
```diff
@@ -0,0 +1,14 @@
+node_modules/*
+es*/compilers/*
+.DS_Store
+.ruby-version
+_site/
+.idea/
+es5/
```
ljharb, line 7: Why ignore all these?

.npmignore
```diff
@@ -0,0 +1,14 @@
...
+es6/
+es7/
+es2016plus/
+esnext/
+esintl/
+non-standard/
+strict-mode/
```
ljharb: Please ensure all files have a trailing newline

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "compat-table",
-  "version": "0.0.0",
+  "version": "0.1.0",
```
ljharb: Version numbers should be bumped only in commits directly to master; not in PRs

ljharb (changes requested):

.npmignore
chicoxyzzy: I don't see any reason why they should be downloaded when you need just test results from npm. Source code is available from GH because npm supports GH and Git

chicoxyzzy: IMO there should be only test results and environments

ljharb: github is for development, npm is for consumption. packages published to npm should have everything - that way it's available when consuming previously-installed-things offline as well.

chicoxyzzy: The only things that other projects need is raw data. Why publish HTML files?

ljharb: Why wouldn't they need the HTML files if they wanted to display them, or if they want to be able to view the HTML pages without an internet connection?

chicoxyzzy:
I'll create new PR instead of this one

williamkapke:
Ooh now I see what the goal is. 🤔  It wasn't immediately clear to me in #965.

So- you seek the results without the test functions.
... _others_ seek the test functions without the results!

Maybe they should be separated altogether. Although- that seems like it would add undesirable complication to this codebase.

ljharb:
I think it would.

PR #989, opened 2016-12-22
Mark FF Dev & FF Nightly to pass tests when feature is enabled in these versions

chicoxyzzy:
...while keeping flag that feature is available only in dev or nightly

### DIFF
--- a/data-common.json
+++ b/data-common.json
@@ -25,11 +25,13 @@
   },
   "firefox": {
     "developer": {
+      "versions": ["firefox52", "firefox53"],
       "val": false,
       "note_id": "firefox-developer",
       "note_html": "The feature is enabled by default only in Firefox Developer and Firefox Nightly."
     },
     "nightly": {
+      "versions": ["firefox53"],
       "val": false,
       "note_id": "firefox-nightly",
       "note_html": "The feature is enabled by default only in Firefox Nightly."
--- a/build.js
+++ b/build.js
@@ -471,8 +471,11 @@ function dataToHtml(skeleton, rawBrowsers, tests, compiler) {
     return ret;
   }
 
-  function testValue(result) {
+  function testValue(result, browserId) {
     if (result && typeof result === "object" && "val" in result) {
+      if (browserId && "versions" in result) {
+        return result.versions.indexOf(browserId) > -1;
+      }
       return result.val;
     }
     return result;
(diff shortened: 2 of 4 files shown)


chicoxyzzy:
Downside of this is that we should edit both `environments.json` and `data-common.json` with every new major release of FF. On the other side it reflects more accurate data because those versions of FF actually pass those tests.

afmenez:
Isn't there a way to define this data so that it is together with the browser versions in environments.json?

chicoxyzzy:
> Isn't there a way to define this data so that it is together with the browser versions in environments.json?

That's exactly what I did at first. To be honest I don't remember why I moved this info to `data-common.json`. I'll try to move it to `environments.json` one more time later today or tommorrow

chicoxyzzy:
> Isn't there a way to define this data so that it is together with the browser versions in environments.json?

OK this doesn't work. This is associated with:
- strategy of taking results from previous version test results (too **much** to refactor in html generation process)
- we should add to env not just a channel, but array of channels (`['nightly', 'developer']`) and that seems weird

afmenez:
How would this work with the "Array.prototype.values" issue on ES6?  It seems this implementation is not compatible with notes yet.

chicoxyzzy:
@afmenez good catch! Fixed

data-common.json
```diff
@@ -25,11 +25,13 @@
   },
   "firefox": {
     "developer": {
+      "versions": ["firefox52", "firefox53"],
```
ljharb, line 28: Is this meant to be a general purpose version mechanism for everything, so that we can consolidate all the nodeX's, for example?

chicoxyzzy: Yes, it's general purpose, but I don't understand what do you mean by consolidating node

ljharb: sorry, i'm not too familiar yet with your rearchitecting of all the infos. Can you explain what data-common is supposed to represent?

chicoxyzzy: This is data that can be used by various tabs
https://github.com/kangax/compat-table/blob/gh-pages/data-common.json

ljharb: So shouldn't version numbers only go in https://github.com/kangax/compat-table/blob/gh-pages/environments.json ?

chicoxyzzy: I've tried that twice working on this PR but that doesn't work well
https://github.com/kangax/compat-table/pull/989#issuecomment-269059385

ljharb: Sorry, I missed the earlier discussion. I think adding an array of channels, including "stable", to each version is a much less weird approach than duplicating the source of truth for versions.

chicoxyzzy: I agree with you. OK, I'll try to refactor build script when I'll have enough time.

PR #990, opened 2016-12-23
Updating ESnext results for Firefox 53

afmenez:
Updating ESnext results for Firefox 53 (21/dec Nigthly).  New feature, no surprises.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -832,6 +832,7 @@ exports.tests = [
     return typeof global === 'object' && global && !global.lacksGlobal && global.__system_global_test__ === 42;
   */},
   res: {
+    firefox53: true,
     node012: true,
     node4: true,
     node6: false,
--- a/esnext/index.html
+++ b/esnext/index.html
@@ -4661,7 +4661,7 @@ return typeof global === &apos;object&apos; &amp;&amp; global &amp;&amp; !global
 <td class="no" data-browser="firefox50">No</td>
 <td class="no unstable" data-browser="firefox51">No</td>
 <td class="no unstable" data-browser="firefox52">No</td>
-<td class="no unstable" data-browser="firefox53">No</td>
+<td class="yes unstable" data-browser="firefox53">Yes</td>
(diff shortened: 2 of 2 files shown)


ljharb (approved):
We do need to add subtests here tho, to ensure global is nonenumerable.

chicoxyzzy (approved):
LGTM

chicoxyzzy:
oops @ljharb din't noticed your comment. can we add this test in another PR?

current test passes for me

chicoxyzzy:
also I checked it's non-enumerable

ljharb:
In ff yes; not in node

ljharb:
Another PR is fine ofc

chicoxyzzy:
@afmenez do you want to add a test?

afmenez:
Sorry, I have no idea how to do this.  :(

chicoxyzzy:
I believe test should be

```js
typeof global === 'object' && 'global' in global && !global.propertyIsEnumerable('global');
```

I'll add PR myself later if nobody will do it before :)

ljharb:
`Object.prototype.propertyIsEnumerable.call(global, 'global')` would be safer, but otherwise good.

I'd prefer to do it in a separate subtest, not a full test, so that node can pass "has global as the global object", but fail "nonenumerable".

chicoxyzzy:
Shouldn't `typeof global === 'object'` be added anyway to prevent `ReferenceError` when there is no `global`?

chicoxyzzy:
also `Object.prototype.propertyIsEnumerable.call(global, 'global')` will return `false` if there is no `global` in `global` object. This is not safe IMO

ljharb:
@chicoxyzzy it's safe after you've done `'global' in global` - i was just saying, not to call `propertyIsEnumerable` directly on the object.

ljharb:
So actually, the test should be getting the actual global with `Function('return this')()`, and checking that.

ljharb:
I can write a PR for it ¯\\\_(ツ)_/¯

ljharb:
Submitted #991.

PR #993, opened 2016-12-24
Add shared memory to Chrome under flag

chicoxyzzy:
Didn't found any info from what version SharedArrayBuffer flag was added so I marked starting from lower version I have


chicoxyzzy:
It was introduced in [Chrome 48](https://github.com/tc39/ecmascript_sharedmem#firefox-and-chrome-implementations)

PR #995, opened 2017-01-01
arguments.caller removed from FF Nightly

chicoxyzzy:
Not sure if this was removed completely or only from Nightly

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -1243,6 +1243,7 @@ exports.tests = [
     res: {
       chrome56: true,
       safaritp: true,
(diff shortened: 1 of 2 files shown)


chicoxyzzy:
we should track https://github.com/tc39/ecma262/issues/562

afmenez:
I believe this is the merge:
https://bugzilla.mozilla.org/show_bug.cgi?id=1324208

There's no mention of a Nightly-only implementation.

chicoxyzzy:
OK I'm going to merge this

PR #1004, opened 2017-01-10
Remove global from FF Nightly

chicoxyzzy:
https://bugzilla.mozilla.org/show_bug.cgi?id=1317422#c17

https://github.com/tc39/proposal-global/issues/20#issuecomment-271647175


ljharb (approved):
Sadly

PR #1008, opened 2017-01-20
Test only for "do" syntax, instead of "let"

sheerun:
Currently this test verifies not only "do" syntax, but also "let".

I'm changing "let" to "var" so only "do" is tested..

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -51,7 +51,7 @@ exports.tests = [
   spec: 'http://wiki.ecmascript.org/doku.php?id=strawman:do_expressions',
   exec: function () {/*
     return do {
-      let x = 23;
+      var x = 23;
       x + 19;
     } === 42;
   */},


chicoxyzzy (approved):
LGTM

zloirock:
Hmm... `var` unlike `let` is not block scoped, `do` is a block. It will work, but I'm not sure that it will be a correct example.

chicoxyzzy:
we also can just omit any variables here. @zloirock what do you think about
```js
return do {
  if (true) {
    42;
  } else {
    'Pan Galactic Gargle Blaster';
  }
} === 42;
```
?

ljharb:
Are there any implementations that would pass this test with `var`, but not with `let`?

Unless there is one, I think leaving `let` is best.

zloirock:
@chicoxyzzy yep, as an option.

ljharb:
I think part of the test is that the do expression creates a scope. A block-scoped variable is required to test that. Any of these changes will be testing fewer things.

chicoxyzzy:
@sheerun please run `npm run build` and add changes to this PR

data-esnext.js
```diff
@@ -51,7 +51,7 @@ exports.tests = [
   spec: 'http://wiki.ecmascript.org/doku.php?id=strawman:do_expressions',
   exec: function () {/*
     return do {
-      let x = 23;
+      var x = 23;
```
ljharb, line 54: This must be a block-level variable to properly test block scope - in other words, it must be `let` or `const`.

Why do you think this is an important change to make? Does there exist any implementation that supports do expressions but not let/const?

ljharb (changes requested):

data-esnext.js
sheerun: I've just noticed that it might test something it's not supposed to, and sent PR. Your call

ljharb: In this case I'm claiming it is supposed to be testing that.

domenic: do expressions aren't specified, but once they are, it's not even clear they'll support var declarations inside of them.

PR #1010, opened 2017-01-22
tests for Promise.try (stage 1)

dnalborczyk:
tests for Promise.try, which is currently "stage 1" in the TC39 process.

https://github.com/ljharb/proposal-promise-try

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1803,7 +1803,7 @@ exports.tests = [
     {
       name: 'basic support',
       exec: function () {/*
-        return typeof Promise.try !== 'undefined';
+        return typeof Promise.try === 'function';
       */},
       res: {
       }
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1794,6 +1794,112 @@ exports.tests = [
     }
   ]
 },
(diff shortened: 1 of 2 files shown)


zloirock:
сс @ljharb

data-esnext.js
```diff
@@ -1794,6 +1794,112 @@ exports.tests = [
...
+{
+  name: 'Promise.try',
+  category: STAGE1,
+  significance: 'small',
+  spec: 'https://github.com/ljharb/proposal-promise-try',
+  'subtests': [
+    {
+      name: 'basic support',
+      exec: function () {/*
+        return typeof Promise.try !== 'undefined';
```
ljharb: This should be checking that it's a function, instead of just "not undefined"

ljharb (changes requested):
Thanks, this is great!

Can we also add tests that ensure it throws if it lacks a receiver (`this` value); that it works with thenables returned from the function' and that it works with subclassing?

ljharb:
Can we also make assertions that it has the name "try" and that it has a length of 1?

PR #1012, opened 2017-01-23
Adding note explaining global property reversal on Firefox and Webkit

afmenez:
Adding note explaining global property reversal on Firefox.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -901,11 +901,15 @@ exports.tests = [
     res: {
       firefox53: {
           val: false,
-          note_id: 'global-property',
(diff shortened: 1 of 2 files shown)


ljharb (commented):
This is also the case for WebKit.

afmenez:
Is there an issue open on their tracker?  Or can the same note be used for both browsers?

chicoxyzzy (commented):
https://bugs.webkit.org/show_bug.cgi?id=165171

ljharb (approved):
LGTM.

The next Safari TP will undoubtedly need the same note as webkit.

PR #1016, opened 2017-01-25
Add Safari 10.1 and latest WebKit

chicoxyzzy:


ljharb (approved):

mgol:
Why is Safari 10.1 marked as stable? Where does it come from? The latest macOS Sierra has Safari 10.0.3.

chicoxyzzy:
it's a mistake. it is available in macOS 10.12.3 beta so it should be marked as unstable

fixed https://github.com/kangax/compat-table/commit/09b4a8b853d0dc3915aa904c356908af54d6eda7

PR #1020, opened 2017-01-26
Shared memory and atomics: correct Symbol.species test

claudepache:
SharedArrayBuffer[Symbol.species] is an accessor, not a method


ljharb (approved):
seems legit

PR #1021, opened 2017-01-29
Move Shared Memory proposal to ES2017

chicoxyzzy:
also move RegExp named capture groups to stage 2


chicoxyzzy:
I'll update link to spec after Shared Memory part of spec will be merged into ecma262

PR #1029, opened 2017-02-12
TypeScript has implemented object rest/spread properties in v2.1.0

Pedro-vk:


data-esnext.js
```diff
@@ -778,6 +778,7 @@ exports.tests = [
     babel: true,
     jsx: true,
     chrome58: 'flagged',
+    typescript: typescript.corejs,
```
chicoxyzzy: this should be `true`, not `typescript.corejs`

data-esnext.js
```diff
@@ -794,6 +795,7 @@ exports.tests = [
...
```
chicoxyzzy: same here

chicoxyzzy (changes requested):
please run `npm run build` and add changes to your PR

data-esnext.js
ljharb: doesn't it require core.js to shim `Object.assign`?

Pedro-vk: It seems no:
https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#object-spread-and-rest

ljharb: Nothing in that link talks about it either way?

chicoxyzzy: @ljharb TS doesn't require `core.js` for that
https://www.typescriptlang.org/play/index.html#src=let%20%7B%20x%2C%20y%2C%20...z%20%7D%20%3D%20%7B%20x%3A%201%2C%20y%3A%202%2C%20a%3A%203%2C%20b%3A%204%20%7D%3B%0Alet%20n%20%3D%20%7B%20x%2C%20y%2C%20...z%20%7D%3B

chicoxyzzy (approved):

chicoxyzzy:
@Pedro-vk thank you!

Pedro-vk:
My pleasure :)

data-esnext.js
ljharb: Thanks, that clarifies.

However, that transpiled code violates the spec; it doesn't check for enumerability of symbols.

PR #1033, opened 2017-02-22
Add Node 7.6 results, fix the labels for node65 & node76 entries

mgol:
1. Add Node 7.6 results
2. Fix the labels for `node65` & `node76` entries.

Re 2: I originally added Node 6.5 results in https://github.com/kangax/compat-table/pull/913 with some doubts on what to put in their short names. @chicoxyzzy suggested just `Node 6.5`; it was correct back then but now it's only confusing - current latest Node 6 is `6.10.0` and Node 6 is an LTS starting from `6.9.0`. "Node 6.5" seems like it doesn't apply to them. On the other hand, writing `Node 6.5-6.10` will get obsolete once `6.11.0` comes out. So I opted for the semver-like `Node >=6.5 <7` & `Node >=7.6 <8`. Let me know what you think.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -1443,7 +1443,7 @@
   "node65": {
     "full": "Node.js",
     "family": "Node.js",
-    "short": "Node 6.5",
+    "short": "Node &gt;=6.5 &lt;7",
     "platformtype": "engine",
     "note_id": "harmony-flag",
     "equals": "chrome51",
@@ -1458,10 +1458,25 @@
   "node7": {
     "full": "Node.js",
     "family": "Node.js",
-    "short": "Node 7",
+    "short": "Node 7.0-7.5",
     "platformtype": "engine",
     "note_id": "harmony-flag",
     "equals": "chrome54",
+    "obsolete": true,
+    "test_suites": [
+      "es6",
+      "es2016plus",
+      "esnext",
+      "esintl"
+    ]
+  },
+  "node76": {
+    "full": "Node.js",
+    "family": "Node.js",
+    "short": "Node &gt;=7.6 &lt;8",
+    "platformtype": "engine",
+    "note_id": "harmony-flag",
+    "equals": "chrome55",
     "obsolete": false,
     "test_suites": [
       "es6",
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -925,6 +925,7 @@ exports.tests = [
       node6: true,
       node65: true,
       node7: true,
+      node76: true,
     }
   }, {
     name: '"global" global property has correct property descriptor',
@@ -961,6 +962,7 @@ exports.tests = [
       node6: false,
       node65: false,
       node7: false,
+      node76: false,
     }
   }]
 },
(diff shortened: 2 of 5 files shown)


chicoxyzzy (approved):
LGTM

Lets see what others think

environments.json
```diff
@@ -1458,10 +1458,25 @@
   "node7": {
     "full": "Node.js",
     "family": "Node.js",
-    "short": "Node 7",
+    "short": "Node 7.0-7.5",
```
ljharb, line 1461: should `node7` also include `7.6`? otherwise you'll need to add `node76` to everything that has `node7`, no?

mgol: I don't know the exact logic used to inherit stuff between different versions of the same thing. That is, Node versions are defined to equal specific Chrome versions and have only differences specified and as far as I see all Node versions are mentioned separately in such cases (just look at the two places where I added `node76`).

In any case, what would you propose? `7.0-7.5` catches the intent perfectly.

williamkapke: There are feature changes between 7.6 and 7.5 because the v8 version has been bumped up. I now have them grouped on node.green and it's easier to see when v8 changes cause feature changes.

mgol: The changes caused by the v8 version change are already covered by the `"equals": "chrome55"` field. All non-v8 specific things seem to be equal.

ljharb: I just want to make sure that anything passing in `node7` that's not explicitly failing in `node75` is implicitly passing in `node75`.

mgol: I see. What exactly is needed for one environment to extend results of another by default?

This shouldn't be specific to my PR, the Node 6/6.5 setup should share those problems.

ljharb: Agreed, it should, and it very well already might :-) I'm just hoping that during the course of this PR we can re-verify that that's the case.

ljharb: Agreed, it should, and it very well already might :-) I'm just hoping that during the course of this PR we can re-verify that that's the case.

chicoxyzzy: All should be fine. New version inherits data from previous versions unless otherwise specified in `data-*.js`-files

mgol: @chicoxyzzy How is it determined that two environments are really different versions of the same one? How is it determined which one is the newer?

chicoxyzzy: According to code in `build.js`
- If they have same `full`(name) or `family` then they are different versions of same thing
- `Object.keys()` is used for order so newer one should go after previous one

chicoxyzzy: Can you resolve conflict please? I'm going to merge this PR.

graingert: @chicoxyzzy conflict resolved in #1037

mgol: Ahh, sorry, I missed the rebase request comment. Do I need to do anything else then?

ljharb: @mgol it'd be ideal for this PR to be rebased and #1037 to be closed.

mgol: PR rebased.

PR #1042, opened 2017-03-07
Firefox 52 released, updating expiration data

afmenez:
Firefox 52 was released today this was already pushed.  This patch fixes some retirement dates and removes "ESR" from the current version to avoid confusion.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -537,7 +537,7 @@
     "short": "FF 38<br> ESR",
     "release": "2015-05-12",
     "retire": "2016-06-07",
-    "obsolete": "very"
+    "obsolete": true
   },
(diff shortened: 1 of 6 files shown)


chicoxyzzy:
Actually I think we should have both ESR and non-ESR versions but I'll accept these changes for now. Thank you!

afmenez:
Thanks.  The ESR and non-ESR are always identical feature-wise, so it is
easy to add both for every release.  I am not sure it is worth it though,
since it would add a lot of clutter with no actual gain on information...

chicoxyzzy:
@afmenez that's not always true. ESR may have some features disabled by default.

PR #1047, opened 2017-03-15
Update information on decorator proposal

### LINKED ISSUE #1003: Decorator proposal has been changed in Stage 2
The current test of class & property decorator is still based on the proposal in stage 1, and currently there is no implementation that fullfils the proposal in Stage 2, nor Babel or TypeScript.

It's not correct while list decorator in Stage 2 but test the proposal for decorator in Stage 1.

hybrist:
* Fixed link to current proposal
* Updated information on babel & typescript support
* Updated test case to reflect current proposal

Fixes https://github.com/kangax/compat-table/issues/1003

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -716,21 +716,23 @@ exports.tests = [
   name: 'class decorators',
   category: STAGE2,
   significance: 'medium',
-  spec: 'https://github.com/wycats/javascript-decorators',
+  spec: 'https://tc39.github.io/proposal-decorators/',
   exec: function(){/*
     class A {
       @nonconf
       get B() {}
     }
-    function nonconf(target, name, descriptor) {
-      descriptor.configurable = false;
-      return descriptor;
+    function nonconf(target) {
+      target.descriptor.configurable = false;
+      target.key = 'C';
+      return target;
     }
-    return Object.getOwnPropertyDescriptor(A.prototype, "B").configurable === false;
+    return Object.getOwnPropertyDescriptor(A.prototype, 'B') === undefined &&
+           Object.getOwnPropertyDescriptor(A.prototype, 'C').configurable === false;
   */},
   res: {
-    babel: {val: false, note_id: "regenerator-decorators-legacy", note_html: "Babel 6 still has no official support decorators, but you can use <a href='https://github.com/loganfsmyth/regenerator-plugin-transform-decorators-legacy'>this plugin</a>."},
-    typescript: true,
+    babel: {val: false, note_id: "regenerator-decorators-legacy", note_html: "Stage 2 decorators are in progress, see <a href='https://github.com/babel/babel/issues/2645'>babel/babel#2645</a>."},
+    typescript: {val: false, note_id: "typescript-decorators-legacy", note_html: "TypeScript currently implements an earlier, incompatible version of the decorator proposal."},
   }
 },
 {
--- a/esnext/index.html
+++ b/esnext/index.html
@@ -6093,22 +6093,24 @@ return result === &apos;tromple&apos;;
 <td class="no" data-browser="ios9">No</td>
 <td class="no" data-browser="ios10">No</td>
 </tr>
-<tr significance="0.5"><td id="test-class_decorators"><span><a class="anchor" href="#test-class_decorators">&#xA7;</a><a href="https://github.com/wycats/javascript-decorators">class decorators</a></span><script data-source="
+<tr significance="0.5"><td id="test-class_decorators"><span><a class="anchor" href="#test-class_decorators">&#xA7;</a><a href="https://tc39.github.io/proposal-decorators/">class decorators</a></span><script data-source="
 class A {
   @nonconf
   get B() {}
 }
-function nonconf(target, name, descriptor) {
-  descriptor.configurable = false;
-  return descriptor;
+function nonconf(target) {
+  target.descriptor.configurable = false;
+  target.key = &apos;C&apos;;
+  return target;
 }
-return Object.getOwnPropertyDescriptor(A.prototype, &quot;B&quot;).configurable === false;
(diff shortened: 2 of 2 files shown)


ljharb (approved):
Seems legit. @wycats, could you confirm?

chicoxyzzy:
Looks good! Waiting for someone who can confirm that tests are correct.

ghost:
@jkrems I noticed the legacy decorators link was removed from Babel. AFIAK that's the only way to allow use of decorators in Babel 6. Is there a particular reason to remove that when it's a clear path to making things work (and has been for the last year or so)?

wycats:
I don't think "pending a proposal update" is the current status.

data-esnext.js
```diff
@@ -716,21 +716,23 @@ exports.tests = [
...
+      return target;
     }
-    return Object.getOwnPropertyDescriptor(A.prototype, "B").configurable === false;
+    return Object.getOwnPropertyDescriptor(A.prototype, 'B') === undefined &&
+           Object.getOwnPropertyDescriptor(A.prototype, 'C').configurable === false;
   */},
   res: {
-    babel: {val: false, note_id: "regenerator-decorators-legacy", note_html: "Babel 6 still has no official support decorators, but you can use <a href='https://github.com/loganfsmyth/regenerator-plugin-transform-decorators-legacy'>this plugin</a>."},
-    typescript: true,
+    babel: {val: false, note_id: "regenerator-decorators-legacy", note_html: "Decorators are disabled in Babel v6, pending a proposal update – see <a href='https://github.com/babel/babel/issues/2645'>babel/babel#2645</a>."},
```
ljharb: Indeed, this is simply pending an implementation

hybrist: Copied that sentence from the official babel website. Agreed that it seems further along than that. I could drop the "pending a proposal update" and just defer to that issue link for the current state.

hybrist:
@jhabdas This bullet point is about the current decorator proposal, not the stage-0 version implemented by babel/typescript. There were changes to both the interface of decorators and the syntax (although the syntax changes are only minor). The legacy transform and decorators developed against it are not compatible with the current spec.

data-esnext.js
chicoxyzzy: @hzoo @loganfsmyth are we ok with this change?

ghost: @hzoo et. al - if we remove this it will steer users away from using legacy decorators and into the issue stream. if they use "legacy" decorators it's clear they're legacy and that spec is still pending.

hzoo: I changed that statement, we moved the legacy transform into the transform in stage-0 http://babeljs.io/docs/plugins/transform-decorators for Babel 7

hybrist: Does that mean that "real" stage 2 decorators won't land in babel 7? Or will they override the stage-0 decorators once landed?

hzoo: Work is being done, but instead of just throwing an error we are going to move the legacy transform as the one in the package, and yes we will update when landed.

Also we are going to independently version all proposal plugins so that they can be major bumped without bumping the rest of babel since we expect them to change. And maybe we need to make those 0.x instead to signify that.

hybrist: So babel 7 w/ `preset-stage-0` would encourage writing code that is almost guaranteed to break "randomly" during a babel 7 minor or patch version..? And given that many (most?) decorators will come out of some library, it wouldn't even be easy to fix those apps.

Or is the idea that `preset-stage-0` & `preset-stage-2` would both do a major bump, completely independent of babel 7, once the official decorators are implemented?

hybrist:
Rebase & updated wording to match babel website.

data-esnext.js
```diff
@@ -716,21 +716,23 @@ exports.tests = [
...
+      return target;
     }
-    return Object.getOwnPropertyDescriptor(A.prototype, "B").configurable === false;
+    return Object.getOwnPropertyDescriptor(A.prototype, 'B') === undefined &&
+           Object.getOwnPropertyDescriptor(A.prototype, 'C').configurable === false;
   */},
   res: {
-    babel: {val: false, note_id: "regenerator-decorators-legacy", note_html: "Babel 6 still has no official support decorators, but you can use <a href='https://github.com/loganfsmyth/regenerator-plugin-transform-decorators-legacy'>this plugin</a>."},
-    typescript: true,
+    babel: {val: false, note_id: "regenerator-decorators-legacy", note_html: "Stage 2 decorators are in progress, see <a href='https://github.com/babel/babel/issues/2645'>babel/babel#2645</a>."},
```
loganfsmyth, line 734: It seems like we should at least include legacy decorators with the same type of note Typescript has, saying that Babel has a transform that implements an earlier version of the spec too.

hybrist: My main reason for creating this PR is that I keep running into people who find the legacy transform, assume that it's basically the same thing, want to create libraries based on the interface, etc.. So I'd rather not promote it, not even in a footnote. The rollout of the revised decorators will be painful enough already without even more code tied to the legacy transform.

Fair point about TypeScript. Maybe the footnote could be removed because afaict there's no public signals yet that TypeScript wants to implement the current proposal anytime soon.

ghost: We should assume people are smart, and give them all the details.

hybrist: People don't have to be stupid to get confused by the stage process. Not everyone who uses JavaScript is an expert in TC39's process. And this website is consumed by a lot of everyday developers.

People *are* confused about what exactly "stage-0 decorators" and "stage-2 decorators" means. I know capable engineers who wrote a bunch of code against stage-0 decorators assuming that it'll mostly continue to work against stage-2 decorators. It's not like that's an unreasonable expectations, given the precedent set by other proposals that had little to no changes while going through the stages.

The audience of the kangax table is very diverse, so I think it shouldn't just dump all the information w/o considering the impact.

ghost: > People are confused about what exactly "stage-0 decorators" and "stage-2 decorators" means.

Perhaps that's where the effort should be placed, then.

chicoxyzzy: Just noticed `note_id`  is `"regenerator-decorators-legacy"`but it should be `"babel-decorators-legacy"`

chicoxyzzy (changes requested):

data-esnext.js
hzoo: yep the latter, except not "once the official decorators are implemented" but whenever the spec changes like a new stage, etc. although I know some wouldn't be opposed to the former given that if it's stage-0 you shouldn't be using it like that.

chicoxyzzy:
@jkrems could you please update your PR according to requested changes and rebase on `gh-pages` branch?

hybrist:
@chicoxyzzy I punted on updating the PR because it didn't feel like there was agreement on what the requested changes were. Is the only outstanding point to change the `note_id`?

hybrist:
The info on the website currently states that Typescript supports stage 2 decorators which still isn't quite correct. But I don't think this PR will land, given how much time has passed. Closing for now since I won't have time anytime soon to wrap this up.

PR #1054, opened 2017-03-19
add some Function.prototype.toString tests

michaelficarra:


ljharb (approved):
This LGTM but it'd be good to add some human-readable tests as well, for common (non-unicode char) cases.

michaelficarra:
@ljharb It's a bit hard to do that since the tests involve preserving comments and I have to write the whole test inside of a comment :confused:.

ljharb:
Sure, but not every engine passes `Function.toString.call(class Foo{}).startsWith('class Foo')` either :-/

michaelficarra:
Okay, would you like them in addition to these or in place of them?

ljharb:
Oh, in addition to, more tests are better :-)

michaelficarra:
@ljharb Okay, I added a few simple tests, one of which even IE11 passes!

chicoxyzzy (approved):

PR #1055, opened 2017-03-19
improved template escape sequence test

michaelficarra:
Now tests all previously invalid productions and tests both cooked and raw values for sections with invalid escapes and sections without invalid escapes.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -897,15 +897,14 @@ exports.tests = [
   category: STAGE3,
   significance: 'small',
   exec: function() {/*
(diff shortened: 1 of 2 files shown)


ljharb (approved):
LGTM - do your test changes change the results in firefox53, chrome59 with a flag, safari TP, or webkit?

michaelficarra:
I tested Firefox 53. It still passes. I'm not sure about the others, but I would be surprised.

chicoxyzzy (approved):

PR #1058, opened 2017-03-21
Add Template Literal Revision to ES2018 features

chicoxyzzy:
https://github.com/tc39/ecma262/pull/773#issuecomment-288177650

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -1819,11 +1819,33 @@ exports.tests = [
       firefox53: true,
     },
   },
+  {
+    name: 'template literal revision',
+    spec: 'https://github.com/tc39/proposal-template-literal-revision',
+    category: '2018 features',
+    significance: 'small',
+    exec: function() {/*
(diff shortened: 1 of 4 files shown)


afmenez (approved):
LGTM

afmenez:
On second thought, I wonder if this shouldn't go under "2018 misc" instead of "2018 features"...

chicoxyzzy:
> I wonder if this shouldn't go under "2018 misc" instead of "2018 features"...

Not sure either

afmenez:
Based on previous features classification, "2018 features" would be for entirely new features, like the SIMD proposal.  Refinements to existing features, like this one, would go under "2018 misc".  But I may have got it wrong.

ljharb (approved):
I think 2018 misc is more appropriate but I don't think it matters. Technically it is a feature since it enables DSLs that couldn't previously be implemented.

chicoxyzzy:
OK, I'll merge this as is for now

PR #1059, opened 2017-03-22
Promote RegExp features to stage 3

chicoxyzzy:
according to https://github.com/tc39/proposals/commit/e0bef6932441300cd54b8c0aed6d591e8a18a752

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1307,7 +1307,7 @@ exports.tests = [
 {
   name: 'RegExp named capture groups',
   spec: 'https://github.com/goyakin/es-regexp-named-groups',
-  category: STAGE2,
+  category: STAGE3,
   significance: 'small',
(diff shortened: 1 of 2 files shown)


afmenez:
I was about to do this too.  Why was it dropped?

chicoxyzzy:
@afmenez what do you mean by dropped?

afmenez:
It shows a red "x" in the patch and says this branch was deleted, so I assumed this was dropped.  But I see that the feature was indeed promoted already. :)

ljharb:
The x is because the PR was merged before tests finished (we probably shouldn't be doing that), and the branch was deleted after merging.

chicoxyzzy:
Yes Jordan is right. I run tests locally but I agree that I shouldn't have merge branch before tests pass on CI

Issue #1044, opened 2017-03-08
Intl tests are probably incorrect


chicoxyzzy:
Latest Chrome Canary, Firefox Nightly and WebKit Nightly fail on following test:
- calling Collator with Collator instance throws error
- calling NumberFormat with NumberFormat instance throws error
- calling DateTimeFormat with DateTimeFormat instance throws error

It looks like spec has changed and we should update / delete these tests

claudepache:
The spec has indeed changed between v1.0 and v2.0. If we remove or correct those tests (as we should), we should also update the links so that they don't point to the obsolete 1st edition anymore.

afmenez:
We should probably update all the results for these tests, since browsers that passed before may fail now.

chicoxyzzy:
other Intl-related issues:
https://github.com/kangax/compat-table/issues/1032
https://github.com/kangax/compat-table/issues/930
https://github.com/kangax/compat-table/issues/926
https://github.com/kangax/compat-table/issues/900
https://github.com/kangax/compat-table/issues/467

chicoxyzzy:
Some of these issues are duplicates. I will manage that as soon as I can (probably tommorrow).

claudepache:
This issue is a duplicate of #900... although the web-reality and the specced behaviours of those constructors are still disagreeing in a semi-unrelated way as of v3 (per https://github.com/tc39/ecma402/issues/57)

chicoxyzzy:
We'll back to this after March TC39 meeting

afmenez:
How about disabling the 3 tests while they are not updated, to avoid showing false negatives?

chicoxyzzy:
we can do it as a temporary solution

afmenez:
What is the best way to disable a test?

chicoxyzzy:
I think we can just comment them and add a comment that these tests should be updated

chicoxyzzy:
Duplicate of #467

PR #1061, opened 2017-03-29
Disabling esintl tests that became invalid with the updated spec

afmenez:
As described in issue #1044, three of the esintl tests are obsolete according to the updated spec.  This patch disables them until they can be fixed.

### DIFF
--- a/data-esintl.js
+++ b/data-esintl.js
@@ -85,7 +85,7 @@ exports.tests = [
         node012: true,
       },
     },
-// The spec was updated making this test ivalid.  It was disabled until it can be fixed
+// The spec was updated making this test invalid.  It was disabled until it can be fixed
 //    {
 //      name: 'calling Collator with Collator instance throws error',
 //      spec: 'http://www.ecma-international.org/ecma-402/1.0/#sec-10.1.1.1',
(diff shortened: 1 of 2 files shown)


chicoxyzzy:
Urgh... We should change something in build process to prevent conflict solving every time some PRs get merged.

chicoxyzzy (approved):

ljharb:
There's really no way to fix it unless we make travis build-and-commit after every master merge, which is super complex and brittle and awkward.

data-esintl.js
```diff
@@ -85,25 +85,26 @@ exports.tests = [
...
-      */},
-      res: {
-        ie11: true,
-        edge12: true,
-        firefox29: true,
-        chrome24: true,
-        node012: true,
-      },
-    },
+// The spec was updated making this test ivalid.  It was disabled until it can be fixed
```
ljharb: ivalid → invalid

data-esintl.js
```diff
@@ -251,25 +252,26 @@ exports.tests = [
...
```
ljharb: ivalid

data-esintl.js
```diff
@@ -363,25 +365,26 @@ exports.tests = [
...
```
ljharb: ivalid

ljharb (approved):

PR #1064, opened 2017-04-03
Fix MDN logo path

chicoxyzzy:
There was wrong path for MDN logo. I fixed it.


ljharb (approved):
sad to use a relative URL, but you're right that this works

ljharb:
We should probably set the base href so that `/mdn.png` works.

chicoxyzzy:
@KTamas could you review this PR?

chicoxyzzy:
I created new issue about base href https://github.com/kangax/compat-table/issues/1065

PR #1069, opened 2017-04-05
Add initial test runner and initial results for Duktape 2.0

svaarala:
Add initial ES5 and ES6 results for Duktape 2.0.

Also adds `duktape.js` which runs the tests using `./duk` and reports discrepancies between `data-*.js` results and actual results so that the data files can be fixed manually. (I'm not sure this is the correct process; it's a bit awkward for initial results but incremental updates should be easy enough.)

I'm new to compat-table so let me know if I've misunderstood something. Based on previous pulls it seems both data and HTML files are updated by each pull.

- [x] Use fs.readdir() to read in the tests
- [ ] Enable all tests, populate results (TBD: false vs. missing?)

### DIFF
--- a/duktape.js
+++ b/duktape.js
@@ -1,0 +1,109 @@
+/*
+ *  Node.js test runner for running data-*.js tests with Duktape 'duk' command.
+ *
+ *  Reports discrepancies to console; fix them manually in data-*.js files.
+ *  Expects a './duk' command in the current directory.  Example:
+ *
+ *    $ cp /path/to/duk ./duk
+ *    $ node duktape.js
+ */
+
+var fs = require('fs');
+var child_process = require('child_process');
+
+var testCount = 0;
+var testSuccess = 0;
+var testOutOfDate = 0;
+
+var dukCommand = './duk';
+
+// Key for .res (e.g. test.res.duktape20), automatic based on Duktape.version.
+var dukKey = (function () {
+    var stdout = child_process.execFileSync(dukCommand, [ '-e', 'print(Duktape.version)' ], {
+        encoding: 'utf-8'
+    });
+    var dukVersion = Number(stdout);
+    console.log('Duktape version is: ' + dukVersion);
+    return 'duktape' + (Math.floor(dukVersion / 10000)) + (Math.floor(dukVersion / 100 % 100));
+})();
+console.log('Duktape result key is: test.res.' + dukKey);
+
+// Run test / subtests, recursively.  Report results, indicate data files
+// which are out of date.
+function runTest(parents, test, sublevel) {
+    var testPath = parents.join(' -> ') + ' -> ' + test.name;
+
+    if (typeof test.exec === 'function') {
+        var src = test.exec.toString();
+        var m = /^function\s*\w*\s*\(.*?\)\s*\{\s*\/\*([\s\S]*?)\*\/\s*\}$/m.exec(src);
+        var evalcode;
+        if (m) {
+            evalcode = '(function test() {' + m[1] + '})();';
+        } else {
+            evalcode = '(' + src + ')()';
+        }
+        //console.log(evalcode);
+
+        var script = 'var evalcode = ' + JSON.stringify(evalcode) + ';\n' +
+                     'try {\n' +
+                     '    var res = eval(evalcode);\n' +
+                     '    if (res !== true && res !== 1) { throw new Error("failed: " + res); }\n' +
+                     '    console.log("[SUCCESS]");\n' +
+                     '} catch (e) {\n' +
+                     '    console.log("[FAILURE]", e);\n' +
+                     '    /*throw e;*/\n' +
+                     '}\n';
+
+        fs.writeFileSync('duktest.js', script);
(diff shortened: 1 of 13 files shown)


duktape.js
```diff
@@ -0,0 +1,109 @@
...
+        var newParents = parents.slice(0);
+        newParents.push(test.name);
+        test.subtests.forEach(function (v) { runTest(newParents, v, sublevel + 1); });
+    }
+}
+
+[
+    'data-es5',
+    'data-es6',
+    //'data-es2016plus',
```
ljharb: why are the other 4 commented out? we might as well get their results populated now too, even if it's a bunch of `false`s

ljharb: actually, it occurs to me that this should maybe use `fs.readdir` to get the list of files dynamically?

svaarala: What's the policy for populating with "falses"? I looked at the other results and for most tests it seems there ": true" entries but not ": false" when they don't pass (at least for newer tests).

I can add the other tests too, and populate them with falses if that's the right way?

I'll fix that fs.readdir() too.

svaarala: Ok, rebased and pushed, now uses fs.readdir() and runs all the tests.

`duktape20: true` has been added to all passing tests. I didn't add `duktape20: false`s yet - let me know what's the preferred way for that (i.e. omit failing tests by default or use false for them too).

There are two tests that pass incorrectly due to a Duktape bug (https://github.com/svaarala/duktape/pull/1463). I marked them "false" in the data files with a comment that they actually pass incorrectly. Should these be "true" instead? (It would seem odd because they would then incorrectly indicate the feature works when it really doesn't.)

svaarala:
Ok, rebased and pushed, now uses fs.readdir() and runs all the tests.

duktape20: true has been added to all passing tests. I didn't add duktape20: falses yet - let me know what's the preferred way for that (i.e. omit failing tests by default or use false for them too).

There are two tests that pass incorrectly due to a Duktape bug (svaarala/duktape#1463). I marked them "false" in the data files with a comment that they actually pass incorrectly. Should these be "true" instead? (It would seem odd because they would then incorrectly indicate the feature works when it really doesn't.)

duktape.js
ljharb: Falses should reflect known failures, as opposed to implicit failures.

If a test passes incorrectly, we should modify the test so this doesn't happen.

ljharb (approved):
This is great; I love the automated data population; we should do that for node.

We can defer the falses to another time.

duktape.js
svaarala: Hmm, so to be clear, a bug/known issue in an implemented feature would be "false" whereas something that's not implemented at all would just be missing?

ljharb: No, both would be false. Missing ideally should mean "we have no idea because nobody's checked yet".

We don't tend to differentiate between different ways of being incorrect' either it's correct or it fails the test (or it's impossible for it to be correct, if we're talking about shims in an older env)

svaarala: Re: that testcase passing, it's caused by a pretty specific Duktape bug: Duktape 2.0.0 incorrectly accepts anything like `/(?X.../` without a SyntaxError (there's a missing error clause). So, even though `/(?P.../` is not supported, it gets accepted and just happens to test true for `.test("aa")`. This is the test: https://github.com/kangax/compat-table/blob/gh-pages/data-non-standard.js#L741-L744. The other case is similar but in data-esnext.js.

To change the test to cover this particular bug seems a bit awkward so I was wondering if there are existing cases like this and how they're handled.

svaarala: Ok, got it. I can add the 'false's right away so there's no need to handle that later?

ljharb: I don't know how it's generally handled, but I'd say that if the test only is checking for the syntax error, it's not complete - it should also check for the substance of the feature.

svaarala: It's the reverse in this case: the test is expecting .test('aa') to test true, and expects a non-conforming engine to throw a SyntaxError. But because Duktape mistakenly parses the expression without a SyntaxError, and what's more the resulting regexp happens to match .test('aa'), the feature test passes incorrectly.

ljharb: Right - I'm saying "happens to match" is a bug in the test.

svaarala:
Ok, all falses are now in place - the only remaining issue is what to do about the two tests which are marked false (because the feature doesn't actually work but the test happens to pass):

```
data-esnext -> RegExp lookbehind: test result out of date, res: false, actual: true
data-non-standard -> RegExp named groups: test result out of date, res: false, actual: true
```

duktape.js
svaarala: Ok, pushed a fix which extends the condition a bit to detect the Duktape bug indirectly.

svaarala:
A minimal fix for the two problematic cases is now included in the pull, and the Duktape test run is now clean.

ljharb (approved):
LGTM

chicoxyzzy (approved):
awesome!

chicoxyzzy:
Is this ready to merge?

PR #1077, opened 2017-04-15
Add Duktape 2.1 results

svaarala:
- Add results for Duktape 2.1
- Move duktape engines in environments.json to fix their HTML layout grouping (they were grouped under "mobile" for some reason)

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -1500,6 +1500,20 @@
       "esintl"
     ]
   },
+  "duktape20": {
+    "full": "Duktape 2.0",
+    "family": "Duktape",
+    "short": "DUK 20",
+    "platformtype": "engine",
+    "obsolete": false
+  },
+  "duktape21": {
+    "full": "Duktape 2.1",
+    "family": "Duktape",
+    "short": "DUK 21",
+    "platformtype": "engine",
+    "obsolete": false
+  },
   "android40": {
     "full": "Android Browser",
     "short": "AN 4.0",
@@ -1598,13 +1612,6 @@
     "short": "iOS &gt;=10.3 &lt;11",
     "platformtype": "mobile",
     "equals": "safari10_1",
-    "obsolete": false
-  },
-  "duktape20": {
-    "full": "Duktape 2.0.x",
-    "family": "Duktape",
-    "short": "DUK 20",
-    "platformtype": "engine",
     "obsolete": false
   }
 }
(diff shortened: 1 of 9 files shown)


chicoxyzzy (approved):

data-es2016plus.js
```diff
@@ -37,6 +37,7 @@ exports.tests = [
           safaritp: true,
           webkit: true,
           duktape20: true,
+          duktape21: true,
```
afmenez: You don't need to add every version here.  What is true for `duktape20` is also true for `duktape20` and all following versions.

environments.json
```diff
@@ -1500,6 +1500,20 @@
       "esintl"
     ]
   },
+  "duktape20": {
+    "full": "Duktape 2.0.x",
```
afmenez: I believe you could drop de ".x", it is implicit in the other browsers listed here.

afmenez (changes requested):
Removing the redundancies as described in the comments would simplify this patch a lot.

data-es2016plus.js
svaarala: Ok, I'll make that change. But what metadata tells that duktape21 is a newer version of duktape20? There doesn't seem to be explicit metadata to that effect in environments.json. Or is it simply string sort based?

environments.json
svaarala: Ok, sounds good.

data-es2016plus.js
afmenez: I believe this is the purpose of the "family": "Duktape" field.  And yes, I believe it is sort based.

svaarala: Right, but how does one know X is newer than Y in the same family? I guess it's based on ordering in environments.json, i.e. same family but later in the file is newer?

svaarala: In any case the results are correct with the proposed change, so I'll apply that change to all the data files and update the pull. Thanks!

svaarala:
Suggested changes made, rebased and pushed.

svaarala:
Lint failed due to a duplicate key; fixed that.

afmenez (approved):
LGTM

svaarala:
The fix caused a HTML change and a Travis error from that. Hopefully both are now fixed :)

svaarala:
Ok, everything seems to pass now.

afmenez (approved):
LGTM

PR #1078, opened 2017-04-18
Fix a layout issue that breaks in Firefox

afmenez:
There's a rounding issue in Firefox which causes the first pixel row of the table header is obscured by the "% complete" box.  This patch is a workaround.

More details at https://github.com/webcompat/web-bugs/issues/5874.


afmenez:
According to the discussion in the bug mentioned above, the problem is indeed in the layout.  This workaround will work for the specified font but may break if there are font substitutions (on some platform that doesn't have it installed).

afmenez:
Do I have to do something else to get this in?

PR #1084, opened 2017-04-21
Add Chrome 60

sscotth:


data-esnext.js
```diff
@@ -1179,6 +1179,7 @@ exports.tests = [
         babel: true,
         typescript: typescript.corejs,
         duktape20: false,
+        chrome60: true,
```
chicoxyzzy: That's not true for Canary 60.0.3077.0

chicoxyzzy (changes requested):

data-esnext.js
sscotth: Are you sure?

screen shot 2017-04-21 at 3 44 44 am

chicoxyzzy: Yes I'm pretty sure.
screen shot 2017-04-21 at 10 46 53

Canary has A/B testing as was discussed before in older issues and PRs so maybe you are just the one of those lucky testers :)

sscotth: Interesting. I actually tried it on two (macOS) computers. I'll remove it.

afmenez:
How about retiring some old versions when a new one is added?  Chrome 50 was released more than a year ago, I believe anything older than that should be tagged with "obsolete:very" ...

sscotth:
That sounds more appropriate for a separate issue/pull request. It could even be automated in the build script with proper retire dates.

chicoxyzzy (approved):
I am ready to merge this after resolving conflicts 👍

chicoxyzzy:
Thank you @sscotth!

PR #1092, opened 2017-05-05
Updating the test for RegExp.prototype[Symbol.search] on the ES6 tab

afmenez:
According to [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1362379#c1), the specification has changed and the results are wrong.

I have updated the test itself and the results for Firefox 49-55, Chrome 51-59 and Edge 13-14, but I was unable to test for Safari10 and XS6.  If someone can re-run the tests on these and post the results here soon, I'll update the patch accordingly.  Or this can be done in a follow-up patch.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -8776,8 +8776,6 @@ exports.tests = [
         firefox55: true,
         chrome51: false,
         chrome56: true,
-        xs6: true,       // needs checking
-        safari10: true,  // needs checking
         duktape20: false,
       },
     },
--- a/data-es6.js
+++ b/data-es6.js
@@ -8768,15 +8768,16 @@ exports.tests = [
         var get = [];
         var p = new Proxy({ exec: function() { return null; } }, { get: function(o, k) { get.push(k); return o[k]; }});
         RegExp.prototype[Symbol.search].call(p);
-        return get + '' === "lastIndex,exec";
+        return get + '' === "lastIndex,exec,lastIndex";
       */},
       res: {
-        edge14: "flagged",
-        firefox49: true,
-        chrome51: true,
-        chrome56: false,
-        xs6: true,
-        safari10: true,
+        edge14: false,
+        firefox49: false,
+        firefox55: true,
+        chrome51: false,
+        chrome56: true,
+        xs6: true,       // needs checking
+        safari10: true,  // needs checking
         duktape20: false,
       },
     },
--- a/data-es6.js
+++ b/data-es6.js
@@ -8770,6 +8770,8 @@ exports.tests = [
         RegExp.prototype[Symbol.search].call(p);
         return get + '' === "lastIndex,exec,lastIndex";
       */},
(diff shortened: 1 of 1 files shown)


data-es6.js
```diff
@@ -8768,15 +8768,16 @@ exports.tests = [
...
-        chrome56: false,
-        xs6: true,
-        safari10: true,
+        edge14: false,
+        firefox49: false,
+        firefox55: true,
+        chrome51: false,
+        chrome56: true,
+        xs6: true,       // needs checking
+        safari10: true,  // needs checking
```
ljharb: if they need checking, let's just remove them from the list.

ljharb (changes requested):

data-es6.js
afmenez: Done.

ljharb (approved):

claudepache:
Safari (10.1 and TP) does not pass the updated test.

afmenez:
OK, the patch is still valid then.

claudepache:
Because the spec change that invalidated the old test (https://github.com/tc39/ecma262/pull/627) was made after ES6 was published, there should be a note saying that the tested semantics are those of a subsequent edition of ES (ES2017 I presume).

afmenez:
Done, but the note didn't appear on the generated html.  Is there a problem on my patch or is there a bug in the generator script?

data-es6.js
```diff
@@ -8768,15 +8768,16 @@ exports.tests = [
         var get = [];
         var p = new Proxy({ exec: function() { return null; } }, { get: function(o, k) { get.push(k); return o[k]; }});
         RegExp.prototype[Symbol.search].call(p);
-        return get + '' === "lastIndex,exec";
+        return get + '' === "lastIndex,exec,lastIndex";
       */},
+      note_id: "regexp-prototype-symbolsearch",
```
ljharb, line 8773: i believe the notes have to go on a "tr" key inside "res" (or on individual results), as opposed to just top-level here.

afmenez: I believe we do this when the note is about the result and not the test like this one.  There are several instances of notes on top-level ("block-level function declaration", "well-known symbols", "non-strict function semantics", etc).  The trouble is I found no other note associated with a subtest instead of the main test, and that's why I believe there's a bug (or limitation) on the generator.

ljharb: Gotcha. In that case we may need to modify the generator as well.

afmenez: I'll open an issue about this.  Do we have to wait until it is fixed to merge this one, or is it better to have it merged to work in the generator bug?

ljharb: I guess they can go in separately; but the generator enhancement can't go in without a note, so either this goes in first, or they go in in the same PR.

afmenez: I have opened #1098 with a reference to this one.  Hopefully this will be merged soon to allow that one to be fixed.

PR #1095, opened 2017-05-08
Improve SIMD tests and add spec links on ESnext tab

afmenez:
Most of the current tests check only one type (float or bool).  I am complementing them to also check for int and uint where applicable.
I am also adding appropriate spec links on all subtests.


afmenez:
I considered merging the "load" (load, load1, load2, load3) and "store" (store, store1, store2, store3) in single subitems, as they are in the spec.  Opinions?

ljharb:
I'd say tests should err on the side of oversplitting, not overconsolidating. Organizing subtests under tests is something I'm always in favor of, though.

afmenez:
Do you mean we should create subsubtests?  I don't think this is supported currently, but I am totally in favor.

ljharb:
I'm just saying in general and I would rather have a plethora of small tests, then a smaller number of consolidated tests. I don't have a specific opinion as it relates to SIMD.

afmenez:
OK, let's leave the tests as they are then.   I still need reviews and merge on this one though. ;)

data-esnext.js
```diff
@@ -271,9 +284,10 @@ exports.tests = [
     },
     {
       name: 'SIMD.%booleanType%.and',
+      spec: 'https://tc39.github.io/ecmascript_simd/#simd-and',
       mdn: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SIMD/and',
       exec: function(){/*
-        return typeof SIMD.Bool32x4.and === 'function' && typeof SIMD.Bool16x8.and === 'function' && typeof SIMD.Bool8x16.and === 'function';
+        return typeof SIMD.Bool32x4.and === 'function' && typeof SIMD.Int32x4.and === 'function' && typeof SIMD.Uint32x4.and === 'function';
```
ljharb: why is a boolean type test checking for Int32x4 and Uint32x4?

data-esnext.js
```diff
@@ -310,9 +326,10 @@ exports.tests = [
     },
     {
       name: 'SIMD.%type%.check',
+      spec: 'https://tc39.github.io/ecmascript_simd/#simd-check',
       mdn: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SIMD/check',
       exec: function(){/*
-        return typeof SIMD.Float32x4.check === 'function';
+        return typeof SIMD.Float32x4.check === 'function' && typeof SIMD.Int32x4.check === 'function' && typeof SIMD.Uint32x4.check === 'function' && typeof SIMD.Bool32x4.check === 'function';
```
ljharb: it might be easier to maintain here if we defined a single array of boolean types, and another of all types, and iterated it rather than having to hardcode them in multiple places?

ljharb (changes requested):

data-esnext.js
afmenez: Because the spec says this operation is available for all these types:

SIMD.Int8x16.and(a, b)
SIMD.Int16x8.and(a, b)
SIMD.Int32x4.and(a, b)

SIMD.Uint8x16.and(a, b)
SIMD.Uint16x8.and(a, b) 
SIMD.Uint32x4.and(a, b)

SIMD.Bool8x16.and(a, b) 
SIMD.Bool16x8.and(a, b) 
SIMD.Bool32x4.and(a, b)

I felt it was better to check all types than all sizes of one type.

ljharb: then the test name shouldn't be "booleanType", i'd think?

data-esnext.js
afmenez: Yep, but I don't know how to do this. :(
And it is more complicated than that.  Some operations support only Float, some support Float+Int+Uint, some support Bool+Int+Uint, and some support all.  But then some operations support only some of the sizes.  The array would probably have to be defined for each test.

data-esnext.js
afmenez: Ah, you mean the name of the test.  i haven't noticed it was wrong.  I'll fix that in a moment.  For all operations.

afmenez: Done for all operations.

data-esnext.js
ljharb: Just defining an array of strings (like `Float32x4`) at the top of the file, and reference it where needed, should do it

afmenez: Sorry, I am not a ninja JavaScripter like you guys.  I tried doing it like this:
```
return 'Float32x4,Int32x4,Uint32x4,Bool32x4'.split(',').every(function(type){
    return typeof SIMD.[type].check === 'function';
});
```
But it always fails.

ljharb: `SIMD[type]`, not `SIMD.[type]` - the latter is a syntax error.

Also, rather than splitting, let's just create the array directly - and then reuse it across tests if possible.

data-esnext.js
```diff
@@ -91,6 +91,14 @@ exports.tests = [
       name: 'basic support',
       mdn: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SIMD',
       exec: function () {/*
+        simdFloatTypes=['Float32x4'];
```
ljharb, line 94: hmm - due to the eval-like nature of the tests, i don't think this will actually work. I think they might need to be globals, or else inlined into each test :-/

ljharb (changes requested):
Thanks - this might be tricky to get right, but i think we can do it.

data-esnext.js
afmenez: I was under the impression that if I declared them without the preceding "var" they would have global scope.  And they seem to be working like this: all the tests still pass, but they fail if I use the wrong array on a given test.

ljharb: Ha, good call. id prefer explicit globals but this works too.

afmenez: Great.  The instructions on the home page says to prefer the most compatible form when coding tests, I believe this is it.  But github is indicating that you still expect changes, did I miss something?

ljharb: nope, looks good :-)

ljharb (approved):

chicoxyzzy (approved):

Issue #1096, opened 2017-05-09
Remove String trim-left/right from non-standard?


adam77:
As they're being tracked in esnext. Confusing to track them in two places.

adam77:
Also, how about removing the rows from non-standard for which no tracked engine has support?

ljharb:
Which rows are those? Be sure you're checking all platforms, even obsolete ones.

ljharb:
As to the OP, I'm fine moving all the nonstandard data for trim* to esnext.

adam77:
\__count\__
\__present\__

appear to be No in all columns (for example)

i'm guessing FF has dropped a bunch of obsolete es4 era stuff.

ljharb:
Sorry, was mobile previously. 

It seems that `count` should be true for Firefox 4, but the browser isn't in the list for some reason. I think that we should go back and add a browser entry such that every row *does* have a Yes in it - I don't think we should remove a historical reference to a feature, even if it no longer applies.

adam77:
that's of historical interest but is it any use to anyone?

ljharb:
It's of use to me whenever I have to fix bugs in the es5-shim.

Also, it might potentially be of use to one person in 100 years, which means it's worth keeping forever anyways.

afmenez:
I have already proposed #1025, so we would still keep historical reference but would not pollute the table.

afmenez:
I already merged the removal on #1097, but I can't close this one.

PR #1097, opened 2017-05-10
Remove String.trimLeft/Right from non-standard tab

afmenez:
They are now a proposed standard, already tested on the ESnext tab.  This patch removes the test form the non-standard tab and consolidates the results on the ESnext tab.


afmenez:
Fixes #1096.

ljharb (approved):

PR #1100, opened 2017-05-12
Unify the version naming convention

mgol:
So far some browsers were just omitting a dot in the version string and some
replaced it by an underscore. This commit changes all those versions to use
the underscore translation. This will reduce confusion; with some environments
using rapidly increasing major version numbers it becomes hard to distinguish
"41" from "4.1" when both are translated to "41" in the browser name.

The only exception is Duktape which needs to have names maching keys matching
what Duktape.version reports.

Also:
### 1)
Rename Duktape shortcuts from "DUK 20" to "DUK 2.0" & "DUK 21" to "DUK 2.1"

The previous versions were incorrect. Lots of engines have high major version
numbers so version 21 sounded plausible.

### 2)
Change iOS 10.0-10.3 to 10.0-10.2

iOS 10.3 was incorrectly included in a group with 10.0-10.2.


ljharb (approved):
We should fix the duktape result generation as well.

mgol:
> We should fix the duktape result generation as well.

This means there will be no direct matching between the key we use and the one reported by `Duktape.version`. Is that OK?

ljharb:
`Duktape.version` reports `duktape20`? O.o

mgol:
Unfortunately, yes...

mgol:
Hmm, no, actually it seems the comment is not 100% correct. What it actually reports is:
```
duk> Duktape.version
= 20100
```
Even weirder but then if we already have to transform it we might as well add an underscore.

ljharb:
yeah, that's a really stupid version string; it should use semver. I see no value in trying to match that.

mgol:
It's not even a string, it's a number. :)

mgol:
@ljharb Should I change that script in this PR and unify Duktape entries as well? Or in a new one?

ljharb:
Let's do another commit in this pr

mgol:
@ljharb PR updated.

ljharb (approved):

environments.json
```diff
@@ -1265,7 +1265,7 @@
       "non-standard"
     ]
   },
-  "safari51": {
```
afmenez: This will conflict with my patch on #1099

environments.json
```diff
@@ -1283,7 +1283,7 @@
     "short": "SF 6.1,<br>SF 7",
     "obsolete": true
   },
-  "safari71_8": {
+  "safari7_1_8": {
```
afmenez: This is not Safari 7.1.8, but Safari 7.1 & Safari 8.  Also fixed on my patch on #1099

afmenez (changes requested):
It would be really nice if you could revert the Safari changes so we can merge my more complete patch for Safari (#1099) without conflicts.

mgol:
@afmenez You mean Safari & iOS changes, don't you?

@ljharb Does that sound OK to you? I can revert those particular changes.

afmenez:
Yep, I did a larger change on Safari and iOS, adding some versions and results, and it already included all the changes you did.  If you revert just these parts, I believe both of our patches will be able to merge.  :)

environments.json
ljharb: Do you think it'd be easier to merge yours first, and then rebase this one? or would it be easier to merge this, and rebase yours on top?

afmenez: I believe both would be about the same amount of work.  But I pushed my patch first... :)

mgol: @ljharb I can wait for the @afmenez's patch to be landed & then rebase mine. I'm fine with solving those merge conflicts, I just hope I won't have to do that multiple times.

ljharb: ok, sounds good

mgol:
PR rebased.

PR #1101, opened 2017-05-15
SharedArrayBuffers and Atomics to be enabled by default on Firefox 55

afmenez:
According to this [comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1361520#c7), SharedArrayBuffers and Atomics to be enabled by default on Firefox 55.


ljharb:
Since FF 55 isn't out yet, it's probably too soon to claim that the tests pass.

afmenez:
The implementation is already available on FF 53 and 54 behind flags, and they do pass.  Nightly builds for FF 55 are out and it does pass too.

data-es2016plus.js
```diff
@@ -819,6 +819,7 @@ exports.tests = [
           firefox46: firefox.nightly,
           firefox51: firefox.developer,
           firefox53: firefox.sharedmem,
+          firefox55: true,
```
ljharb, line 822: then this should say `firefox.nightly`, no?

afmenez: Nope, this is for features that are present on nightly builds but the code is not on the released version (wrapped with #ifdef).  A flagged feature is present but disabled by default (which is the current situation for Firefox 53 and 54).  For Firefox 55, the feature will be enabled by default, which grants it a "true".

chicoxyzzy (approved):
LGTM

ljharb (approved):

Issue #1089, opened 2017-04-27
Add android 6-8 to the table


mlucool:
Android 6/7 common and 8 is coming ([see here](https://en.wikipedia.org/wiki/Android_version_history). Can it be added to https://kangax.github.io/compat-table/es6/?

afmenez:
There's no Android Browser after Android 4:
http://androidandme.com/2012/02/applications/goodbye-old-browser-chrome-to-become-the-standard-browser-on-android-4-0-and-above/
So maybe the real issue here should be "remove Adroid 4+ from the table".  I don't know where those results come from (Chrome for Android? Samsung Browser?), but they don't seem to be named correctly.

mlucool:
Yeah, it is confusing. I guess something that would be useful is moving android browser to 'obsolete platforms' .

afmenez:
I'll do that once the pull queue moves forward since this would all the patches there.

afmenez:
I believe this can be closed now.

PR #1104, opened 2017-05-16
Reorganizing Android Browser versions

afmenez:
- Adding metadata for all versions that had a browser (no results yet);
- Obsoleting ancient versions;
- Removing newer Android versions that are just aliases for Google Chrome (>4);

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -1609,12 +1609,28 @@
     "release": "2009-04-30",
     "obsolete": "very"
   },
+  "android1_6": {
+    "full": "Android Browser 1.6 (Donut)",
+    "family": "Android Browser",
+    "short": "AN 1.6",
+    "platformtype": "mobile",
+    "release": "2009-09-15",
+    "obsolete": "very"
+  },
   "android2_0": {
     "full": "Android Browser 2.0 (Eclair)",
     "family": "Android Browser",
     "short": "AN 2.0",
     "platformtype": "mobile",
     "release": "2009-10-26",
+    "obsolete": "very"
+  },
+  "android2_1": {
+    "full": "Android Browser 2.1 (Eclair)",
+    "family": "Android Browser",
+    "short": "AN 2.1",
+    "platformtype": "mobile",
+    "release": "2010-01-12",
     "obsolete": "very"
   },
   "android2_2": {
@@ -1641,6 +1657,22 @@
     "release": "2011-02-22",
     "obsolete": "very"
   },
+  "android3_1": {
+    "full": "Android Browser 3.1 (Honeycomb)",
+    "family": "Android Browser",
+    "short": "AN 3.1",
+    "platformtype": "mobile",
+    "release": "2011-05-10",
+    "obsolete": "very"
+  },
+  "android3_2": {
+    "full": "Android Browser 3.2 (Honeycomb)",
+    "family": "Android Browser",
+    "short": "AN 3.2",
+    "platformtype": "mobile",
+    "release": "2011-07-15",
+    "obsolete": "very"
+  },
   "android4_0": {
     "full": "Android Browser 4.0 (Ice Cream Sandwich)",
     "family": "Android Browser",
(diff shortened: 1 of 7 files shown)


afmenez:
This patch addresses #1089 .

chicoxyzzy (approved):

mgol:
What about Android Browser 4.2 & 4.3?

afmenez:
It's complicated.  As of Android 4.1, Chrome became the default Android browser.  But it is still possible to build the Android Browser unrtil 4.4, and some distros like Cyanogen do.  But it's the same version that was in 4.1.
This link has a decent explanation: https://decadecity.net/blog/2013/11/21/android-browser-versions

mgol:
I don't know if there are any differences between Android Browser in Android 4.2 & 4.3 but Can I Use reports some differences between 4.1 & 4.2/4.3: https://caniuse.com/#compare=android+4.1,android+4.2-4.3

So at least Android 4.2 should still be included.

mgol:
Another issue is that [babel-preset-env](https://www.npmjs.com/package/babel-preset-env) uses `compat-table` data so if you don't include Android 4.2 or 4.3, an entry like `Android >=4.3` will generate a bundle that may not compile in Android 4.3.

I think we should just include every browser version and - in particular - never remove existing ones from the data. Android 4.2 & 4.3 were present so it's best to not remove them.

afmenez:
OK, I have added all the missing versions up to Android 4.4.  I don't see the point of faking versions 5.0 and beyond, since it is just the Chrome browser.

mgol:
Yeah, I don't care that much about versions 4.4 & newer as they're based on Chrome. My remark was relevant for versions up to 4.3.

environments.json
```diff
@@ -1595,50 +1595,117 @@
...
+    "family": "Android Browser",
     "short": "AN 4.3",
     "platformtype": "mobile",
-    "obsolete": true
+    "release": "2013-07-24",
+    "obsolete": "very"
   },
   "android4_4": {
-    "full": "Android Browser",
+    "full": "Android Browser 4.4 (KitKat)",
```
mgol: I don't think this is correct. Android Browser ended with Android 4.3.

Also, it lacks  the `"equals": "chrome30"` declaration which was previously present. Doesn't this break correctness of the results?

As for me, Android 4.4 & newer was never correct as there's no Android Browser in Android 4.4 & newer so I'd be fine with removing them (If maintainers agree).

mgol (changes requested):

environments.json
afmenez: The info I have is that Chrome became the default browser for Android on JellyBean (4.1), and that's why I had removed 4.1-4.4 initially.
But my info also says that the sources were still available until KitKat (4.4), and some distros shipped with it.  I checked the link that you provided, and it mentions activity on 4.4 too (https://caniuse.com/#compare=android+4,android+4.1,android+4.2-4.3,android+4.4&compare_cats=JS).  And that's why I added them back.

afmenez:
The info I have is that Chrome became the default browser for Android on
JellyBean (4.1), and that's why I had removed 4.1-4.4 initially.
But my info also says that the sources were still available until KitKat
(4.4), and some distros shipped with it.  I checked the link that you
provided, and it mentions activity on 4.4 too (
https://caniuse.com/#compare=android+4,android+4.1,android+4.2-4.3,android+4.4&compare_cats=JS).
And that's why I added them back.

On Tue, Jun 6, 2017 at 7:02 PM, Michał Gołębiowski <notifications@github.com

environments.json
mgol: What Can I Use names as Android 4.4 browser is actually Chrome. You can see they have a separate entry for 4.4.3-4.4.4 as that's when the Chromium-based (!) WebView was updated to a newer Chromium version.

The link you provided at the beginning also says that the WebView in Android 4.4 is Chromium-based and that there is no "Browser" browser there.

afmenez: Done.

chicoxyzzy:
@mgol do you approve this PR now?

mgol (approved):
Looks good! Thanks for listening me out. :)

PR #1105, opened 2017-05-30
Add results for Chrome 61 Canary

chicoxyzzy:
Needs review because Shared memory could be enabled in my Canary build because of feature testing.


chicoxyzzy:
Also added commit ae77d64 here. Chrome 59 stable will be released tomorrow.

ljharb (approved):

afmenez (approved):
LGTM

chicoxyzzy:
Thank you for review! I'll merge this after Chrome 59 release!

afmenez:
Then you forgot to change Chrome 59 to stable...

afmenez:
While we're at it, why do we keep adding an Opera reference on every Chrome release?  We don't do this for any other derivative browser (Iceweasel, Tor, and so on...).

afmenez (approved):
LGTM

PR #1106, opened 2017-05-31
use the faster to install google-closure-compiler-js

graingert:


build.js
```diff
@@ -269,7 +268,7 @@ process.nextTick(function () {
         fs.writeFileSync(fpath, code);
         var output;
         try {
-          output = "" + child_process.execSync('node_modules/closurecompiler/bin/ccjs ' +
+          output = "" + child_process.execSync('./node_modules/.bin/google-closure-compiler-js ' +
```
ljharb, line 271: we shouldn't ever be (or have been) hardcoding `node_modules` - let's use `which('google-closure-compiler-js')` and https://www.npmjs.com/package/npm-which instead?

ljharb (changes requested):
Thanks, seems like a good change.

build.js
graingert: require('npm-which')(__dirname) or process.cwd() ?

ljharb: `process.cwd()`, i think, since all commands should be run from project root (assuming both work)

graingert: everything in this file hardcodes node_modules, and I don't want to touch everything in this PR

ljharb: alright, fair enough - we can leave it to a separate PR

ljharb (approved):

chicoxyzzy (approved):

chicoxyzzy:
Thank you @graingert!

PR #1108, opened 2017-06-05
Add Safari 11

chicoxyzzy:


ljharb (commented):
How are you installing Safari 11 to test this? I don't have it available on my Mac OS app update; it's not on browserstack; iOS hasn't released with it yet.

chicoxyzzy:
@ljharb I have Apple Developer Account. It's developer betas. I have both MacOS High Sierra and iOS 11 installed on my devices.

environments.json
```diff
@@ -1381,6 +1381,13 @@
     "release": "2017-03-27",
     "obsolete": false
   },
+  "safari11": {
+    "full": "Safari 11",
+    "family": "JavaScriptCore",
+    "short": "SF 11",
+    "release": "2017-06-05",
+    "unstable": false
```
afmenez: Was it released already?  I was under the impression that it is still in beta...

environments.json
```diff
@@ -1717,5 +1724,14 @@
...
     "release": "2017-03-27",
     "obsolete": false
+  },
+  "ios11": {
+    "full": "iOS Safari",
+    "family": "iOS Safari",
+    "short": "iOS 11",
+    "platformtype": "mobile",
+    "equals": "safari11",
+    "release": "2017-06-05",
```
afmenez: I believe it was not released yet, so this date seems wrong...

afmenez (changes requested):
Almost there.

afmenez (approved):
LGTM

environments.json
ljharb: good call; beta should be considered unstable.

chicoxyzzy: Yes, I fixed that. Thank you @afmenez

PR #1111, opened 2017-06-12
Firefox 54 to be released tomorrow (jun/13)

afmenez:
Firefox 54 will be released tomorrow (june/13).  This patch updates the version table accordingly.


chicoxyzzy (approved):
1 comment. otherwise 👍

environments.json
```diff
@@ -601,7 +601,7 @@
     "short": "FF 46",
     "release": "2016-04-26",
     "retire": "2016-06-07",
-    "obsolete": true
+    "obsolete": "very"
```
chicoxyzzy, line 604: i think this is a mistake

chicoxyzzy (changes requested):

environments.json
afmenez: The rule I am using is true after retirement, and "very" one year later.  This lets me remove one version every time I add a new one.
ESR releases work differently, and usually there is more than one supported version at a time.  That's why I am removing 46 but 45 is still active.

afmenez:
Firefox 54 is already out and this can be merged.

environments.json
chicoxyzzy: Oh, 45 is ESR. OK than

chicoxyzzy (approved):

chicoxyzzy:
Thank you!

PR #1112, opened 2017-06-13
Updating some spec titles, links and grouping on the esnext tab

afmenez:
The esnext tab has fallen out of sync with the current proposals.  What I have done:
- Updated the titles of some of the specs according to the latest specs;
- Updated the links to some specs to the latest version;
- Created a new 'Class and Property Decorators' group, with the old ´class decorators' tests only for now;
- Created a new 'object rest/spread properties' group, with the previous rest/spread single tests inside;
- Created a new 'class fields' group, with the previous tests for 'class properties' and 'private fields' inside;

What I could not resolve:
- 'Math methods for 64-bit integers' seems to be dropped, is it?
- Stage 0 proposals

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -50,7 +50,7 @@ exports.tests = [
   name: 'do expressions',
   category: STAGE1,
   significance: 'small',
-  spec: 'http://wiki.ecmascript.org/doku.php?id=strawman:do_expressions',
+  spec: 'https://gist.github.com/dherman/1c97dfb25179fa34a41b5fff040f9879',
   exec: function () {/*
     return do {
       let x = 23;
@@ -63,7 +63,7 @@ exports.tests = [
   }
 },
 {
-  name: 'function.sent',
+  name: 'Generator function.sent Meta Property',
   category: STAGE2,
   significance: 'small',
   spec: 'https://github.com/allenwb/ESideas/blob/master/Generator%20metaproperty.md',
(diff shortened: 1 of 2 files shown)


chicoxyzzy:
> 'Math methods for 64-bit integers' seems to be dropped, is it?

They are called Arbitrary-precision Integers [(stage 2 now)](https://github.com/tc39/proposals/commit/50b9dbfb5645493d92ec672bbd4a88031d17f929)
https://github.com/tc39/proposal-integer

ljharb:
I don't think that's accurate; they're separate proposals.

chicoxyzzy:
Yes, AFAIU 64bit Math proposal became unnecessary with Arbitrary-precision Integers, so former may be dropped. Sorry for my previous comment, it seems inaccurate and confusing.

afmenez:
Well, the new BigInt seems to be a completely new approach to the problem, and the current tests would not apply.  I'll just remove them.

afmenez:
I considered moving the tests to the non-standard tab, but since this feature was only implemented on transpilers and had no real engine implementation, I dropped them.

chicoxyzzy:
I'l try this branch locally later today. LGTM so far

chicoxyzzy (approved):
👌

Iooks perfect

ljharb (approved):

chicoxyzzy:
Thank you!

PR #1114, opened 2017-06-17
Dropping ancient IE versions

afmenez:
Microsoft has EOLed all versions of IE but IE11 more than a year ago:
https://www.microsoft.com/en-us/windowsforbusiness/end-of-ie-support

I am retiring these ancient versions from all tables.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -144,11 +144,12 @@
     "family": "Chakra",
     "short": "IE 7",
     "release": "2006-10-18",
-    "obsolete": true,
+    "retire": "2016-01-12",
+    "obsolete": "very",
     "test_suites": [
       "es5",
       "es6",
-      "esnext",
+      "esintl",
       "non-standard"
     ]
   },
@@ -157,11 +158,12 @@
     "family": "Chakra",
     "short": "IE 8",
     "release": "2009-03-19",
-    "obsolete": true,
+    "retire": "2016-01-12",
+    "obsolete": "very",
     "test_suites": [
       "es5",
       "es6",
-      "esnext",
+      "esintl",
       "non-standard"
     ]
   },
@@ -170,14 +172,16 @@
     "family": "Chakra",
     "short": "IE 9",
     "release": "2011-03-14",
-    "obsolete": true
+    "retire": "2016-01-12",
+    "obsolete": "very"
   },
   "ie10": {
     "full": "Internet Explorer",
     "family": "Chakra",
     "short": "IE 10",
     "release": "2012-10-26",
-    "obsolete": true
+    "retire": "2016-01-12",
+    "obsolete": "very"
   },
   "ie11": {
     "full": "Internet Explorer",
--- a/environments.json
+++ b/environments.json
@@ -148,7 +148,6 @@
     "obsolete": "very",
     "test_suites": [
       "es5",
-      "esintl",
       "non-standard"
     ]
   },
@@ -188,6 +187,7 @@
     "test_suites": [
       "es5",
       "es6",
+      "es2016plus",
       "esintl",
       "non-standard"
     ]
--- a/environments.json
+++ b/environments.json
@@ -148,7 +148,6 @@
     "obsolete": "very",
     "test_suites": [
       "es5",
-      "es6",
       "esintl",
       "non-standard"
     ]
@@ -159,10 +158,9 @@
     "short": "IE 8",
     "release": "2009-03-19",
     "retire": "2016-01-12",
-    "obsolete": "very",
+    "obsolete": true,
     "test_suites": [
       "es5",
-      "es6",
       "esintl",
       "non-standard"
     ]
@@ -173,7 +171,12 @@
     "short": "IE 9",
     "release": "2011-03-14",
     "retire": "2016-01-12",
-    "obsolete": "very"
+    "obsolete": true,
+    "test_suites": [
+      "es5",
+      "esintl",
+      "non-standard"
+    ]
   },
   "ie10": {
     "full": "Internet Explorer",
@@ -181,7 +184,13 @@
     "short": "IE 10",
     "release": "2012-10-26",
     "retire": "2016-01-12",
-    "obsolete": "very"
+    "obsolete": true,
+    "test_suites": [
+      "es5",
+      "es6",
+      "esintl",
+      "non-standard"
+    ]
   },
   "ie11": {
     "full": "Internet Explorer",
--- a/es2016plus/index.html
+++ b/es2016plus/index.html
@@ -110,6 +110,7 @@
 <th class="platform closure compiler" data-browser="closure"><a href="#closure" class="browser-name"><abbr title="Closure Compiler v20161024">Closure</abbr></a></th>
 <th class="platform typescript compiler" data-browser="typescript"><a href="#typescript" class="browser-name"><abbr title="TypeScript 1.8 + core-js 2.4">Type-<br>Script +<br><nobr>core-js</nobr></abbr></a></th>
 <th class="platform es7shim compiler" data-browser="es7shim"><a href="#es7shim" class="browser-name"><abbr title="es7-shim">es7-shim</abbr></a></th>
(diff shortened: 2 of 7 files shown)


ljharb:
I'm totally fine to add the EOL dates, but imo while "EOL" may mean "unsupported by the browser vendor" it doesn't mean "unused" or "unsupported by websites".

IE 9+ is still heavily used, and I think it's pretty important that they show up on all the tables by default.

afmenez:
I am not sure what you mean by "heavily used", but available statistics point to the contrary.
According to Statcounter, IE9 is less popular than IE8, and the sum of IE7 to IE10 amounts to 0.77%:
http://caniuse.com/usage-table

ljharb:
It depends on the site, and the country. Global stats aren't actually useful for determining usage on individual sites.

Either way, the primary consumer here isn't imo people looking to see what the latest browser supports, it's people who want to see what "not the latest" browser supports, because that's what's hard to test for themselves.

chicoxyzzy:
IMO we should keep this versions until extended end of support

ljharb:
Also, 0.77% of 1 billion people (just to pick an arbitrary number of internet users) is 7.7 million humans. That's quite a lot.

chicoxyzzy:
BTW I think we can mark IE 7 as very obsolete. It comes with Windows Vista which end of extended support date is April 11, 2017

ljharb:
Yes, I'm comfortable marking IE 7 as very obsolete :-)

afmenez:
OK, new patch.  I am keeping IE7 out, and I am bringing back IE8-IE10 where they make sense.  I believe there's no point in having IE9 on the esnext table, for instance.

environments.json
```diff
@@ -144,11 +144,11 @@
...
     "short": "IE 7",
     "release": "2006-10-18",
-    "obsolete": true,
+    "retire": "2016-01-12",
+    "obsolete": "very",
     "test_suites": [
       "es5",
-      "es6",
-      "esnext",
+      "esintl",
```
afmenez: IE7 is zeroed on the esintl table, should it be removed?

chicoxyzzy (approved):

environments.json
```diff
@@ -157,11 +157,11 @@
     "family": "Chakra",
     "short": "IE 8",
     "release": "2009-03-19",
+    "retire": "2016-01-12",
     "obsolete": true,
     "test_suites": [
       "es5",
-      "es6",
-      "esnext",
+      "esintl",
```
afmenez, line 163: IE8 is zeroed on the esintl table, should it be removed?

es2016plus/index.html
```diff
@@ -5937,9 +5765,7 @@
 <td class="no not-applicable" data-browser="closure" title="This feature is optional on non-browser platforms, and doesn&apos;t contribute to the platform&apos;s support percentage.">No</td>
 <td class="no not-applicable" data-browser="typescript" title="This feature is optional on non-browser platforms, and doesn&apos;t contribute to the platform&apos;s support percentage.">No</td>
 <td class="no not-applicable" data-browser="es7shim" title="This feature is optional on non-browser platforms, and doesn&apos;t contribute to the platform&apos;s support percentage.">No</td>
-<td class="no obsolete" data-browser="ie9">No</td>
-<td class="yes obsolete" data-browser="ie10">Yes</td>
-<td class="yes" data-browser="ie11">Yes</td>
+<td class="no" data-browser="ie11">No</td>
```
afmenez: I believe this is a bug:  the data says this is true for ie10, but ie10 is not on this table anymore.  But this should not cause the feature to be negated here for ie11.

afmenez (commented):
I believe we should wait for the bug in the es2016+ table to be fixed before pushing this one.

afmenez:
Please see my comments on the review before pushing this one.

environments.json
ljharb: I'd say that the oldest version in a browser line that should be displayed is the newest one that has zero support - so if IE 8 also has zero support and is visible, then it makes sense to hide IE 7.

environments.json
ljharb: (same logic as above; if IE 9 is zeroed and displayed, then it makes sense to hide IE 8)

environments.json
afmenez: Removed.

environments.json
afmenez: Nope, IE9 has half of the items implemented.  I'll keep it like this then.

es2016plus/index.html
afmenez: If I follow the rule by @ljharb above, I'll have to add IE10 (0 items implemented) since IE11 has one item implemented, which would hide the bug again.   This doesn't make much sense since the browsers were obsolete before the standard was in discussion.
I'll also have to add IE9 (0 items implemented) and IE10 (one item implemented) to the es2016+ table, which also don't make much sense to me.

ljharb: I think that in general it's important to be able to see when support for a thing started landing.

It would also be a fair policy to show the oldest browser that had nonzero support (instead of the newest with zero), as long as we're consistent.

mgol:
@chicoxyzzy 
> IMO we should keep this versions until extended end of support

What do you mean by "extended end of support"? IE<11 are not supported by Microsoft in their consumer versions at all, the extended support period has ended as well. Only server/embedded versions are supported on some versions of Windows.

chicoxyzzy:
@mgol I mean this https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet

chicoxyzzy:
| OS                 | preinstalled browser | end of extended support |
|-------------|----------------------|--------------------------|
|~Vista~              | IE 7                              | April 11, 2017                    |
|Windows 7    | IE 8 (IE 10*)                 | January 14, 2020              |
|Windows 8   | IE 10 (IE 11**)               |  January 10, 2023                 |

\* IE 10 comes with Windows 7 SP 1 update which is necessary for extended EOL
\** IE 11 comes with Windows 8.1 update which is necessary for extended EOL

chicoxyzzy:
Hmm... looks like Windows 7 support is available only with SP 1. I'll update my table

mgol:
@chicoxyzzy This is a list of supported Windows versions, not IE versions. Microsoft has dropped support for IE<11 on these systems 1.5 years ago.

mgol:
Similarly how Windows 7 is supported only with SP 1, IE on Windows 7 is supported only if updated to v11. I don't really see the difference, anything below is unsupported.

chicoxyzzy:
Ok, I can agree with this, despite the fact that every previous version of Chrome and Firefox (excluding ESR versions) are not supported as well. By the way it's still useful to have them in table considering theirs usage. Actually IE 8 usage is almost equal to Safari 9 + Safari 9.1 usage. Also either IE8 / Safari 9 usage is greater than Edge 15 usage so IE 8 compatibility info is still useful.

mgol:
> By the way it's still useful to have them in table considering theirs usage. Actually IE 8 usage is almost equal to Safari 9 + Safari 9.1 usage. Also either IE8 / Safari 9 usage is greater than Edge 15 usage so IE 8 compatibility info is still useful.

Sure; I agree with what @ljharb said:
> while "EOL" may mean "unsupported by the browser vendor" it doesn't mean "unused" or "unsupported by websites"

So this information may be useful enough to enough many people to still list those versions on the site.

chicoxyzzy:
@afmenez @ljharb do you think that this PR is ready to merge?

afmenez:
I think so.  I believe all the comments have been addressed, and I added back IE8-IE10 where they make sense.
The bug I mentioned above does nor manifest anymore now that IE10 is back at the es2016+ table.

ljharb (approved):
Let's still file a separate issue about that IE 10 bug, because that should be fixed.

LGTM in the meantime.

chicoxyzzy:
Thank you, as always! :)

afmenez:
I filed issue #1118 about the bug.

PR #1115, opened 2017-06-21
Update test results for Chrome 60 beta

mgol:
Tests run on Chrome beta 60.0.3112.32 on macOS.


data-es2016plus.js
```diff
@@ -824,7 +824,7 @@ exports.tests = [
           firefox53: firefox.sharedmem,
           firefox55: true,
           chrome48: chrome.sharedmem,
-          chrome61: true,
```
ljharb, line 827: Im confused; why are there v61 results here before v60 is released?

mgol: Usually 2 Chrome pre-releases are shown in the table; people run tests on Canary before it reaches beta.

ljharb (approved):

data-es2016plus.js
chicoxyzzy: shared memory was turned on by default in v8 6.0 just recently (when it became beta)

chicoxyzzy (approved):

chicoxyzzy:
Thank you!

PR #1116, opened 2017-06-22
Object rest and Object spread were removed from Safari and Safari TP

chicoxyzzy:
For some reason Object rest and Object spread were removed from Safari 11 and Safari TP 33. I think it's also applies to Webkit Nightly but I can't check this (Webkit Nightly doesn't work on macOS 10.13 yet)

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1045,7 +1045,7 @@ exports.tests = [
         chrome60: true,
         safari11: false,
         safaritp: false,
-        webkit: true,
+        webkit: false,
         duktape2_0: false,
       }
     },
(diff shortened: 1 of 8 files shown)


ljharb (approved):
It'd be great if we could figure out what happened :-/

chicoxyzzy:
@ljharb I found some info about that
https://bugs.webkit.org/show_bug.cgi?id=167962#c48
https://bugs.webkit.org/show_bug.cgi?id=167963#c19

both features were rolled out in r216891 and weren't fixed yet so I think it's ok to mark them as unsupported in latest Webkit Nightly r218633

chicoxyzzy:
Probably we should investigate the problem to add some tests

ljharb (approved):

Issue #1026, opened 2017-02-07
ES6: Add tests for updated identifier syntax


claudepache:
Reference:

ES6: http://www.ecma-international.org/ecma-262/6.0/#sec-names-and-keywords

compare with ES5.1: http://www.ecma-international.org/ecma-262/5.1/#sec-7.6

Testcases:

(1) The following program

```js
var ⸯ;
```

(where ⸯ is U+2E2F VERTICAL TILDE) is correctly rejected in Chrome, Edge and Firefox Nightly, but is incorrectly accepted in Safari and Firefox 51.

Webkit bug: https://bugs.webkit.org/show_bug.cgi?id=121541
Mozilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=917436 (fixed in FF52)

(2) The following program

```js
var 𐋀;
```

(where 𐋀 is U+102C0 CARIAN LETTER G) is correctly accepted in Firefox, Chrome and Edge, but is incorrectly rejected in Safari, .

Curiously, `var \u{102C0};` does work in Safari.

Webkit bug: https://bugs.webkit.org/show_bug.cgi?id=167940

(3) The testcase `var v\u0061r;` (should be rejected) is already included in the compat table (in Miscellaneous > no escaped reserved words as identifiers).

(4) Are there other interesting cases?

afmenez:
I can try to write the tests.  Do you have title suggestions for (1) and (2)?

afmenez:
I am trying to write the tests.  Do you have title suggestions for (1) and (2)?

I have tried to use the same "framework" set by the existing (3) test, but was not successful.
(1) For this one the test is not passing on Firefox 54, Chrome 59 and Edge 14, but according to the description it should (on what versions?).  Any hint?
```
      exec: function() {/*
        var ⸯ;
        try {
          eval('var ⸯ');
        } catch(e) {
          return true;
        }
      */},
```
(2) For this one the test is passing on Firefox 54, Chrome 59 and Edge 14, and it is failing on IE11.  Can assume it is OK?  What are the first version of each browser where the test should pass?
```
      exec: function() {/*
        var 𐋀;
        try {
          eval('var 𐋀');
          return true;
        } catch(e) {
          return false;
        }
      */},
```
(3) I just moved the test from the misc section to the new one.  I assume it is already fine.

claudepache:
> (1)
> exec: function() {/*
> var ⸯ;
> try {
> eval('var ⸯ');
> } catch(e) {
[...]

PR #1121, opened 2017-06-30
Adding tests for updated identifier syntax

afmenez:
This one addresses #1026.
I don't know how to name the new tests, so I just used the declaration tested.  i am open to better ideas.
Also, I have only tested with the browsers/versions listed in patch.


chicoxyzzy:
Probably we could use `null` values (will be rendered as gray cell with ? sign) for untested browsers / environments

ljharb:
I would definitely prefer explicit values of `true` (tested, passes), `false` (tested, fails), `null`/`undefined` (untested).

afmenez:
I agree and will do, but shouldn't null be the default instead of false then?

ljharb:
Yes, it should. I'd changed that at one point but got pushback; i'm not sure why.

afmenez:
Opened issue #1123 about that.

afmenez:
Done adding null to untested browsers/transpilers/engines.  It was a lot of work to get it right, and if this is going to be enforced in the future we need to fix #1123 fast.

chicoxyzzy (approved):

chicoxyzzy:
@afmenez sorry for conflicts after merging previous PRs. This one looks good to me

afmenez:
Done rebasing.  Please prioritize merging this PR over my later ones.

ljharb (approved):

PR #1128, opened 2017-07-07
Update Safari TP and WebKit

Constellation:


data-es5.js
```diff
@@ -710,8 +710,6 @@ exports.tests = [
       ie9: true,
       firefox5: true,
       safari10_1: true,
-      safaritp: true,
-      webkit: true,
```
chicoxyzzy, line 714: Let's keep browsers info explicit

chicoxyzzy (changes requested):

data-es5.js
Constellation: OK, I've just updated the patch.

chicoxyzzy:
Thank you!

Issue #1127, opened 2017-07-06
Make tables more compatible with old browsers


afmenez:
I was trying to get results for some old browsers, bur unfortunately they can't handle the current page layout.

IE11: locks up on every tab;
Opera 7: crashes on every tab;
Opera 7.5-10: works on "non-standard", get garbled layout (and missing "results" column) on every other page;

Can't the DOM/js/layout of all tabs be like the "non-standard" tab?

ljharb:
Seems like a bug to me that the pages don't work on all browsers listed in all the tables.

chicoxyzzy:
hmm... I think this is a problem with wrong `colspan` for some columns

afmenez:
I have run an HTML validator on the problematic pages:
https://validator.w3.org/nu/?doc=http%3A%2F%2Fkangax.github.io%2Fcompat-table%2Fesnext%2F
There are a lot of problems with invalid characters ([, ], #, %), which seem serious to me.
And yes, colspan does not match the number of columns on a lot of places, bu I believe this may be a problem of mismatched </tr> tags.

afmenez:
I have run an HTML validator on the problematic pages:
https://validator.w3.org/nu/?doc=http%3A%2F%2Fkangax.github.io%2Fcompat-table%2Fesnext%2F
And yes, colspan does not match the number of columns in a lot of places.

I believe these are the problems affecting the initial rendering:
1. There are a lot of problems with invalid characters ([, ], #, %), which seem serious to me (see #1144);
2. ~~On all tabs, category titles are one column short.~~  PR #1134 has the fix (already merged);
3. On all tabs, the test lines are initially created with 2 columns missing (current browser result and a spacer), which are then added later dynamically.  I believe a better approach would be to create all lines with the correct number of columns, and just fill (or replace) the browser result cell later.  I was able to fix build.js to generate the initial table correctly, but I don't know how to fix master.js to replace fill the cells instead of adding new ones (see #1145);
[...]

Issue #1118, opened 2017-06-23
The handling of features for cascading versions fails when a version is retired


afmenez:
The handling of features for cascading versions is failing when a version is retired.

For example, this is the current configuration for the IE10 browser on environments.json:
 ```
 "ie10": {
    "full": "Internet Explorer",
    "family": "Chakra",
    "short": "IE 10",
    "release": "2012-10-26",
    "retire": "2016-01-12",
    "obsolete": true,
    "test_suites": [
      "es5",
      "es6",
      "es2016plus",
      "esintl",
      "non-standard"
    ]
  },
```
Notice that it is present on the es2016plus tab, but not on the esnext tab.

Now this is a feature on the es2016plus tab:
```
  {
    name: 'assignments allowed in for-in head in non-strict mode',
    spec: 'https://tc39.github.io/ecma262/#sec-initializers-in-forin-statement-heads',
    mdn: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Compatibility_Initializer_expressions_in_strict_mode',
    category: '2017 annex b',
    significance: 'tiny',
    exec: function(){/*
     for (var i = 0 in {}) {}
     return i === 0;
     */},
    res: {
      tr: true,
      ie10: true,
      edge12: true,
      firefox2: true,
      firefox31: true,
      firefox40: false,
      firefox52: true,
      chrome30: true,
      safari3_1: true,
      safari9: false,
      safari10_1: true,
      safaritp: true,
      webkit: true,
      node0_12: true,
      android4_0: true,
      ios5_1: true,
      duktape2_0: true,
    },
  },
```
Notice that it is enabled for ie10 and not ie11, but the generator script takes care of this and cascades the feature for all versions >=10.  Which is great.

But what happens if you remove es2016plus from IE10 in the environments file?  Then IE10 is dropped from the es2016plus tab, as expected.  But also the feature above is marked as false for IE11, which is unexpected (and wrong!).

afmenez:
The same problem happens when a browser is changed to obsolete: very.  It disappears from all tables, and its results are not cascaded to later versions.

afmenez:
Well, I found that in both situations the browser version is dropped too early and its features are indeed not considered for cascading.  A better approach would be to add all browsers to the feature table, and when all the data is complete, remove obsolete and missing versions.

ljharb:
That sounds like the right solution.

afmenez:
[...]

PR #1129, opened 2017-07-08
Reviewing Opera versions and results, retiring ancient versions

afmenez:
I am adding separate entries for all Opera versions between 7.0 and 12.1.
Due to issue #1127 , I was only able to test with versions 10.50 and later, so no results for previous versions.
Also notice that due to issue #1118 , there are a lot of wrong results on the generated html (there were before, and there are more now).  Actually, I am not sure this patch can be merged until the issue is fixed.


data-es5.js
```diff
@@ -67,8 +67,7 @@ exports.tests = [
       webkit: true,
       chrome5: null,
       chrome7: true,
-      opera10_50: null,
-      opera12_10: true,
```
chicoxyzzy: Let's keep things explicit where possible (keep `opera12_10: true`) here and elsewhere

chicoxyzzy (changes requested):

data-es5.js
afmenez: Sorry, do you want me to add "true" to all tested versions?  This would be a lot of work with little to no gain.

chicoxyzzy: No, just those which was already marked as `true`

afmenez:
Reverted the result consolidation.

chicoxyzzy (approved):

afmenez:
Notice that even after restoring previous explicit results, this patch still introduces more errors on the generated HTML due to #1118 (there are some wrong values currently).

afmenez:
As I was rebasing this patch, I noticed that it causes some wrong results but it fixes a lot more.  Even with #1118 still pending, I believe this patch is a gain.

afmenez:
Tks!

Issue #1123, opened 2017-07-05
Change the default result class to "unknown"


afmenez:
Currently, if a test result is not specified for a given browser, it's assumed to be "false" (not supported).  I believe it would be more appropriate to default to "null" (support unknown) instead.

afmenez:
Fixed by #1131

PR #1131, opened 2017-07-11
Change the default result class to "unknown"

afmenez:
Currently, if a test result is not specified for a given browser, it's assumed to be "false" (not supported).  I believe it would be more appropriate to default to "null" (support unknown) instead.
This fixes issue #1123 .


chicoxyzzy:
👍 I'll check this later today

afmenez:
~~This changes the results for a lot of browsers.  I am preparing separate patches explicitly setting "false" to all the browsers I have tested (firefox 2-56, IE11, Edge 14-16, Opera 10-12).~~ This one was merged by itself with #1143

afmenez:
I have added the explicit falses for IE11 on the non-standard table, but I can't check the other tabs due to #1127

afmenez:
The explicit falses for Opera are already on #1129 , I will not repeat the changes here.

afmenez:
~~For some reason, the undefined results for STAGE0 tests on esnext become green with the patch, when they should be gray.  I was not able to identify what is causing this.~~ (Fixed by #1192)

afmenez:
This PR is more than 2 years old.  What is holding it back?

chicoxyzzy:
Someone should check that everything looks fine in this branch. I think no one just did it yet =)

nicolo-ribaudo:
This locally looks fine to me. I only have a comment, not directly caused by this PR but made more considerable.



Should the first row (the header) be gray rather than red?

afmenez:
I agree with you, but I believe it is a separate issue...

nicolo-ribaudo (approved):

ljharb (approved):

chicoxyzzy (approved):
Wow. That's a really barrel aged PR! Thank you for bearing with us @afmenez ❤️

PR #1137, opened 2017-07-25
Promote Chrome 60 to a stable version

mgol:


environments.json
```diff
@@ -1284,14 +1284,14 @@
     "family": "V8",
     "short": "CH&nbsp;59,<br>OP&nbsp;46",
     "note_id": "experimental-flag",
-    "obsolete": false
+    "obsolete": true
```
ljharb, line 1287: Is Chrome 59 really obsolete, just because 60 is available? My Chrome is still on 59, and it took me explicitly checking the version to initiate an update.

mgol: That's been the strategy for all browsers that are on the rolling-release schedule (Edge is an exception as it's not really that rapid). Within a few days most people will get an upgrade and will get it applied when they restart the browser.

Giving up on this approach would just create the need for lots of little arbitrary decisions on when to flip the switch which would happen soon enough anyway.

ljharb: Fair enough.

ljharb (approved):

chicoxyzzy (approved):

Issue #1034, opened 2017-02-22
Lots of features missing from the "esnext" tab


afmenez:
Looking at https://github.com/tc39/proposals, I see a lot of features missing from esnext tab.

Stage4:
- [Dynamic import()](https://github.com/tc39/proposal-dynamic-import) (~FFox ticket [#1342012](https://bugzilla.mozilla.org/show_bug.cgi?id=1342012)~, merged o FFox 66)
- [`for-in` mechanics](https://github.com/tc39/proposal-for-in-exploration) (~FFox ticket [#1557722](https://bugzilla.mozilla.org/show_bug.cgi?id=1557722)~)
- [`import.meta`](https://github.com/tc39/proposal-import-meta) (~FFox ticket [#1427610](https://bugzilla.mozilla.org/show_bug.cgi?id=1427610)~, merged on FFox 62)
- [RegExp Match Indices](https://github.com/tc39/proposal-regexp-match-indices) (~FFox ticket [#1519483](https://bugzilla.mozilla.org/show_bug.cgi?id=1519483)~, merged on FFox 88)
- [Top-level `await`](https://github.com/tc39/proposal-top-level-await)  (~FFox ticket [#1519100](https://bugzilla.mozilla.org/show_bug.cgi?id=1519100)~, merged on FFox 85)
- Stable sort (normative, [arrays](https://github.com/tc39/ecma262/pull/1340), [typed arrays](https://github.com/tc39/ecma262/pull/1433))
- [Ergonomic brand checks for Private Fields](https://github.com/tc39/proposal-private-fields-in-in) (~FFox ticket [#1648090](https://bugzilla.mozilla.org/show_bug.cgi?id=1648090)~, merged on FFox 90)
- [Symbols as WeakMap keys](https://github.com/tc39/proposal-symbols-as-weakmap-keys) (FFox ticket [#1710433](https://bugzilla.mozilla.org/show_bug.cgi?id=1710433))
- [Well-Formed Unicode Strings](https://github.com/tc39/proposal-is-usv-string) (~FFox ticket [#1803523](https://bugzilla.mozilla.org/show_bug.cgi?id=1803523)~, merged on FFox 115)
- [`Atomics.waitAsync`](https://github.com/tc39/proposal-atomics-wait-async)  (FFox ticket [#1467846](https://bugzilla.mozilla.org/show_bug.cgi?id=1467846))
[...]

PR #1140, opened 2017-07-31
Promoting class fields to Stage 3

afmenez:
The "class fields" proposal was promoted to Stage 3 last week.  I believe none of the other changes affect this table because there are no tests for them yet (see #1034).

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1860,7 +1860,7 @@ exports.tests = [
 },
 {
   name: 'class fields',
-  category: STAGE2,
+  category: STAGE3,
   significance: 'medium',
   spec: 'https://github.com/tc39/proposal-class-fields',
   subtests: [
--- a/esnext/index.html
+++ b/esnext/index.html
@@ -1034,6 +1034,270 @@ return regexGreekSymbol.test(&apos;&#x3C0;&apos;);
 <td class="no" data-browser="ios10_3">No</td>
 <td class="no unstable" data-browser="ios11">No</td>
 </tr>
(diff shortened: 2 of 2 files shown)


chicoxyzzy (approved):

ljharb (changes requested):
The shorthand syntax was deferred to a separate proposal, so that test needs to be deleted.

afmenez:
Sorry, can you elaborate?  Do you mean one of the subtests must be split to a separate proposal? Which subtest (I assume it is the last one, "private fields initializers")?  Which separate proposal?

chicoxyzzy:
Does new proposal for shorthand syntax of class fields already exist?

ljharb:
The shorthand proposal does not yet exist; for now we should delete or fix any test that does not have `this.` in front of the #.

afmenez:
So, the only change would be in the last test (private fields initializers)
```
class C {
-   #x = 42;
+   this.#x = 42;
    x() {
        return this.#x;
    }
}
return new C().x() === 42;
```
Is that it?

ljharb:
Yep!

It'd also be good (but not necessary) to also add a test that accesses a private field on another instance of the same class.

afmenez:
I did fix the shorthand use, but I left the additional tests to a separate patch, since it is not the intention of this one.

data-esnext.js
```diff
@@ -1902,7 +1902,7 @@ exports.tests = [
       name: 'private fields initializers',
       exec: function () {/*
         class C {
-          #x = 42;
+          this.#x = 42;
```
ljharb: oh crap sorry, this is my bad :-( the shorthand is `#x` in code - but in the class body, `#x = 42` is already correct (－‸ლ) apologies

ljharb (changes requested):

ljharb:
k, i excised the extra commit, sorry about that

afmenez:
No problem.

PR #1143, opened 2017-08-01
Explicitly set tested Firefox 2...56 results to false where appropriate

afmenez:
About one year ago I tested all versions of Firefox since 2.0, but added only the "true" results because "false" is the default value.  I am now explicitly adding all the "false" results to prepare for the default changing to "null".


ljharb (approved):

chicoxyzzy (approved):

chicoxyzzy:
thank you!

Issue #1148, opened 2017-08-03
Add entries for Number.parseInt/Number.parseFloat to es2015 data


existentialism:
From:
https://github.com/babel/babel-preset-env/issues/385

Ref:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseFloat
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseInt

Will try and get a PR for this asap, but wanted to create an issue in case someone beats me to it!

PR #1151, opened 2017-08-08
Add Number.parseFloat and Number.parseInt

existentialism:
Fixes https://github.com/kangax/compat-table/issues/1148


data-es6.js
```diff
@@ -16210,6 +16210,40 @@ exports.tests = [
...
+        node0_12: true,
+        safari9: true,
+        typescript: typescript.corejs,
+      },
+    },
+    {
+      name: 'Number.parseInt',
+      mdn: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseInt',
+      exec: function () {/*
+        return typeof Number.parseInt === 'function';
```
ljharb: this should also test that `Number.parseInt === global.parseInt`, which the spec [requires](https://tc39.github.io/ecma262/#sec-number.parseint)

data-es6.js
```diff
@@ -16210,6 +16210,40 @@ exports.tests = [
       },
     },
     {
+      name: 'Number.parseFloat',
+      mdn: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseFloat',
+      exec: function () {/*
+        return typeof Number.parseFloat === 'function';
```
ljharb: this should also test that `Number.parseFloat === global.parseFloat`, which the spec [requires](https://tc39.github.io/ecma262/#sec-number.parsefloat)

ljharb (changes requested):
LGTM, just two comments :-)

existentialism:
@ljharb thanks, updated!

ljharb (approved):
Please reconfirm that every engine in this list passes the new tests :-)

chicoxyzzy (changes requested):
`Number.parseFloat === global.parseFloat` and `Number.parseInt === global.parseInt` parts are failing in all browsers. I think we should use `Function('return this')()` or something like that to get global object.

ljharb:
oh, well yeah `global` isn't available by default - I agree with your suggestion (altho i think we could also create a global object variable using `Function('return this')()` in all the skeletons, and just rely on that existing in tests)

zloirock:
I think that something like `Number.parseFloat === parseFloat` is enough.

ljharb:
The extra checks can't hurt; and there's a proposal to change some behavior of these methods, so it'll be important to cover the details.

PR #1154, opened 2017-08-08
Adding duktape versions, results

afmenez:
Adding obsolete versions 1.0...1.8 and unstable version 2.2.
Adding some results for the new versions (based on release notes).


ljharb (approved):

chicoxyzzy:
cc @svaarala

svaarala:
Awesome :)

chicoxyzzy (approved):

afmenez:
Rebased to resolve conflicts.

chicoxyzzy:
thank you!

afmenez:
I'd like to reinforce that all the data on the patch came from the duktape site, I have not run any tests.  It would be nice if someone more knowledgeable could:
- Check/add the retire dates for all versions.  I added the ones that I could deduce, but they may be worng;
- Check the test results.  I added the results from the release notes, but I don't know if the tests can actually run.  Also, I found no info for the ES5, International and non-standard tabs;

svaarala:
The data for 2.0 and 2.1 are from actual test runs, and I'll update the 2.2 results from an actual test run too when that release happens. I think you should be able to run tests for 1.x too, using the same test runner as I've used for 2.0 and 2.1 (it's now committed to compat-table).

PR #1159, opened 2017-08-21
tests for 'optional catch binding' proposal, stage 3

dnalborczyk:
**optional catch binding**, stage 3

supported by Safari TP

https://github.com/tc39/proposal-optional-catch-binding
https://tc39.github.io/proposal-optional-catch-binding/

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -2653,6 +2653,64 @@ exports.tests = [
       }
     },
   ]
+},
+{
+  name: 'optional catch binding',
+  category: STAGE3,
+  significance: 'tiny',
+  spec: 'https://tc39.github.io/proposal-optional-catch-binding/',
+  subtests: [
+    {
+      name: 'basic',
+      exec: function(){/*
+        try {
+          throw new Error();
+        }
+        catch {
+          return true;
+        }
+      */},
+      res: {
+        safaritp: true
+      },
+    },
+    {
+      name: 'await',
+      exec: function(){/*
+        (async function (){
+          try {
+            await Promise.reject();
+          }
+          catch {
+            asyncTestPassed();
+          }
+        })();
+      */},
+      res: {
+        safaritp: true
+      },
+    },
+    {
+      name: 'yield',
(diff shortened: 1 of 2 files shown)


data-esnext.js
```diff
@@ -2653,6 +2653,64 @@ exports.tests = [
...
+  subtests: [
+    {
+      name: 'basic',
+      exec: function(){/*
+        try {
+          throw new Error();
+        }
+        catch {
+          return true;
+        }
```
ljharb, line 2671: should this have an explicit `return false` at the end?

data-esnext.js
```diff
@@ -2653,6 +2653,64 @@ exports.tests = [
...
+            yield;
+          }
+          catch {
+            return true;
+          }
+        }
+
+        var it = foo();
+        it.next();
+        return it.throw().value;
```
ljharb: this seems like it requires proper implementation of `.throw()` - could we test this with `yield; throw new Error();` instead?

data-esnext.js
```diff
@@ -2653,6 +2653,64 @@ exports.tests = [
...
+      exec: function(){/*
+        try {
+          throw new Error();
+        }
+        catch {
+          return true;
+        }
+      */},
+      res: {
+        safaritp: true
```
chicoxyzzy: WebKit Nightly also supports these features

data-esnext.js
dnalborczyk: I assumed an implicit "return undefined" (false) -- as with the other 2 tests -- but I don't have a strong opinion about it, hence I added it.

chicoxyzzy (approved):

chicoxyzzy:
I just merged another PR and it introduced conflicts. Sorry for that :(

dnalborczyk:
no problem. Just pulled upstream, and rebuild the html.

ljharb (approved):

chicoxyzzy:
@dnalborczyk thank you for your contribution!

PR #1163, opened 2017-08-27
declare traceur, phantom.js, and besen as obsolete

dnalborczyk:
I think it's pretty safe to say that we won't expect any additions, changes from Traceur, Phantom.js, and Besen. As a bonus it'll also create more space for the test result table to display.

**Traceur**, last commit Sep 15, 2016
https://github.com/google/traceur-compiler

**Phantom.js** (unmaintained, some sporadic commits in June)
maintainer stepped down with the advent of Headless Chrome.
https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE
https://github.com/ariya/phantomjs

**Besen**, last commit April 23, 2016
https://github.com/BeRo1985/besen

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -12,7 +12,7 @@
     "full": "Traceur",
     "short": "Traceur",
     "platformtype": "compiler",
-    "obsolete": false,
+    "obsolete": true,
     "test_suites": [
       "es6",
       "es2016plus",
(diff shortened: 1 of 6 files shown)


ljharb:
(replaces #1162)

ljharb (changes requested):
"no more maintenance" doesn't necessarily mean it's obsolete; headless chrome can't yet replace phantom, so I'd say phantom is still not obsolete.

I think an ES5 engine is clearly obsolete - but altho Traceur indeed no longer seems to be under active development, I'm not sure if "obsolete" applies yet.

zloirock:
I agree with @ljharb.

afmenez:
The Phantom.js part was already merged with  #1364, the other two need updating.

PR #1168, opened 2017-09-06
Fixing Tail Call Optimization and Promise.prototype.finally support in Google Chrome

### LINKED ISSUE #1166: V8 TCO compatibility data is out of date
Héllo,

I can't reproduce the results regarding TCO even with the harmony flag using current unstable Chrome dubbed 62.0.3198.0.

@chicoxyzzy can you help? Which version of Chrome did you use? Which OS?

afmenez:
Updating the es6 table to reflect the changes in TCO support in Chrome.  This fixes #1166.


chicoxyzzy:
According to http://www.chromium.org/developers/calendar, `chrome.tco` version should be Chrome 57 and `false` version should be Chrome 60

afmenez:
I may be wrong, but the chrome.tco was merged on Feb 11, 2017, which means 57 was branched and 58 was HEAD then.
The feature removal was on Jul 13, 2017, which means 60 was branched and 61 was HEAD.  I'll fix this one.

chicoxyzzy:
@bmeurer sorry for bothering you but how can we check which versions of V8 and/or Chrome support features by bugs in bug tracker? Or maybe there is a better way to check this?

bmeurer:
We have a tool for that (now), but it doesn't seem to be public yet.

image

cc @natorion who's responsible for that tool and who might know more about ETA

afmenez:
Rebased to current HEAD.
I realized that the new chrome.harmony note may replace other "flagged" tags too, but I am not sure which ones.  I have opened issue #1169 about this.

natorion:
@chicoxyzzy what exactly do you have in mind? If you are talking about language features I recommend consulting https://www.chromestatus.com/features. We **should** have an entry for every language feature there with correct release information.

The associated V8 version of a random commit can be found out via v8/tools/release/mergeinfo.py. The tool Benedikt linked is not public (yet) and still under heavy development.

afmenez:
@natorion, the page you mentioned does indeed mention release versions for when the features were enabled by default, but what we are discussing here is when the TCO feature became available behind a specific flag, and when it was removed.  Me and @chicoxyzzy are interpreting the merges dates differently.

chicoxyzzy (approved):

afmenez:
I have added the same note to more tests, to begin addressing #1169.  And I believe this is it for this pull.

afmenez:
Adding the flag for Promise.prototype.finally to address issue #1170 too.

natorion:
@afmenez in that case please use v8/tools/release/mergeinfo.py for now. That should be accurate enough for this purpose.

Issue #1170, opened 2017-09-08
Promise.prototype.finally has been available behind a flag since chrome 61


gsathya:

afmenez:
What is the flag?  I tried `--harmony`, but the feature is not enabled.

gsathya:
> What is the flag? 

`--harmony-promise-finally`

> I tried --harmony, but the feature is not enabled.

I just [staged it](https://github.com/v8/v8/commit/4de946d96ddf2e4f7b033c9b5e763dada8fb5d61). `--harmony` should work in the next canary release. 
EDIT: It's on 63.0.3212.0 canary release (on Mac at least)

afmenez:
Thanks, updated #1168 to address this one too.

afmenez:
Hey @gsathya, are there other flags for specific Harmony features?  I am trying to document them all on #1169, but don't know where to find this information...

gsathya:
> Hey @gsathya, are there other flags for specific Harmony features?

https://cs.chromium.org/chromium/src/v8/src/flag-definitions.h?q=flag-defi&sq=package:chromium&l=192

The more stable, staged features (all available behind `--harmony` flag) are:
https://cs.chromium.org/chromium/src/v8/src/flag-definitions.h?q=flag-defi&sq=package:chromium&l=210 (defined by 

Thinking more about this, maybe we shouldn't mark the individually flagged features on the compat table because they are still in development and may not even work. It might be better to just mark the staged features as 'flagged'.

afmenez:
So you mean that for this feature (Promise.prototype.finally) we should have just "flagged" for Chrome 61 and "--harmony" for Chrome 63?  And remove the notes for the TCO and SIMD flags too?

gsathya:
> So you mean that for this feature (Promise.prototype.finally) we should have just "flagged" for Chrome 61 and "--harmony" for Chrome 63? And remove the notes for the TCO and SIMD flags too?

I only see "flagged" on the compat table. Are you planning on adding a "--harmony" state to the table as well? That might be confusing to users. 

I would suggest --
(a)  "No" for features behind individual `--harmony-*` flags.
(b) "Flag" for features available behind `--harmony` flag.
(c) "Yes" for features available unflagged.

I would change Promise.prototype.finally as "Flag" for Chrome 63 and "No" for Chrome 61, 62.

afmenez:
[...]

Issue #1169, opened 2017-09-08
"Flagged" issues for Chrome should have a note on how to enable them


afmenez:
There are many issues marked as `"flagged"` for Chrome that have no clue on how to enable them.  On pull #1168 I added a new note `chrome.harmony` that may apply to some of these issues, but I was not able to find which ones.  Someone which more Chrome insight could update them.

afmenez:
I have added all the other features enabled by chrome.harmony on Chrome 61 to pull #1168.  What is missing:
- Notes for the `chrome.harmony` flag on versions < 61;
- Other flags that enable the remaining issues;

afmenez:
According to @gsathya, only the `--harmony` flag enables complete and stable features (see #1170).  The other flags enable features that are incomplete or unstable.
My plan now is to remove the notes for these flags, keeping only the one for `--harmony`, and changing the associated features to `false` instead.  Any opposition?  @chicoxyzzy? @ljharb?

chicoxyzzy:
compat-table shows not only stable features. Actually all features from esnext tab are unstable. We should indicate if some ebvironments support those features (no matter under distinct flag or not)

afmenez:
We are talking about different kinds of unstable.  An unstable feature means it is not complete yet and may change.  An unstable implementation means it does not follow the specification or that it may crash the browser.
If I understand correctly, what @gsathya is saying on #1170 is that when a feature is behind a specific flag, it is because it is not considered ready for testing.  As soon as it is considered ready to test, it is moved to the `--harmony` flag.

chicoxyzzy:
I still don't agree with you.

> An unstable feature means it is not complete yet and may change.

All esnext features are unstable in that meaning.

> An unstable implementation means it does not follow the specification or that it may crash the browser.

esnext features are not in specification

> it is because it is not considered ready for testing

If it is implemented and shipped than we should (and we actually do) test them

afmenez:
>
> An unstable feature means it is not complete yet and may change.
>
> All esnext features are unstable in that meaning.
>
Yep.

> An unstable implementation means it does not follow the specification or
> that it may crash the browser.
>
> esnext features are not in specification
>
My bad, they have a proposal.

> it is because it is not considered ready for testing
>
> If it is implemented and shipped than we should (and we actually do) test
> them
>
My patch keeps the note on how to enable them, just removes the "flagged"
marking.

chicoxyzzy:
[...]

PR #1173, opened 2017-09-11
Change unstable Chrome features to false instead of "flagged"

afmenez:
Addressing the issues on #1169.
If I understand correctly, what @gsathya is saying on #1170 is that when a feature is behind a specific flag, it is because it is not considered ready for testing. As soon as it is considered ready to test, it is moved to the `--harmony` flag.


ljharb (approved):

chicoxyzzy (approved):

gsathya:
Perfect. Thanks @afmenez!

Issue #1176, opened 2017-09-15
Move the `Array.prototype.sort` `comparefn` test away from /es5/


mathiasbynens:
https://kangax.github.io/compat-table/es5/#test-Array_methods_Array.prototype.sort:_compareFn_must_be_function_or_undefined

This test has nothing to do with ES5. It’s a formerly undefined behavior that only [very recently](https://github.com/tc39/ecma262/commit/564b30b3b1bbba3772a2ebeb58208043713d64fe) (years after ES5 was published) got standardized. If anything, it’s an ES2018 feature.

Should this be moved from `…/es5/` to another page (e.g. `…/es2018/`)?

ljharb:
It certainly is part of ES2018, technically - but it's been a defacto standard since ES5, which is why https://github.com/tc39/ecma262/pull/785 was able to be a PR, and not a proposal. 

I think it should remain in ES5.

mathiasbynens:
What makes you say that it’s been a de facto standard since **ES5** as opposed to ES3/ES2/ES1?

mathiasbynens:
Oh, and this repo has a compat table specifically for de facto standards: https://kangax.github.io/compat-table/non-standard/ This “feature” either belongs in `…/non-standard/` or in `…/es2018/`. IMHO the latter makes the most sense at this point.

ljharb:
Perhaps it's been one for longer, but we don't have an older compat table :-)

It's worth noting that https://github.com/tc39/ecma262/pull/785 didn't actually standardize this behavior, but rather made it explicit - it was the original intention of the editor of ES5, it was implied by the ES5 text, and everyone implemented it that way - in other words, the PR didn't *newly standardize* it, it merely cleared up the spec text to reflect both the ES5 intention and the majority ES5 implementation reality. It's thoroughly an ES5 test.

mathiasbynens:
> everyone implemented it that way

That’s not true, as you know :) I don’t think it’s fair to blame an engine for not implementing something that was never part of a normative spec before.

ljharb:
lol fair, mostly everyone then :-p

A failing test isn't "blame" imo, and we shouldn't be making changes in the table so a browser can "make everything green".

In this case, it was always part of a normative spec - it was just easy to misinterpret. If an engine failed to interpret it correctly, failed to compare to other engines and ask why the interpretations were different, and also failed to ask TC39 for clarification, I'm not sure why their failing test should be swept under the rug.

[...]

PR #1179, opened 2017-09-16
Update `Array#sort` `comparefn` data for Chrome 63

mathiasbynens:
Ref. https://chromium.googlesource.com/v8/v8.git/+/a10e4a179e8a71b9e2c5470d3f50ec325cc1576e%5E%21/

(Note that this is not actually an ES5 test — see #1176.)

### DIFF
--- a/data-es5.js
+++ b/data-es5.js
@@ -743,6 +743,8 @@ exports.tests = [
       safari10_1: true,
       safaritp: false,
       webkit: false,
+      chrome1: false,
+      chrome63: true,
       opera10_10: null,
       opera10_50: true,
       konq4_3: null,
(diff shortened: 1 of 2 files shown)


afmenez (approved):
LGTM

data-es5.js
```diff
@@ -743,6 +743,7 @@ exports.tests = [
       safari10_1: true,
       safaritp: false,
       webkit: false,
+      chrome63: true,
```
afmenez: You could add "chrome1: false" too.

mathiasbynens: Done.

ljharb (approved):

Issue #1175, opened 2017-09-13
Feature marked incorrectly for Safari 10


ssilve1989:
`babel-preset-env`, which uses this table, ends up incorrectly excluding a plugin `transform-es2015-block-scoping` when targeting support of Safari 10.

I believe this is due to the features `const` and `let` under bindings being marked completed but there is a bug in Safari 10, [shown here](https://bugs.webkit.org/show_bug.cgi?id=159270) where the following 
```javascript
{
  let a = 3
}

let a = 3
```
throws `SyntaxError: Can't create duplicate variable that shadows a global property: 'a'`

References babel/babel-preset-env#411

afmenez:
@chicoxyzzy, by reading the description I thought it was about writing two new subtests (for `let` and `const`), but it is now marked as "test code bug".  Do you mean the existing tests should be changed to fail on the Safari buggy implementation?

ljharb:
Separate tests are good, but yes, absolutely a test should always fail on a buggy implementation.

afmenez:
I was not able to cram this issue on the existing subtests, so I added new ones.

ssilve1989:
Just upgraded to Safari Version 11.0 (12604.1.38.1.7) today and safari **still** throws this error with that code snippet :rage:

afmenez:
Yep, my patch #1183 is not merged yet.

afmenez:
I believe one of the reasons for not merging is that I haven't actually run the new tests on Safari.  Could you do that and comment on the PR if it actually solves your issue?

ssilve1989:
So I ran the build and opened up Safari to view it. In the Safari 11 column, it lists `No` for `scope-shadow-resolution` and `scope-shadow-resolution (strict mode)`. However in the `Current Browser` column they show as passing. The current browser, that i was running it on was `Safari Version 11.0 (12604.1.38.1.7)`, and indeed it **does** work on that version of 11. I'm not sure why it didn't 13 days ago when I first upgraded Safari, its the same exact version as then

afmenez:
Yes, the result columns are hard-coded, only the "current browser" column is live.  I marked it `no` for Safari because you said it should not work.  I can change it to `yes`, but some questions remain:
- How about Safari 10.1 and earlier?
- If they work too, is the test incorrect?

ssilve1989:
[...]

PR #1183, opened 2017-09-22
Add new subtests for scope shadow resolution

afmenez:
This tries to address issue #1175.

Possible issues with this patch:
- I don't have a Mac to test, so I am not sure it actually fails on Safari. Can anyone test this?
- I have only tested on current Firefox, IE, Edge and Chrome, the other results are assumed to be the same as the block-scope tests.  Should they be changed to "null"?


chicoxyzzy:
I'll try to check this PR tomorrow

afmenez:
I am not sure this is needed anymore.  The original issue was about a problem on Safari 10 to 11, but now it seems the issue does not exist on Safari 10.1 and 11.  So it would only affect Safari 10, but we were not able to test it with that version.
I see 2 alternatives here:
- Change the Safari 10 result to `null`;
- We find someone that can test this on Safari 10.  If it passes, we can drop this patch, if it doesn't we add the `false` result;

ljharb:
Let's wait on the latter option.

afmenez:
Again I can't find what is the problem here.  I looked at the logs but could not find the cause of the error.  I have run `npm test` locally, and this is the output:
```
> compat-table@0.0.0 pretest C:\Users\alexandre.menezes\Documents\Kangax\compat-table
> npm run lint

> compat-table@0.0.0 lint C:\Users\alexandre.menezes\Documents\Kangax\compat-table
> jshint .

> compat-table@0.0.0 test C:\Users\alexandre.menezes\Documents\Kangax\compat-table
> echo 'Error: no test specified'

'Error: no test specified'
```
Any help will be appreciated.

chicoxyzzy (approved):
I had no chance to check results in Safari 10. I think we can merge this as it is after rebase

afmenez:
Rebased.

PR #1184, opened 2017-09-22
Fixing mismatched result for 'Array.prototype.sort' subtest

afmenez:
The current results don't make sense.
- I don't have a Mac, so I haven't actually tested;
- Shouldn't all the safaritp/webkit identical results be removed when they cascade from safari/safaritp?


data-es5.js
```diff
@@ -740,9 +740,8 @@ exports.tests = [
       ie9: true,
       firefox2: false,
       firefox5: true,
+      safari1: false,
       safari10_1: true,
-      safaritp: false,
```
ljharb, line 744: Def leave these; webkit, safari, and safaritp don't necessarily have the same features.

ljharb (changes requested):

data-es5.js
afmenez: Well, according to environments.json, they all belong in the same "family", so the results cascade.  But I am adding them back.

ljharb (approved):

data-es5.js
ljharb: Right but anything explicitly tested should never be removed - cascading is a sane default, but explicit results are never redundant.

afmenez: Well, as I stated in the first comment I haven't explicitly tested. :D

PR #1185, opened 2017-09-22
Updating Safari TP and Webkit release info and results

afmenez:
Updating Safari TP and Webkit release info and results according to the release announcement:
https://webkit.org/blog/7922/release-notes-for-safari-technology-preview-40/

I haven't run any tests.


afmenez:
Updating Safari TP and Webkit release info again, according to the release announcement:
https://webkit.org/blog/7989/release-notes-for-safari-technology-preview-41/

There are no JS changes on this release.  I haven't run any tests.

ljharb (approved):

data-esnext.js
```diff
@@ -1719,6 +1719,8 @@ exports.tests = [
     ie11: false,
     firefox2: false,
     chrome61: chrome.harmony,
+    safari1: false,
```
afmenez: actually, this was supposed to be safari11, but i guess it is correct anyway.

ljharb: might as well update it :-)

afmenez: Actually, i'll remove this line, since I haven't actually tested it.

afmenez: Hum, I found a lot of safari1 results that I believe I added previously.  I'll remove them too.

afmenez: Should I remove previous safaritp and webkit results too?  I mean, the results refer to a previous version (not TP41).  I would obviously only remove when the results are the same as Safari 11.

ljharb: I'd say we should be keeping only results that were explicitly tested.

afmenez:
I don't understand why PR #1193, which updates the results without updating the header, was merged and this one was not.  Anyway, I rebased the the patch and now no results are changed, just the header.  i still believe it needs merging.

chicoxyzzy (approved):

chicoxyzzy:
Sorry. Merged.

afmenez:
Tks!

PR #1186, opened 2017-09-22
Update results for Duktape 2.2

svaarala:
- Add Duktape 2.2 engine.
- Update results for Duktape 2.2.0 release.
- Change `duktape2_0: false` to `duktape_1.0: false` now that Duktape 1.x engine versions have been added; previously results started from Duktape 2.0.0.
- Remove a few redundancies, e.g. the 2.0 line for `duktape1_3: true` and  `duktape_2.0: true`.

### DIFF
--- a/duktape.js
+++ b/duktape.js
@@ -26,6 +26,9 @@ var dukKey = (function () {
     });
     var dukVersion = Number(stdout);
     console.log('Duktape version is: ' + dukVersion);
+    if ((dukVersion % 100) == 99) {
+        dukVersion++;  // Treat e.g. 2.2.99 (built from master) as 2.3.0 for testing
+    }
     return 'duktape' + (Math.floor(dukVersion / 10000)) + '_' + (Math.floor(dukVersion / 100 % 100));
 })();
 console.log('Duktape result key is: test.res.' + dukKey);
--- a/environments.json
+++ b/environments.json
@@ -1993,6 +1993,7 @@
     "short": "DUK 1.1",
     "platformtype": "engine",
     "release": "2015-01-09",
+    "retire": "2015-04-05",
     "obsolete": "very",
     "test_suites": [
       "es5",
@@ -2005,6 +2006,7 @@
     "short": "DUK 1.2",
     "platformtype": "engine",
     "release": "2015-04-05",
+    "retire": "2016-04-19",
     "obsolete": "very",
     "test_suites": [
       "es5",
@@ -2017,6 +2019,7 @@
     "short": "DUK 1.3",
     "platformtype": "engine",
     "release": "2015-09-12",
+    "retire": "2016-04-17",
     "obsolete": "very",
     "test_suites": [
       "es5",
@@ -2029,6 +2032,7 @@
     "short": "DUK 1.4",
     "platformtype": "engine",
     "release": "2016-01-10",
+    "retire": "2016-08-29",
     "obsolete": "very",
     "test_suites": [
       "es5",
@@ -2091,8 +2095,8 @@
     "family": "Duktape",
     "short": "DUK 2.0",
     "platformtype": "engine",
-    "release": "2017-01-17",
-    "retire": "2017-04-15",
+    "release": "2017-01-02",
+    "retire": "2017-05-04",
     "obsolete": true
   },
   "duktape2_1": {
@@ -2101,14 +2105,16 @@
     "short": "DUK 2.1",
     "platformtype": "engine",
     "release": "2017-04-15",
-    "obsolete": false
+    "retire": "2017-09-23",
+    "obsolete": true
   },
   "duktape2_2": {
     "full": "Duktape 2.2",
     "family": "Duktape",
     "short": "DUK 2.2",
     "platformtype": "engine",
-    "unstable": true
+    "release": "2017-09-23",
+    "obsolete": false
   },
   "android1_5": {
     "full": "Android Browser 1.5 (Cupcake)",
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -1435,6 +1435,7 @@ exports.tests = [
           safari9: true,
           android4_0: true,
           duktape2_0: false,
+          duktape2_2: true,
         }
       },
       {
@@ -1458,6 +1459,7 @@ exports.tests = [
           safari5_1: true,
           safari9: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1484,6 +1486,7 @@ exports.tests = [
           android4_0: true,
           ios5_1: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1509,6 +1512,7 @@ exports.tests = [
           safari9: true,
           android4_0: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1532,6 +1536,7 @@ exports.tests = [
           safari5_1: true,
           safari9: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1560,6 +1565,7 @@ exports.tests = [
           android4_0: true,
           ios5_1: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1587,6 +1593,7 @@ exports.tests = [
           android4_0: true,
           ios5_1: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1614,6 +1621,7 @@ exports.tests = [
           safari9: true,
           android4_0: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1637,6 +1645,7 @@ exports.tests = [
           safari5_1: true,
           safari9: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1659,6 +1668,7 @@ exports.tests = [
           safari9: true,
           ios5_1: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1687,6 +1697,7 @@ exports.tests = [
           android4_0: true,
           ios5_1: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1714,6 +1725,7 @@ exports.tests = [
           android4_0: true,
           ios5_1: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1741,6 +1753,7 @@ exports.tests = [
           safari9: true,
           android4_0: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1764,6 +1777,7 @@ exports.tests = [
           safari5_1: true,
           safari9: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       },
       {
@@ -1786,6 +1800,7 @@ exports.tests = [
           safari9: true,
           ios5_1: true,
           duktape2_0: false,
+          duktape2_2: true,
         },
       }
     ]
--- a/data-es6.js
+++ b/data-es6.js
@@ -411,7 +411,7 @@ exports.tests = [
         xs6: true,
         safari10: true,
         duktape2_0: false,
-        duktape2_2: true,
+        duktape2_2: false,
       },
     },
   ],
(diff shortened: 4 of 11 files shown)


environments.json
```diff
@@ -2008,7 +2008,7 @@
     "family": "Duktape",
     "short": "DUK 2.2",
     "platformtype": "engine",
-    "unstable": true
+    "obsolete": false
```
afmenez: You could add the release date here.  Actually, I guessed the "retire date" (end of support) for all the previous versions, you could check them and add the retire date for 2.1 too. ;)

data-esintl.js
```diff
@@ -22,7 +22,7 @@ exports.tests = [
         safari10: true,
         opera10_50: false,
         node0_12: true,
-        duktape2_0: false,
+        duktape1_0: false,
```
afmenez: I did some "result optimization" myself in the beginning, but it is not the way it is supposed to work.  You should keep here the results for the versions that were actually tested, and let the compiler guess the others.  So you are not supposed to remove correct results, only add new ones.

environments.json
svaarala: Hmm, 2.1 will still get at least one round of maintenance fixes so I guess it's not retired yet?

data-esintl.js
svaarala: I've received a little contradicting feedback on this: previously when I added results for 2.1 that were the same as for 2.0 -- both tested explicitly -- I got feedback that the 2.1 results shouldn't be added if they are not different from 2.0.

afmenez: Interesting.  Maybe it's a major/minor thing then?  Let's hear from the committers then.

environments.json
afmenez: Sure.  But the release data from 2.2 is still missing. :)

svaarala: No disagreement there, I'll add it.

afmenez: Actually, your comment on 2.1 still having one round yet makes me doubt of the retire dates I guessed for all other versions.  Can you update or remove them?

data-esintl.js
svaarala: Also the results for at least Duktape 2.2 (and possibly Duktape 1.3 and 1.4) were added (by someone else) without actually running tests. So some of the 2.2 results were incorrect before the pull. I'm not sure if the 1.x data is correct either (i.e. whether it has actually been tested or is just based on release notes or something like that).

svaarala: @afmenez It seems that the previous pull with 2.2 results was actually from you - can you confirm whether the 1.x results are based on actual test runs or release notes only? If the latter, I could try to run the tests with 1.x to make the results exact?

svaarala:
Regarding the "retire" dates, I'm not sure what the best (desired) semantics are, as it probably depends on the release policy of each project. But for Duktape releases one clear approach is:
- If the release version has maintenance releases beyond the next minor release, the retire date is the last such maintenance release.
- If the last maintenance release was before the next minor release, the retire date is that next release. In effect the new release replaced the previous one. In between the last maintenance release and the next minor release the last maintenance release was still current (and would have been fixed for any serious bugs).

svaarala:
Okay, made the following changes:
- Updated release/retire dates as described above, fixing some 1.x data and adding a release date for 2.2.
- Dropped the previous duktape1_0 data changes, and added only duktape2_2 results for any tests whose result has changed since duktape2_0 or duktape2_1.
- However, the result from duktape2_2 is not added if it hasn't changed since duktape2_1 or duktape2_0; this is consistent with how results for Duktape 2.1 were added (= only changes were recorded).

Squashed and pushed.

data-esintl.js
afmenez: Yes, all the data I added was based on the release notes.  You can sure run the tests if you can and correct everything that is wrong!

svaarala: Ok, I'll do that in a separate pull later on if these results are interesting for old versions :)

afmenez (approved):
Nice!

data-esintl.js
afmenez: Actually it would be great if you could add the actual results for 1.x as you did before.  There's a separate PR #1131 that changes the default from "false" to "unknown", and it would affect most of the results for 1.x if they are not explicitly set.

svaarala: Hmm, so the most important thing would be to get 1.0 results for the full test set, because those are inherited forwards automatically.

afmenez: Exactly.

duktape.js
```diff
@@ -26,6 +26,9 @@ var dukKey = (function () {
     });
     var dukVersion = Number(stdout);
     console.log('Duktape version is: ' + dukVersion);
+    if ((dukVersion % 100) == 99) {
+        dukVersion++;  // Treat e.g. 2.2.99 (built from master) as 2.3.0 for testing
```
ljharb, line 30: Can you elaborate on why this is in here, or why `2.2.99` would be anything more special than the patch version after `2.2.98`?

svaarala: Right after 2.2.0 has been released, Duktape master switches to 2.2.99 which is used until the 2.3.0 release (http://duktape.org/guide.html#versioning.4). While no version is correct for a master build, the .99 patch is used because the build is closer to 2.3.0 than any 2.2.x maintenance version for comparison purposes.


When one runs compat-table/duktape.js with the 2.2.99 release, prior to this change it will appear as Duktape 2.2 and compare against `duktape2_2` results. This is a bit useless because there's no good way to fix the data entries; one would need to alter 2.2 results.

With this change it will compare against `duktape2_3` results and allow one (in practice me) to maintain the to-be-released 2.3.0 results incrementally.

This is not an important change, but before this, I had to manually force dukVersion in duktape.js for testing and maintaining the results incrementally.

ljharb: In that case, duktape is not following semver; it should be using prerelease builds for that. `.100` is no more special than `.1` or `.1000`, and misusing `99` like that means semver-minor changes appear in the 2.2 line after .0.

We'll have to alter the entire way we handle duktape versions inside compat-table unless it changes this policy.

svaarala: Note that builds from master are not releases at all, so Duktape *releases* do follow semver but master builds are not releases.

svaarala: A bit more background, DUK_VERSION is a number of the form 10203 (for 1.2.3) and between releases I guess there are roughly the following options (assuming master build after 2.2.0 but before 2.3.0):
- 2.2.99: current policy, patch level never reaches .99. It's possible to detect an unofficial build from the version number, and 'version >= 2.3.0' check won't catch the unofficial build which is good.
- 2.3.0: this would also be workable, but official release would have to be 2.3.1 to avoid confusion. But this would also break semver: 2.3.0 -> 2.3.1 might entail arbitrary changes (not just patch level changes), e.g. if the 2.3.0 was built right after 2.2.0.
- Alternate releases, e.g. 2.2.0 official release, 2.3.0 reserved for development, 2.4.0 official release, etc. This would be closest to fulfilling semver, but even with this 2.3.x patch level guarantees would be broken because development between minor versions often involves multiple individual changes deserving a minor version bump.

For a numeric version like 1.2.3, with no space for `-alpha` or similar qualifiers, it's not really possible to fulfil semver as far as I can see it, unless one bumps the version literally for every commit going into (unreleased) master, which is very inconvenient. So, that's why the current compromise is to use strict semver for releases only, not manual builds from Duktape master.

ljharb: No, it wouldn't - you'd use "prereleases", which are part of the semver spec and defined for exactly this purpose. You'd do `2.3.0-alpha.0` or `2.3.0-rc.1` etc, and then `2.3.0` when ready.

Using `.99` makes no sense whatsoever.

svaarala: Given that DUK_VERSION is a number of the form major * 10000 + minor * 100 + patch, there's no way to add a suffix like that. The reason it is a number is to allow C programmers to do things like:

```c
#if (DUK_VERSION >= 20300)
/* Use Duktape 2.3 features */
#else
/* Older */
#endif
```

C preprocessor cannot do version comparisons from strings, which is why version number defines are common for C libraries.

ljharb: Since that's the case, that means duktape does not follow semver, and the versioning logic and automated scripts for it in this repo were allowed based primarily on following semver.

svaarala: Does semver require explicit versioning for every *unreleased* commit in a repository?

Because that's what we're talking about here: there are no releases on duktape.org with a .99 patch level, and there won't be any releases like that.

ljharb: Fair point; in that case, let's find a better hack than a ".99" version; or at least let's not let it pollute this repo.

svaarala: Like I explained, that "hack" is the best compromise in practice. Something must go into DUK_VERSION, and 20299 is better than 20300, at least based on user feedback.

But I can remove that check if it makes things better. It just means running duktape.js for a development build without manual tweaks gives nonsensical results. I fail to see why that's better.

ljharb: Personally I find that reserving the `.0` version for this, and only ever releasing `.1` and higher, would be a much cleaner approach than breaking semver just to accomodate deficiencies in C.

svaarala: But that would also break semver: the changes from 2.3.0 -> 2.3.1 would in many cases be more significant than patch level changes?

To clarify, right after 2.2.0 release master would switch to 2.3.0. If you built Duktape right at that time, it would behave like 2.2.0 but have a version number of 2.3.0. In this case the change from 2.3.0 -> 2.3.1 would be the same as 2.2.0 -> 2.3.1.

ljharb: hm, also a fair point. not really sure what to do here.

svaarala: One option is to detect the .99 patch level and refuse to run, indicating the cause: you're running a development version which exists outside the semver version number space and it's not obvious which results you'd want to compare against.

duktape.js could then provide an option to force the version to be used for results purposes. I could then run 2.2.99 as `nodejs duktape.js --force-duk-version 20300` to update the 2.3 results incrementally. This would avoid the need to manually hack the source file.

But I'm fine with whatever really -- I can keep on editing duktape.js manually every time I run the tests, and I doubt many others will run a Duktape development build against the test set.

chicoxyzzy (approved):
LGTM

ljharb (dismissed):
I'm still very displeased with duktape not following semver - that it's doing it due to flaws in C++ versioning doesn't excuse it.

svaarala:
> let's file a new issue to address how we can deal with duktape not following semver, moving forward.

I quite disagree with this statement as discussed above.

chicoxyzzy:
TypeScript and all browsers don't use semver for versioning. Why should it be a blocker?

ljharb:
@chicoxyzzy because duktape has special handling inside compat-table based on the assumption that it uses semver (it's also the only engine that has tools to add automated results)

svaarala:
@ljharb What special handling are you referring to?

All the Duktape versions listed in environments (i.e., are releases) do follow semver. The .99 versions are only used for builds from master, i.e. non-releases.

ljharb:
@svaarala the existence of https://github.com/kangax/compat-table/blob/7db11bd87d2a51e7562a869d3533af01b00d5f4c/duktape.js combined with https://github.com/kangax/compat-table/blob/gh-pages/duktape.js#L43 / https://github.com/kangax/compat-table/blob/gh-pages/duktape.js#L29 - no other engine has special treatment in this way (although many engines have - perhaps problematic - automatic result inheritance based on "not semver")

svaarala:
That check makes no semver assumptions as such - it just interprets environments.json as enumerating the Duktape versions in increasing order of "newness".

Results are inherited based on that order as far as I can tell, and as far as I understood the same inheritance approach is used for all engines? IOW, if no result is provided, it is assumed to be the same for a new version described later in environments.json regardless of any versioning conventions?

But, I might have understood the process incorrectly.

svaarala:
Also just to be clear, duktape.js does not update the results automatically. When it is executed, it just points out what results don't match the values in the data-* files. The changes to the data files are then made manually just like for other engines (again, as far as I understood).

In this process duktape.js "inherits" results based only on the order of Duktape versions listed in environments.json. There is no inference based on the version number itself, or any intepretation of the version number components. This is done here: https://github.com/kangax/compat-table/blob/gh-pages/duktape.js#L33-L49.

ljharb:
You make fair points. Either way, this PR merely needs rebasing to be mergeable.

svaarala:
Rebased and pushed.

chicoxyzzy (approved):

PR #1190, opened 2017-10-02
Reorganizing node.js versions

afmenez:
- Added missing versions of io.js;
- Added release and retire dates for every version;
- Added reference to base V8 (chrome) engine where missing;
- io.js test data was mixed for all three versions, I removed all the data and let it inherit the V8 data;
- Obsoleted ancient versions;
- ~~Node.js 8.x is currently considered unstable;~~

All of this is based on data from:
[Node.js releases](https://nodejs.org/en/download/releases/)
[Node.js on Wikipedia](https://en.wikipedia.org/wiki/Node.js)


environments.json
```diff
@@ -1874,8 +1961,10 @@
     "platformtype": "engine",
     "note_id": "harmony-flag",
     "equals": "chrome60",
-    "obsolete": false,
+    "release": "2017-08-08",
+    "unstable": true,
```
ljharb: Node 8 is not "unstable"; every release is stable in semver, always. It's just not LTS. That's why they call it "current".

ljharb (changes requested):

data-es2016plus.js
```diff
@@ -1662,7 +1658,6 @@ exports.tests = [
           opera10_50: true,
           chrome30: true,
           node0_12: true,
-          iojs: true,
```
ljharb, line 1665: `iojs: true` here indicates it was explicitly tested; it's better to have explicit tests preserved and not just inherit the data.

ljharb (changes requested):

afmenez:
I agree, but it does not which specific version was tested. I believe it is
better to remove than guess. The resulting html is exactly the same
(assuming 3.0 results before)

ljharb (approved):
Fair point :-)

data-es2016plus.js
afmenez: Fixed

chicoxyzzy (approved):

PR #1192, opened 2017-10-04
Fix "unknown" results in esnext Stage0 with green background

afmenez:
This fixes a css problem that was painting green all non-false results with "not-applicable" class.  This was affecting "unknown" results, which should not be green.  There are no such results currently, but they will appear if #1131 gets merged.


ljharb (approved):

chicoxyzzy (approved):

chicoxyzzy:
Thank you!

PR #1196, opened 2017-10-12
Add Node 8.7 results

mgol:
Node 8.7 updated V8 from 6.0 to 6.1.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -1960,11 +1960,28 @@
   "node8_3": {
     "full": "Node.js",
     "family": "Node.js",
-    "short": "Node &gt;=8.3 &lt;9",
+    "short": "Node &gt;=8.3 &lt;8.7",
     "platformtype": "engine",
     "note_id": "harmony-flag",
     "equals": "chrome60",
     "release": "2017-08-08",
+    "obsolete": true,
+    "test_suites": [
+      "es5",
+      "es6",
+      "es2016plus",
+      "esnext",
+      "esintl"
+    ]
+  },
+  "node8_7": {
+    "full": "Node.js",
+    "family": "Node.js",
+    "short": "Node &gt;=8.7 &lt;9",
+    "platformtype": "engine",
+    "note_id": "harmony-flag",
+    "equals": "chrome61",
+    "release": "2017-10-11",
     "unstable": false,
     "test_suites": [
       "es5",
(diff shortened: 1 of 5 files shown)


ljharb:
This seems fine, but it doesn't appear to have altered the results at all - iow, node 8.6 and 8.7 appear to have identical results. Is this the case?

mgol:
The site currently shows that Chrome 61 has some features flagged that Chrome 60 doesn't although I've just checked a few of them in Node 8.6 and they still passed with `--harmony` so maybe these are just missing pieces of data.

ljharb:
Meaning the results for node 8.6 or chrome 60 are incomplete?

mgol:
Node 8.6 & Chrome 60 should have identical results; I've checked only on Node 8.6. I suppose both are incomplete with regards to what's available under a flag.

afmenez (approved):
LGTM

afmenez:
I have just re-run the esnext tests with Chrome 61 and the results are correct with and without the --harmony flag.  Where did you see the inconsistency?

afmenez:
Ah, I see what you mean now.  You are seeing both "RegExp named capture groups" and "s (dotAll) flag for regular expressions" passing in Chrome 60 with the --harmony flag?  It is easy to write the patch, but I have no access to Chrome 60 to actually run the tests.

mgol:
> It is easy to write the patch, but I have no access to Chrome 60 to actually run the tests.

I don't have one either. It's easy to run tests on older Nodes, especially when you use [nvm](https://github.com/creationix/nvm) but not so much with Chrome. Firefox makes all past releases available on FTP, I'm not aware of such official hosting for Chrome binaries.

I do have a BrowserStack account but it runs the browser without the `--harmony` flag.

mgol:
Now that I think about it, perhaps I can navigate to chrome://flags, enable experimental JS features & restart the browser; that has a chance of working.

Another question is how far in the past those flags were available.

afmenez:
I submitted the patch to chrome 60 already (#1197), but if you find when they where introduced I'll updated it.  I would guess Chrome 59, which implemented "RegExp Unicode Property Escapes", but you would have to test 58 too to be sure. :)

mgol:
PR rebased. Still ready to land!

ljharb (approved):

Issue #1146, opened 2017-08-01
The colspan for the "platform type" header cells is initially wrong


afmenez:
The colspan for the "platform type" header cells is initially wrong. It is corrected dynamically later by `setColSpans()` on master.js, but since this is the first row, it defines the initial width for the entire table and I believe older browsers can't handle this (see #1127).

One solution is instead of hard-coding the platform headers on skeleton.html, adding them dynamically like the browser headers in `dataToHtml()` on build.js.

Issue #1198, opened 2017-10-17
Problem with the "platform type" header for empty categories


afmenez:
The colspan for the "platform type" header cells is calculated dynamically by setColSpans() on master.js, but there is a problem when the category is empty.  On the non-standard tab, the "Compilers/polyfills" column is empty and is (correctly) calculated with colspan 0, but the browsers always show at least one column.

PR #1199, opened 2017-10-17
Temporary workaround for "platformtype" column widths problem

afmenez:
This is a temporary workaround for the "platform type" column widths problem (#1146), by harcoding the correct sizes.

### DIFF
--- a/es2016plus/index.html
+++ b/es2016plus/index.html
@@ -105,11 +105,11 @@
 
           <th colspan="5" class="platformtype" id="compiler-header" style="background: #ffdfa4">Compilers/polyfills</th>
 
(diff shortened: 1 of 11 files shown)


chicoxyzzy (approved):

ljharb (changes requested):
I don't think a temporary workaround is beneficial; this will just create unnecessary diff churn in the next PR, until we fix #1146.

afmenez:
Adding a temporary workaround for issue #1198, by removing the empty column header.

afmenez:
I still believe that this patch would be beneficial.  Yes, the pre-filled column witdhs will degrade soon, but the fix to the zero-width header will last.

ljharb (approved):

afmenez:
Tks!

PR #1200, opened 2017-10-18
Edge 16 release

chicoxyzzy:

PR #1201, opened 2017-10-18
Chrome 62 release

chicoxyzzy:


environments.json
```diff
@@ -1496,15 +1496,17 @@
     "family": "V8",
     "short": "CH&nbsp;61,<br>OP&nbsp;48",
     "release": "2017-09-05",
+    "retire": "2017-10-17",
     "note_id": "experimental-flag",
-    "obsolete": false
+    "obsolete": true
```
ljharb: I'm really not comfortable with v61 becoming "obsolete" merely because v62 came out *today*. I'd expect "obsolete" to be, perhaps, 2 versions back at least.

mgol: That would cause more Chrome versions to be shown in the table, wouldn't it?

If anything gets changed in that regard, this should apply to Firefox as well.

environments.json
```diff
@@ -1513,6 +1515,13 @@
     "note_id": "experimental-flag",
     "unstable": true
   },
+  "chrome64": {
+    "full": "Chrome Canary",
```
mgol, line 1519: If you're adding this entry then Chrome 63 should be marked as `Chrome Dev` and Chrome 62 as `Chrome Beta`. In its current state the PR makes two different Canary versions to co-exist which is confusing.

mgol (changes requested):

environments.json
ljharb: Yes, I'd definitely expect, for all browsers, that "a new version was released today" is not enough to make "yesterday's newest version" obsolete.

chicoxyzzy: OK. This sounds reasonable to me. Let's make Firefox changes in another PR

environments.json
chicoxyzzy: Chrome 62 is stable, not beta

chicoxyzzy:
IDK why Travis build failed  ¯\\\_(ツ)\_/¯

environments.json
chicoxyzzy: Fixed according to OmahaProxy
https://www.chromium.org/developers/calendar
(both stable and beta are version 62 there so I omit beta for now)

mgol: True, there should be no version marked as beta currently in that case.

mgol:
@chicoxyzzy It claims you haven't committed all the changes in generated templates.

chicoxyzzy:
Actually I did. I'll try to rebase to rerun CI build

chicoxyzzy:
CI is broken :(
See last commit in #1200

mgol:
I've tried `npm run build` on this PR locally and there were, indeed, no changes, Which is weird because when I rebased #1196 today the tests passed.

I don't understand those errors.

afmenez:
Sorry, I opened a similar PR (#1204) because I haven't seen yours, But I believe my results are more precise.

environments.json
```diff
@@ -1496,23 +1496,32 @@
     "family": "V8",
     "short": "CH&nbsp;61,<br>OP&nbsp;48",
     "release": "2017-09-05",
+    "retire": "2017-10-17",
     "note_id": "experimental-flag",
     "obsolete": false
```
afmenez, line 1501: This should be `true` now.

afmenez (changes requested):
There are some updates to esnext results too.

environments.json
mgol: @afmenez See https://github.com/kangax/compat-table/pull/1201#discussion_r145312366

environments.json
afmenez: I disagree.  I DO believe that a new version of Chrome and Firefox makes the previous version obsolete, since the previous version is discontinued and stops receiving security updates.  In Chrome's case, the previous version is not even available for installation anymore.

environments.json
afmenez: I disagree and commented there.  Also, the retire date should be when all the new versions are released, Android and iOS are still missing.

environments.json
mgol: Safari 10.1 won't receive further updates either and it's not marked as obsolete. So upstream support isn't treated as a hard requirement for not being obsolete in the data. The same applies to iOS.

chicoxyzzy: Showing last two versions makes a lot of sense to me. Browser update doesn't happen immediately.

ljharb: Agreed; obsolescence has nothing to do with support status imo.

afmenez:
What's the point of having two levels of obsolescence then (true and very)?  My understanding is that we show by default current (and supported) versions, and hide the old versions still in use.  Versions that are
unsupported and unused are marked with "obsolete:very" and don't show at the table at all.
This is how it's been so far, and this is about updating the current state.  If we want to change how this are done, it should be a separate PR, and for all browsers, not just Chrome.
And the table is already too big (and slow) as it is now.

ljharb:
Table speed is something that needs to be fixed inherently, not by just sweeping browsers under the "obsolete" rug.

As for "big", it's *supposed* to be big - there's lots of browsers out there, and *all* of their results matter.

"Very" obsolete is indeed to soft-delete, without actually deleting, results for browsers nobody actually cares about anymore. "Obsolete" to me is for a different purpose; it's simply to notate when a browser is obsolete - but not to hide it.

Chrome 61 isn't conceptually obsolete just because Chrome 62 came out - it takes a bit of time before that's the case.

chicoxyzzy:
Let's track #1204

PR #1204, opened 2017-10-18
Chrome 62 released

afmenez:
Chrome 62 was released yesterday.  Updating metadata and test results.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -1497,7 +1497,7 @@
     "short": "CH&nbsp;61,<br>OP&nbsp;48",
     "release": "2017-09-05",
     "note_id": "experimental-flag",
-    "obsolete": true
+    "obsolete": false
   },
   "chrome62": {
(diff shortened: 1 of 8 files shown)


environments.json
```diff
@@ -1497,22 +1497,30 @@
     "short": "CH&nbsp;61,<br>OP&nbsp;48",
     "release": "2017-09-05",
     "note_id": "experimental-flag",
-    "obsolete": false
```
ljharb: I'm not comfortable with v61 being marked obsolete just because v62 was released.

chicoxyzzy: Me too

afmenez: This is how this was done so far.  Ditto for Firefox.

afmenez: I mean, if we want to change how things work, it should be a separate PR for all browsers, not just Chrome.

chicoxyzzy (approved):

PR #1206, opened 2017-10-18
Hide ie10 from "es2016plus" tab and ie11 from "esnext" tab

### LINKED ISSUE #1205: Hide IE11 from es2016+ and next tabs by default?
Does it make sense to show IE11 in es2016+ and next tabs by default?

chicoxyzzy:
Closes #1205.

Note: we can't hide IE 11 from "es2016plus" tab yet (until it's not marked as obsolete) because in that case it will also disappear from list of obsolete browsers

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -480,7 +480,6 @@
     "test_suites": [
       "es5",
       "es6",
-      "es2016plus",
       "esintl",
       "non-standard"
     ]
@@ -490,7 +489,14 @@
     "family": "Chakra",
     "short": "IE 11",
     "release": "2013-10-17",
-    "obsolete": false
+    "obsolete": false,
+    "test_suites": [
+      "es5",
+      "es6",
+      "es2016plus",
+      "esintl",
+      "non-standard"
+    ]
   },
   "edge12": {
     "full": "Microsoft Edge 12",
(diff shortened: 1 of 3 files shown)


afmenez (approved):
LGTM

ljharb (dismissed):
What's the value of hiding it here? It's useful when on these tabs to see that a browser supports none of it, and it will forever (as long as people are using that browser) be useful.

chicoxyzzy:
IE11 won't ever support esnext features and IE10 won't ever support es2016+ features. IE10 is still represented in es6 tab under obsolete checkbox and IE11 is still represented in es6 and es2016+ tabs by default

ljharb:
@chicoxyzzy I think it's important that the *last* browser with zero support on a tab show up by default.

afmenez:
The most important thing is to have clear rules on what is shown by default, what can be selected, what is left out, instead of deciding at random on each PR.
There is an issue for this already: #740

ljharb:
Totally agree we need a standard definition for these things.

chicoxyzzy:
> I think it's important that the last browser with zero support on a tab show up by default.

This is what this PR actually do. As I said

> IE10 is still represented in es6 tab under obsolete checkbox and IE11 is still represented in es6 and es2016+ tabs by default

I am not sure that there is any value for IE11 being in esnext tab

PR #1212, opened 2017-11-01
Splitting TypeScript versions and adding results

afmenez:
TypeScript 2.6 was released yesterday, this PR splits the versions and adds results.


ljharb (approved):
Assuming you've tested all these results, this looks great.

afmenez:
Sorry, no.  All the changes are based on release notes archeology.
https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript

ljharb:
ah :-/ still fine for now, but it'd be ideal if we came up with a better value for these tests than `false` or `true` - something like an enum of "allegedly passes, via inheritance or release notes etc", "a human verified it passes", "a tool verified it passes", "it fails", and "untested".

afmenez:
Isn't this too complicated?  I would prefer a simple no/?/yes, with notes indicating an inherited or untested result.

ljharb:
It might be; the value of having two "passes" items is that if a human has determined it, it's inviolate, whereas if a tool has determined it, a bugfix in the tool can safely regenerate that result.

chicoxyzzy (approved):
LGTM. Needs rebase though

afmenez:
Rebased.

PR #1216, opened 2017-11-10
Firefox 58 now implements "Promise.prototype.finally"

afmenez:
The latest Firefox Nightly now passes all the "Promise.prototype.finally" tests.
https://bugzilla.mozilla.org/show_bug.cgi?id=1019116

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -2521,6 +2521,7 @@ exports.tests = [
       res: {
         babel: babel.corejs,
         typescript: typescript.corejs,
+        firefox2: false,
         firefox52: false,
         opera10_50: false,
         duktape2_2: false,
@@ -2534,6 +2535,7 @@ exports.tests = [
       res: {
         babel: babel.corejs,
         typescript: typescript.corejs,
+        firefox2: false,
         firefox52: false,
         opera10_50: false,
         duktape2_2: false,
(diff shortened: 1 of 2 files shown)


data-esnext.js
```diff
@@ -2676,7 +2676,8 @@ exports.tests = [
       res: {
         babel: babel.corejs,
         typescript: typescript.corejs,
-        firefox52: false,
+        firefox2: false,
```
ljharb: Did you explicitly test Firefox 2? If not, it shouldn’t be in here.

Either way, please keep the explicit v52 results.

ljharb (changes requested):

data-esnext.js
afmenez: Yes, I have tested firefox 2 and every ESR version up to 52, and also versions 53..58.  I have also tested with many versions of Chrome and Edge.  I don't see the point of adding every version.

ljharb: Generally, these entries should represent explicit human test results, and thus should never be removed (unless incorrect).

I agree having all of those versions in the list is impractical; maybe we should settle for including 2, 57, and 58?

afmenez: Done.

ljharb (approved):
Thanks!

chicoxyzzy (approved):

afmenez:
This was Firefox 59 nightly actually, I will fix it with the 57 release patch.

PR #1225, opened 2017-11-16
Clarify the meaning of the “Proxy ownKeys handler” test

### LINKED ISSUE #1224: es2016+: test about duplicate keys in Proxy ownKeys handler went mixed with some “(ES Draft Style)” nonstandard feature recently removed from FF
Something strange happened.

es-2016-strangeness

claudepache:
* Replace incorrect “(ES draft style)” with “(ES 2017 semantics)”.
* Add a note saying that the semantics tested was modified later.

Closes #1224

Note: An updated test with the new semantics may (and should) be added in a separate PR.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -1982,11 +1982,17 @@ exports.tests = [
     ]
   },
   {
-    name: 'Proxy "ownKeys" handler, duplicate keys for non-extensible targets (ES Draft style)',
+    name: 'Proxy "ownKeys" handler, duplicate keys for non-extensible targets (ES 2017 semantics)',
     category: '2017 misc',
     significance: 'tiny',
     spec: 'https://github.com/tc39/ecma262/pull/594',
     mdn: 'https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/ownKeys',
+    links: [
+      {
+        note_id: 'proxy-duplictate-ownkeys-updated',
+        note_html: 'The behaviour of the Proxy “ownKeys” handler in presence of duplicate keys has been <a href="https://github.com/tc39/ecma262/issues/833">modified later</a>.',
+      }
+    ],
     exec: function() {/*
      var P = new Proxy(Object.preventExtensions(Object.defineProperty({a:1}, "b", {value:1})), {
      ownKeys: function() {
--- a/es2016plus/index.html
+++ b/es2016plus/index.html
@@ -4458,7 +4458,7 @@ return typeof Atomics.xor == &apos;function&apos;;
 </tr>
 <tr class="category"><td colspan="66">2017 misc</td>
 </tr>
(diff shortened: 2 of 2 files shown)


data-es2016plus.js
```diff
@@ -1982,11 +1982,17 @@ exports.tests = [
     ]
   },
   {
-    name: 'Proxy "ownKeys" handler, duplicate keys for non-extensible targets (ES Draft style)',
+    name: 'Proxy "ownKeys" handler, duplicate keys for non-extensible targets (ES 2017 semantics)',
```
afmenez, line 1985: I believe "Old ES2017 semantics" would be clearer.

afmenez (approved):

data-es2016plus.js
ljharb: This change (https://github.com/tc39/ecma262/issues/833) was incorporated into ES2017. Were the prior semantics introduced only after ES2016?

chicoxyzzy (commented):
Not sure if we should keep this test in es2017+ or move it to non-standard. AFAIR we have some tests with new semantics in old tests (ES5?)

data-es2016plus.js
claudepache: @ljharb [ECMAScript® 2017 has been published in June 2017](https://www.ecma-international.org/ecma-262/8.0/). The change was [merged](https://github.com/tc39/ecma262/commit/5d770579571ff3e10f3fc4ce0bf2c9ce0799d107) later, in August, and will be incorporated in ES2018.

claudepache: > I believe "Old ES2017 semantics" would be clearer.

Yes. But it could be slightly confusing as well, as it may suggest an inexistent “new ES 2017 semantics”. What do other people think, should I add “old”?

ljharb: Ahh right thanks for clarifying.

The. “ES2017 semantics” and “ES2018 semantics” seem like good names?

afmenez: "ES2017 semantics" sounds new to me, which can be misleading.  How about "Deprecated ES2017 semantics"?

ljharb: If there’s a 2018 test next to it, 2017 wouldn’t sound replaced?

afmenez: I thought the ES2017 test would be moved to non-standard, and the ES2018 would be implemented here...

claudepache: > I thought the ES2017 test would be moved to non-standard, and the ES2018 would be implemented here...

Moving to non-standard is not the only possibility. But we should discuss that point in another issue.

ljharb: Technically the ES2017 semantics *are* standard; they've just been superseded by a later one. I can see the argument for putting them in either place.

afmenez: If they are both on the same tab, it will be impossible to reach 100% on the tab.

ljharb: Fair point; that's an argument to move it to the non-standard tab. However, I don't particularly want to be optimizing for "reaching 100%" :-/

claudepache: Again, this is not the only possibility. Hint: in the ES5 tab, there is a test named “arguments.caller removed or is a TypeError” (originally it was: “arguments.caller is a TypeError”). But we could discuss that point in another issue.

ljharb: True, but I don't want to create a test that claims "passing" if an engine never implements the 2018 semantics.

chicoxyzzy (approved):

ljharb (approved):

ljharb:
Let's address further questions in a new issue/PR.

PR #1226, opened 2017-11-24
update Chrome 64 results

chicoxyzzy:


data-esnext.js
```diff
@@ -1812,6 +1812,7 @@ exports.tests = [
     firefox2: false,
     opera10_50: false,
     chrome60: chrome.harmony,
+    chrome64: true,
```
ljharb, line 1815: is chrome63 also `chrome.harmony`, such that 64 is the first version with these unflagged?

ljharb (approved):

data-esnext.js
chicoxyzzy: Yes it is
https://github.com/tc39/proposal-regexp-named-groups#implementations

PR #1228, opened 2017-11-28
Move Array.prototype.flat{Map,ten} to stage 3

chicoxyzzy:
As per https://github.com/tc39/proposals/pull/89


afmenez (approved):
Nice.  I did a similar path (#1230) with updated results though.

PR #1229, opened 2017-11-28
move Numeric Separators to stage 3

chicoxyzzy:


afmenez (approved):
LGTM.  I did a similar path (#1230) with updated results though.

PR #1230, opened 2017-11-28
Array.prototype.{flatten, flatMap} and Numeric Separators moved to Stage 3

afmenez:
Moving the tests and updating results for old Forefox and Opera versions.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -2776,7 +2776,7 @@ exports.tests = [
 },
 {
   name: 'Array.prototype.{flatten, flatMap}',
-  category: STAGE2,
+  category: STAGE3,
   significance: 'medium',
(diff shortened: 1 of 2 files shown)


afmenez:
This does the same changes as #1228 and #1229, but in a single push to avoid the merge conflicts.  It also adds some results too.

afmenez:
@chicoxyzzy: I did this patch without knowing about the other ones.  Sorry for the duplication.

ljharb (approved):
LGTM

afmenez:
Tks!

chicoxyzzy:
@afmenez that's ok but please check if PR already exists to be sure we don't do extra work and don't create extra noise in notifications

PR #1234, opened 2017-11-30
Class fields downgraded to Stage 2

afmenez:
Class fields downgraded to Stage 2

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -2156,7 +2156,7 @@ exports.tests = [
 },
 {
   name: 'class fields',
-  category: STAGE3,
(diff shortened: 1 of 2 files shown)


ljharb (changes requested):
Let’s hold off on this; the meeting’s not over yet :-)

afmenez:
I believe the "class fields" proposal was actually splitted on a Stage2 and a Stage3 parts.  This patch is not valid anymore, I'll let someone more experienced write a more complete one.

ljharb:
Indeed, it was. Statics are now stage 2, and instances are stage 3.

PR #1235, opened 2017-12-05
default params in arrow functions pass in Chrome 49

adrianheine:
https://github.com/v8/v8/commit/ff932fe8f4f82a6832c4f649ad00b3ae59f53822 fixed default parameters in arrow functions, https://github.com/v8/v8/commit/e752f964b645fc7aafd99fd684914a10831619e8 removed the flag. The latter was released in V8 4.8.284, which was shipped in 49.0.2623. I did not check this, but it wasn't tested in 6c0b2fcfec89d97740ab42e7cf8eafe202ca1f65, either.


data-es6.js
```diff
@@ -12960,7 +12960,7 @@ exports.tests = [
         typescript: true,
         traceur: true,
         edge15: false,
-        chrome50: true,
```
ljharb, line 12963: can we add `chrome48: false` as well, assuming that's the case?

ljharb (changes requested):
(please run the build process as well, to update the html)

data-es6.js
adrianheine: I think it should be `chrome47: "flagged"`.

adrianheine:
This doesn't affect the rendered HTML, since these environments are very obsolete.

data-es6.js
ljharb: Perfect

ljharb (approved):

afmenez (approved):
LGTM

chicoxyzzy (approved):

PR #1238, opened 2017-12-20
Add Node8+ to global proposal

danez:
The first test of the global proposal still works.


afmenez:
Why the result was not cascaded from node7?  I don't mind adding the explicit results, but I believe this is a problem on `environments.json`, not on the results.

ljharb (approved):

ljharb:
Good point. The explicit results are nice, but the cascading should be working here.

chicoxyzzy (approved):

PR #1239, opened 2017-12-20
Add BigInt

danez:
I created tests for BigInt based on the current spec. As no environment supports it yet I cannot really test if they work especially the typed array tests, but to my understanding they should be correct. Maybe someone can crosscheck? (cc @littledan @caiolima)

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -3455,6 +3455,84 @@ exports.tests = [
   ]
 },
 {
+  name: 'BigInt',
+  category: STAGE3,
+  significance: 'medium',
+  spec: 'https://github.com/tc39/proposal-bigint',
+  subtests: [
+    {
+      name: 'basic functionality',
+      exec: function () {/*
+        return (1n + 2n) === 3n;
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'constructor',
+      exec: function () {/*
+        return BigInt("3") === 3n;
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'BigInt.asUintN',
+      exec: function () {/*
+        return typeof BigInt.asUintN === 'function';
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'BigInt.asIntN',
+      exec: function () {/*
+        return typeof BigInt.asIntN === 'function';
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'BigInt64Array',
+      exec: function () {/*
+        var buffer = new ArrayBuffer(64);
+        var view = new BigInt64Array(buffer);
+        view[0] = 0x8000000000000000n;
+        return view[0] === -0x8000000000000000n;
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'BigUint64Array',
+      exec: function () {/*
+        var buffer = new ArrayBuffer(64);
+        var view = new BigUint64Array(buffer);
+        view[0] = 0x10000000000000000n;
+        return view[0] === 0n;
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'DataView.prototype.getBigInt64',
+      exec: function () {/*
+        return typeof DataView.prototype.getBigInt64 === 'function';
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'DataView.prototype.getBigUint64',
+      exec: function () {/*
+        return typeof DataView.prototype.getBigUint64 === 'function';
+      */},
+      res: {
+      },
+    },
+  ],
+},
+{
   name: 'String.prototype.replaceAll',
   significance: 'small',
   spec: 'https://github.com/tc39/proposal-string-replace-all',
--- a/esnext/index.html
+++ b/esnext/index.html
@@ -2433,6 +2433,684 @@ return 1_000_000.000_001 === 1000000.000001 &amp;&amp;
 <td class="no" data-browser="ios11">No</td>
 <td class="no" data-browser="ios11_3">No</td>
 </tr>
+<tr class="supertest" significance="0.5"><td id="test-BigInt"><span><a class="anchor" href="#test-BigInt">&#xA7;</a><a href="https://github.com/tc39/proposal-bigint">BigInt</a></span></td>
+<td class="tally" data-browser="tr" data-tally="0">0/8</td>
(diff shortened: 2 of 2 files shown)


data-esnext.js
```diff
@@ -3382,6 +3382,90 @@ exports.tests = [
...
+        return view[0] === -9223372036854776000n;
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'BigUint64Array',
+      exec: function () {/*
+        var buffer = new ArrayBuffer(64);
+        var view = new BigUint64Array(buffer);        view[0] = 0x10000000000000000;
```
littledan: You probably want to put an `n` on the end of that literal, unless you want it to throw a TypeError (with the current spec); ditto above.

data-esnext.js
```diff
@@ -3382,6 +3382,90 @@ exports.tests = [
...
+      exec: function () {/*
+        return typeof BigInt.asIntN === 'function';
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'BigInt.parseInt',
+      exec: function () {/*
+        return typeof BigInt.parseInt === 'function';
```
littledan: This function has been removed in the current draft specification.

littledan (commented):
Looks almost right, modulo these nits. Generally, if you're interested in compatibility tests for BigInts and other new JS features, integrating Kangax with Test262 could be another path which doesn't require writing all of these parallel tests.

data-esnext.js
danez: Thanks fixed

caiolima:
LGTM.

data-esnext.js
```diff
@@ -3382,6 +3382,82 @@ exports.tests = [
...
+      exec: function () {/*
+        return BigInt("3") === 3n;
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'BigInt.asUintN',
+      exec: function () {/*
+        return typeof BigInt.asUintN === 'function';
```
ljharb, line 3482: it'd be nice if all these tested the functions' behavior, not just their presence

data-esnext.js
```diff
@@ -3382,6 +3382,82 @@ exports.tests = [
...
+        return typeof BigInt.asIntN === 'function';
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'BigInt64Array',
+      exec: function () {/*
+        var buffer = new ArrayBuffer(64);
+        var view = new BigInt64Array(buffer);         view[0] = 0x8000000000000000n;
```
ljharb: i think this is missing a newline before "view[0]" :-)

data-esnext.js
```diff
@@ -3382,6 +3382,82 @@ exports.tests = [
...
+        return view[0] === -0x8000000000000000n;
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'BigUint64Array',
+      exec: function () {/*
+        var buffer = new ArrayBuffer(64);
+        var view = new BigUint64Array(buffer);        view[0] = 0x10000000000000000n;
```
ljharb: here too

afmenez:
The [release notes](https://webkit.org/blog/8042/release-notes-for-safari-technology-preview-46/) say that Safari TP 46 already supports BigInt.

data-esnext.js
danez: If I could test them I could write a test, but right now I would simply guess what the result is and therefore I only added the typeof test.

danez:
I tried TP 46 and although the changelog states it should work it does not for me at least in the developer console.
So either the release notes are wrong, or there is a secret flag to activate them or the dev console does not yet support it.

littledan:
I believe BigInt is being developed behind a command-line flag `--useBigInt=true` that's just for JSC on the command line and not exposed in Safari, cc @caiolima

caiolima:
By default, BigInt into Safari is disabled. As Dan said, you need to enable it from command line or running WebKit with ```JSC_useBigInt=true```.

chicoxyzzy:
@danez do you want to make changes (i.e. at least add flag note)? Or I can do it on top of your branch if you want

danez:
Sure go ahead, I don't have time right now. Thanks

chicoxyzzy:
@caiolima I'm trying to open Safari TP 47 with `--args --JSC_useBigInt=true` but still can't get BigInt work

caiolima:
@chicoxyzzy Sorry for the late response here, but maybe it can be useful yet. You need to use ```JSC_useBigInt=true``` as ENV_VAR or then use ```--useBigInt=true``` if running WebKit from CLI.

mathiasbynens:
Chrome 67 (currently in Canary) supports `BigInt`s out of the box.

chicoxyzzy:
@mathiasbynens good to know, thanks! Though as far as I can see current Canary is 68. Did you meant Beta / Dev 67 or Canary 68?

@caiolima I've tried to run Safari TP from CLI with `--useBigInt=true` but BigInt is still not available even in Safari TP 54

mathiasbynens:
@chicoxyzzy Sorry for the confusion. Anything that has Chrome v67 or greater works.

data-esnext.js
mathiasbynens: @danez You can use Chrome 67 (Dev or Canary) to test now :)

mathiasbynens: @ljharb Is this blocking the merge?

IMHO it would be better to have the data included in the table than to have nothing. We can always improve the tests in a follow-up commit. WDYT?

ljharb (approved):
LGTM; let's add test results in a separate PR.

caiolima:
@chicoxyzzy I don't know how to run Safari TP from command line, but AFIK, it is not possible set ```--useBigInt=true``` there. You probably can do that using ```JSC_useBigInt=true <command to run Safari TP```. I use ```--useBigInt=true``` when I run JSC CLI.

chicoxyzzy:
```
JSC_useBigInt=true /Applications/Safari\ Technology\ Preview.app/Contents/MacOS/Safari\ Technology\ Preview
```
Doesn't work
image

caiolima:
@chicoxyzzy I asked on IRC how can I do that into Safari TP. I never tried enable JSC dynamic flags outside JSC CLI, and I don't know if it is possible to do that.

PR #1241, opened 2018-01-07
split class fields into two proposals

chicoxyzzy:
On November meeting class fields proposal was split into instance class properties and [static class properties](https://github.com/tc39/proposal-static-class-features/) (stage 2)


ljharb (commented):
Both proposals now have public and private; can we add static private tests as well?

ljharb (approved):
LGTM

PR #1243, opened 2018-01-11
Shared Mem and Atomics disabled in FF57 to mitigate Meltdown and Spectre

afmenez:
Shared Memory and Atomics disabled in Firefox 57.0.4 to mitigate the Meltdown and Spectre CPU bugs:
https://blog.mozilla.org/security/2018/01/03/mitigations-landing-new-class-timing-attack/


ljharb (approved):
SharedArrayBuffer has also been disabled in Chrome (I've verified it's not present in Chrome 63; not sure about earlier versions). Could we include that in this PR also?

afmenez:
OK, updated the data on Chrome and fixed a mistake on the ES6 notes fix.

ljharb (approved):
Thanks!

PR #1246, opened 2018-01-17
Safari 11 and TP SharedArrayBuffer temporary removing

### LINKED ISSUE #1245: Shared Mem and Atomics disabled in Safari to mitigate Meltdown and Spectre
Shared Memory and Atomics were disabled in Safari to mitigate the Meltdown and Spectre CPU bugs, according to this blog:
https://webkit.org/blog/8048/what-spectre-and-meltdown-mean-for-webkit/

On #1243, I updated the situation for Firefox and Chrome, but I have no access to Safari to test which versions are affected.

chicoxyzzy:
closes #1245

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -871,6 +871,12 @@ exports.tests = [
             note_html: 'The feature was temporarily disabled to mitigate the Meltdown and Spectre CPU bugs</a>.',
           },
           safari10_1: true,
+          safari11: {
+            val: false,
+            note_id: 'sf-shared-memory-spectre',
+            note_html: 'The feature was temporarily disabled to mitigate <a href="https://webkit.org/blog/8048/what-spectre-and-meltdown-mean-for-webkit/">the Meltdown and Spectre CPU bugs</a>.',
+          },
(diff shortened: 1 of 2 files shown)


data-es2016plus.js
```diff
@@ -914,7 +934,16 @@ exports.tests = [
           chrome48: chrome.sharedmem,
           chrome60: true,
           chrome63: { val: false, note_id: 'chr-shared-memory-spectre' },
-          safari11: true,
+          safari11: {
+            val: false,
+            note_id: 'sf-shared-memory-spectre',
+            note_html: 'The feature was temporarily disabled to mitigate the Meltdown and Spectre CPU bugs</a>.',
```
ljharb: do we need the repeated notes? if so, can they all have the link?

also, this one in particular has a `</a>` but no `<a>`.

ljharb (changes requested):

data-es2016plus.js
chicoxyzzy: You are right. It can be introduced once. Link issue is also fixed. Thank you!

ljharb (approved):

PR #1249, opened 2018-01-22
Add Chrome Canary 66

chicoxyzzy:


ljharb (approved):

ljharb:
Are we sure that these features that are enabled in Canary unflagged will be enabled in stable unflagged when it lands?

chicoxyzzy:
@ljharb I'm not sure about that, but we can add necessary flags later if needed

PR #1251, opened 2018-01-23
Firefox 58 release

chicoxyzzy:


afmenez:
It seems we had the same idea at the same time, but I believe my patch (#1252) took a little longer because it is more thru... ;)

chicoxyzzy:
I'll take my patch this time. It seems that your patch could be applied on top of it

afmenez:
Nope, I was unable to rebase.  Created #1253 instead.

PR #1254, opened 2018-01-24
Features status update according to TC39 January meeting

chicoxyzzy:
Changes:
- object rest/spread properties to stage 4
- RegExp named capture groups to stage 4
- RegExp Lookbehind assertions to stage 4
- Unicode Property escapes to stage 4
- Promise.prototype.finally to ES2018 to stage 4


zloirock:
Hmm... I made some related commits at the same time.

chicoxyzzy:
@zloirock updated my PR

data-es2016plus.js
```diff
@@ -2142,6 +2142,57 @@ exports.tests = [
       duktape2_0: false,
     },
   },
+  {
+    name: 'object rest/spread properties',
+    significance: 'large',
```
chicoxyzzy: For reviewers: notice that I changed significance to large because of huge amount of code on GitHub which uses this feature more than 3 years already

chicoxyzzy:
@zloirock please review PR

data-es2016plus.js
zloirock: I'm not sure that this change makes sense. At this moment, "large" significance only for features with new conceptions like async functions, zones, realms, etc. Object rest/spread is just a simple syntax sugar.

zloirock (approved):
The rest LGTM.

data-es2016plus.js
chicoxyzzy: Ok, reverted

ljharb:
The meeting's not quite over yet :-)

chicoxyzzy:
We'll prepare more PRs 🙂. Also we should add some new tests for missing proposals from previous and current meetings

ljharb:
In addition, matchAll changed semantics (strings passed in now get the global flag) so we need to add a test for that.

In general tho, there's no rush. The compat table should probably wait until a meeting is over before merging updates.

PR #1255, opened 2018-01-24
Chrome 64 release

chicoxyzzy:
According to https://omahaproxy.appspot.com Beta is also still 64. I marked versions older than a year as very obsolete


zloirock (approved):
👍

ljharb (approved):

PR #1256, opened 2018-01-28
Add Safari 11.1

chicoxyzzy:
Tested on Safari 11.1 (13605.1.23.1) which is available in macOS 10.13.4 Beta


chicoxyzzy:
Also Safari 11.1 is available for download from here https://developer.apple.com/safari/download/

chicoxyzzy:
Also I just updated Safari TP version and release date. All tests results are same as previous.

Issue #438, opened 2015-02-15
Add JDK9 Nashorn results


webbedspace:
Nashorn, according to some sources, supports let and const.

kangax:
But supports how... that is the question :)

let and const are the lands of chaos (I was even gonna write a post about it)

anba:
Test scripts + results using nashorn-jdk9 tip: https://gist.github.com/anba/24f2b158aa7e58930bf2

TLDR; the `let` and `const` tests pass except for:

```
let - for-loop iteration scope
let - for-loop iteration scope (strict mode)
```

dead-claudia:
@kangax This and Rhino (which is even more popular) both need added. Rhino is frequently embedded into Java applications, especially Android ones, where Nashorn doesn't exist. They both should be added to the server-side engines.

mattkime:
+1 @impinball

AMorgaut:
+1 for Nashorn & Rhino

we have an Android project in which we embed rhino

chicoxyzzy:
Rhino's compatibility table
http://mozilla.github.io/rhino/compat/engines.html

zeitiger:
It would be quite helpful see which features are available. The documentation of Oracle is not really helpful, there is only the claim that Nashorn/Java8 is ECMAScript 5.1 complaint. But there is also a ES6 flag and complete not idea what will that enable? In the roadmap for Java9 I found the target to to be complaint to ECMAScript 2015.

mkurz:
+1

nilskp:
Not just JDK9 Nashorn, also JDK8.

pixeldrew:
Would really like this so I can use it with Babel-ENV.

PR #1261, opened 2018-02-13
Added a test script for Oracle nashorn engine

pmlopes:
This is a script to test Oracle Nashorn and updates tables, that relates to:

https://github.com/kangax/compat-table/issues/438

The script can be run with any version and I've provided updated runs of:

* Nashorn 1.8 (JDK8)
* Nashorn 9 (JDK9)

Once JDK10 gets released the data can be computed by running:

```
JAVA_HOME=/path/to/jdk10 node nashorn.js
```

And update the tables with the new info.


chicoxyzzy (commented):
Great work! I'll try to check PR before next week!

chicoxyzzy (approved):
I can confirm it works on macOS.

PR #1268, opened 2018-02-26
Add Babel 7

chicoxyzzy:


ljharb (changes requested):
I don't think babel 7 is out yet, and I don't think we should add it until it's out of beta.

chicoxyzzy:
@ljharb it is marked as unstable so I think it should be ok to add it

ljharb (approved):
yeah ok fair

zloirock:
IIRC some `babel@7` features already were added to `babel` column.

chicoxyzzy:
@zloirock yes I removed numeric separators from Babel 6 column

PR #1276, opened 2018-03-13
Explicitly set tested IE11 results to false

afmenez:
I am explicitly setting tested IE11 results to false, in preparation for the default changing to "null" in #1131.


afmenez:
Tks!

PR #1278, opened 2018-03-21
Safari TP 52 release

chicoxyzzy:
flatten and flatMap are enabled by default since TP 51

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -2631,6 +2631,7 @@ exports.tests = [
         firefox58: false,
         firefox59: firefox.nightly,
         opera10_50: false,
+        safaritp: true,
         duktape2_2: false,
       }
     },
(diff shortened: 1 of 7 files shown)


ljharb (approved):
Seems fine, but "flatten" isn't going to be the name anymore, so I'm not sure if it's worth adding this.

chicoxyzzy:
@ljharb sure, but our tests still use 'flatten' and 'flatMap' as method names. Did TC39 reach consensus on new names?

ljharb:
Not yet.

chicoxyzzy:
@ljharb what do you think about last commit? Is wording correct? I think this note makes all results in `Array#flatten` row more clear

ljharb:
Looks good, altho it’s only flatten that is going to change.

PR #1282, opened 2018-03-28
[ESNext] Add some new proposals

zloirock:
- [`Object.fromEntries`](https://github.com/tc39/proposal-object-from-entries)
- [Getting last item from `Array`](https://github.com/keithamus/proposal-array-last)
- [New `Set` methods](https://github.com/tc39/proposal-set-methods)
- [New collections (`Set` and `Map`) methods](https://github.com/tc39/proposal-collection-methods)

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -3559,6 +3559,17 @@ exports.tests = [
       res: {
       }
     },
+    {
+      name: 'Set.prototype.xor',
+      exec: function () {/*
+        var set = new Set([1, 2]).xor([2, 3]);
+        return set.size === 2
+          && set.has(1)
+          && set.has(3);
+      */},
+      res: {
+      }
+    },
   ]
 },
 ];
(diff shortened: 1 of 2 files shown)


data-esnext.js
```diff
@@ -3483,6 +3483,261 @@ exports.tests = [
...
+    {
+      name: 'Array.prototype.lastItem',
+      exec: function () {/*
+        return [1, 2, 3].lastItem === 3;
+      */},
+      res: {
+      }
+    },
+    {
+      name: 'Array.prototype.lastIndex',
```
ljharb, line 3513: this is highly likely to go away, fwiw

ljharb (approved):

data-esnext.js
zloirock: I don't see anything about it in the repo of the proposal.

zloirock: But I agree, I don't see any use cases for this getter.

ljharb: you're right, there's nothing in there yet. but `[].lastIndex` being `0` is super weird, and some on the committee were uncomfortable with it.

PR #1285, opened 2018-04-01
update closure-compiler to v20180319 and v20180402

teppeis:
- v20180204
    - fix some incorrect results
- v20180319
    - improve generators
- v20180402
    - improve Array#includes
    - add Promise#finally

cf. https://github.com/google/closure-compiler/wiki/Releases

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -48,8 +48,33 @@
     ]
   },
   "closure": {
+    "full": "Closure Compiler v20180204",
+    "short": "Closure 2018.02",
+    "family": "Closure",
+    "platformtype": "compiler",
+    "obsolete": true,
+    "test_suites": [
+      "es6",
+      "es2016plus",
+      "esnext"
+    ]
+  },
+  "closure20180319": {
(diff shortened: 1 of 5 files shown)


environments.json
```diff
@@ -48,7 +48,7 @@
     ]
   },
   "closure": {
-    "full": "Closure Compiler v20180204",
```
ljharb: is there no value in keeping results from older versions?

teppeis: @ljharb Closure Compiler is updated once every month or two month.
If this project can store the all releases, it's great and I would update this PR.

ljharb: That seems ideal to me (assuming the old ones can be marked as obsolete/very as appropriate, so the UI stays manageable) - let's see what other collabs think.

teppeis: This is separated column version: https://github.com/teppeis/compat-table/tree/closure-compiler-v20180319-new-column

ljharb (approved):

chicoxyzzy (approved):

PR #1288, opened 2018-04-22
Mark Node 4 as obsolete

chicoxyzzy:


ljharb (changes requested):
I don’t agree that it’s obsolete just because it’s out of LTS.

chicoxyzzy:
@ljharb its end of life is 2018-04-30

https://github.com/nodejs/Release

chicoxyzzy:
https://twitter.com/nodejs/status/987299954174693376

ljharb:
Yes, but obsolescence is not determined solely or even primarily by official vendor support.

afmenez:
This discussion comes back from time to time, and the conclusion is always different.  I do believe that it IS obsolete when it gets out of LTS, and that's what the tag obsolete=true is for.  I do not believe that the results should be removed from the table while there is significant use of this version, and that would be obselete="very".

afmenez (approved):
LGTM

chicoxyzzy:
@ljharb now it's time I think

chicoxyzzy:
ping @ljharb

ljharb (approved):

PR #1294, opened 2018-05-07
Updating test results for Edge 16 and Edge 17

afmenez:
- Edge 16 does not support SIMD extensions anymore;
- Adding a note Edge 17 results for shared memory;
- Adding versions to browser descriptions;
- Obsoleting ancient versions;


ljharb (approved):

chicoxyzzy (approved):

data-es2016plus.js
```diff
@@ -882,7 +882,11 @@ exports.tests = [
           ie11: false,
           edge15: edge.experimental,
           edge16: true,
-          edge17: false,
+          edge17: {
+            val: false,
+            note_id: 'edg-shared-memory-spectre',
+            note_html: 'The feature was temporarily disabled to mitigate the Meltdown and Spectre CPU bugs</a>.',
```
chicoxyzzy, line 888: Just noticed there is only closing `</a>` tag without opening `<a>` tag

afmenez: My bad.  I didn't notice it when I copied the note from Chrome.  Fixed both on #1295 .

PR #1297, opened 2018-05-09
Firefox 60 released

afmenez:
Firefox 60 released today.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -921,9 +921,9 @@
     "obsolete": false
   },
   "firefox60": {
-    "full": "Firefox 60",
+    "full": "Firefox 60 ESR",
     "family": "SpiderMonkey",
-    "short": "FF 60",
+    "short": "FF 60 ESR",
     "release": "2018-05-09",
     "obsolete": false
   },
--- a/environments.json
+++ b/environments.json
@@ -861,7 +861,7 @@
     "family": "SpiderMonkey",
     "short": "FF 52 ESR",
     "release": "2017-03-07",
-    "retire": "2018-08-28",
+    "retire": "2018-08-21",
     "obsolete": false
   },
   "firefox53": {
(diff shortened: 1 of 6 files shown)


ljharb (approved):

environments.json
```diff
@@ -921,20 +921,28 @@
     "obsolete": false
   },
   "firefox60": {
-    "full": "Firefox 60 Beta",
+    "full": "Firefox 60",
```
chicoxyzzy: It should be `Firefox 60  ESR`

chicoxyzzy (changes requested):
The full name should be Firefox 60 ESR. Otherwise LGTM

afmenez:
I believe this may be confusing.  On every ESR release before this one, we only added the ESR suffix when the next version came out.

chicoxyzzy:
Hm, I don't understand what exactly is confusing. Firefox 60 is ESR so we should reflect this in compatibility table. 

https://www.mozilla.org/en-US/firefox/organizations/

image

afmenez:
You see, this is exactly the confusing part. As you can see in the graph, Firefox ESR (in yellow) are separate releases from the mainline (in green).  Firefox 60 IS not an ESR release, it HAS an ESR release.  Most people are in the main release train and don't even know what the ESR is, and that's why it may be confusing.
The rendering engine is identical for the normal and ESR releases, so I don't believe we should have 2 separate entries.

chicoxyzzy:
We don't reflect rendering engine features in our table. Regular release and ESR have the same SpiderMonkey version. Sometimes some features could be disabled in ESR version. Actually it happened once with WebAssembly (it was disabled in 52 ESR) which is also not reflected in table. If something like this will happen in future, we can add note about this.

chicoxyzzy (approved):

Issue #661, opened 2015-10-19
Rewrite master.js


kangax:
..from 420 LOC spaggetti to something more coherent

mkoryak:
I can take a stab at this if you want. Though if i did, would you mind if i prefixed all jquery object's variable names with `$` ? I find that its a pretty good convention in jquery land.

``` js
var $table = $('#table-wrapper');
```

Also, what is the crappiest browser that you support?

ljharb:
The ES6 table needs to run in IE 9. The ES5 table should run in IE 6. The ES7 table, let's stick with IE 9 for now :-)

kangax:
I was thinking to take a stab at this in the next few days actually :) I wanna use React and render table completely dynamically from a data (which would server as a model) and table would be a plain view (re-rendered automatically when changing things with controls).

mkoryak:
react probably wont work with floatThead. floatThead caches dom elements/state internally, and things like react tend to re-create the dom from scratch, and also get upset if anything modifies the dom behind its back. 

I would be interested in trying to get it to work with it, but I doubt that it will be possible without gutting the plugin, if at all. 

On the other hand, you might not need floatThead at all with react.

mkoryak:
@ljharb I highly doubt that the table will load in IE6 without freezing the browser.

sarahkadar:
@ljharb @kangax do we still want to support IE6+? dropping that would make things a _lot_ faster.

ljharb:
For the ES5 table, absolutely; that requirement will never change.

mkoryak:
even when no one uses IE6?

ljharb:
The entire point of the compat table ES5 page is to check support in IE 6+, including shims. If we only support browsers that are ES5-compliant already, it's not very valuable.

Even if solely for historical purposes, yes, the data and working test pages should be left up permanently.

If the page doesn't work in those browsers now, we should fix that; but even then that's not justification for knowingly dropping support for them.

PR #1303, opened 2018-06-03
remove all uses of document.write

### LINKED ISSUE #1289: Remove `document.write(…)`, and never use it again
I found 3 instances of [`document.write(…)`](https://developer.mozilla.org/docs/Web/API/Document/write) that have to be removed due to this resulting in horribly slow page loads (in the range of **MINUTES**) which results [in horrible interactivity problems](https://philipwalton.com/articles/why-web-developers-need-to-care-about-interactivity/).

Even assigning to [`Element.innerHTML`](https://developer.mozilla.org/docs/Web/API/Element/innerHTML) or [`Element.outerHTML`](https://developer.mozilla.org/docs/Web/API/Element/outerHTML) is faster, despite accomplishing the same thing, as those don’t need to synchronously block parsing the web page, unlike the aforementioned `document.write(…)`.

https://github.com/kangax/compat-table/blob/03303b06c337d84e7810505898d2c154214bf059/es6/skeleton.html#L143 https://github.com/kangax/compat-table/blob/7db11bd87d2a51e7562a869d3533af01b00d5f4c/master.js#L20 https://github.com/kangax/compat-table/blob/7db11bd87d2a51e7562a869d3533af01b00d5f4c/master.js#L23

There’s also this which gets printed to the console for every invocation of the `test(…)` on every page, because it’s called from inside a table:
<pre>An unbalanced tree was written using document.write() causing
data from the network to be reparsed. For more information
<a href="https://developer.mozilla.org/docs/Web/HTML/Optimizing_your_pages_for_speculative_parsing">https://developer.mozilla.org/en/Optimizing_Your_Pages_for_Speculative_Parsing</a></pre>

Perelandric:
This commit updates the scripts that used 'document.write' to
instead use DOM creation methods.

The call that wrote a single CSS rules was removed with the rule
being added to 'master.css'.

The 'document.write' calls were always in the currently executing
script, so the 'document.currentScript' or the last script in
'document.scripts' (or its polyfill, which was added) was used to
obtain the proper element to target with the new nodes.

The original structure of the HTML was kept.

Resolves #1289

### DIFF
--- a/master.js
+++ b/master.js
@@ -29,8 +29,6 @@ window.test = function(expression) {
   var td = thisScript.parentNode;
   var tr = td.parentNode;
 
-  //document.write('<td class="' + result.toLowerCase() + ' current">' + result + '</td><td></td>');
-
   // Insert the two new TD elements after the one that contains the SCRIPT
   td = tr.insertBefore(document.createElement("td"), td.nextSibling);
   td.className = result.toLowerCase() + " current";
@@ -38,9 +36,6 @@ window.test = function(expression) {
 
   tr.insertBefore(document.createElement("td"), td.nextSibling);
 };
-
-// Moved to 'master.css'
-//document.write('<style>td:nth-of-type(2) { outline: #aaf solid 3px; }</style>');
 
 // For async tests, this returned function is used to set results
 // instead of returning true/false.
--- a/master.js
+++ b/master.js
@@ -5,6 +5,11 @@ var _gaq = [
 ];
 // jshint ignore:end
 
+
+if (!document.scripts) { // Create "live list" for all scripts if it doesn't exist
+  document.scripts = document.getElementsByTagName("script");
+}
+
 (function() {
   var ga = document.createElement('script');
   ga.type = 'text/javascript';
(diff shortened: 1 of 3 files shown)


ExE-Boss (approved):
LGTM, r+

master.js
```diff
@@ -17,10 +22,25 @@ var _gaq = [
...
   var result = (typeof expression === "string" ? expression : !!expression ? 'Yes' : 'No');
-  document.write('<td class="' + result.toLowerCase() + ' current">' + result + '</td><td></td>');
+
+  // Last SCRIPT in the document is the one currently executing
+  var thisScript = document.currentScript || document.scripts[document.scripts.length-1];
+
+  var td = thisScript.parentNode;
+  var tr = td.parentNode;
+
+  //document.write('<td class="' + result.toLowerCase() + ' current">' + result + '</td><td></td>');
```
ljharb: Should this comment be removed?

master.js
```diff
@@ -17,10 +22,25 @@ var _gaq = [
...
+  td = tr.insertBefore(document.createElement("td"), td.nextSibling);
+  td.className = result.toLowerCase() + " current";
+  td.textContent = td.innerText = result;
+
+  tr.insertBefore(document.createElement("td"), td.nextSibling);
 };
 
-document.write('<style>td:nth-of-type(2) { outline: #aaf solid 3px; }</style>');
+// Moved to 'master.css'
+//document.write('<style>td:nth-of-type(2) { outline: #aaf solid 3px; }</style>');
```
ljharb: also, is there any reason to keep this?

ljharb (commented):
Seems like a good change!

master.js
Perelandric: Not much of a reason to keep those comments. I'll go ahead and drop them.

ljharb (approved):
assuming it works in all browsers, including down to ES3 if possible

Perelandric:
I just tested the *existing* es5 compatibility table using IE6 in a VM, and it looks pretty much broken, which shouldn't be the case according to https://github.com/kangax/compat-table/issues/661.

I'm certain the code I wrote is ES3 compliant and I'm 99% certain it's compatible with IE6's DOM, but wouldn't be able to test it until the existing site is fixed.

PR #1304, opened 2018-06-03
`position: sticky;` is supported correctly in Firefox since version 32

### LINKED ISSUE #1139: Sticky thead works only in Safari
The current implementation of the position:sticky workaround (on master.js) works only in Safari:
```
  if (thead.css('position') === "-webkit-sticky") {
    // Remove floatThead when native position:sticky is usable.
    // Currently, only Safari (which floatThead incidentally does not support),
    // and its -webkit-sticky prefix, works correctly with <thead>.
    $.fn.floatThead = function() { return this };
  }
```
It should check also for the actual "sticky" value, which is implemented on Edge 16 preview and will possibly be implemented in other browsers in the future.

ExE-Boss:
See the MDN article [on the CSS `position` property](https://developer.mozilla.org/docs/Web/CSS/position#Browser_compatibility) for details.

Fixes #1139

### DIFF
--- a/master.js
+++ b/master.js
@@ -45,17 +45,16 @@ $(function() {
   var table = $('#table-wrapper');
   var currentBrowserSelector = ":nth-of-type(2)";
   var thead = $('thead');
-  
-  if (thead.css('position') === "-webkit-sticky") {
-    // Remove floatThead when native position:sticky is usable.
-    // Currently, only Safari (which floatThead incidentally does not support),
-    // and its -webkit-sticky prefix, works correctly with <thead>.
-    $.fn.floatThead = function() { return this };
-  }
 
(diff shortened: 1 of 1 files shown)


ljharb (approved):
Seems fine. Not sure why the block needed to move orderings tho.

ExE-Boss:
> Not sure why the block needed to move orderings tho.

That’s needed because of the Chrome issue.

afmenez:
I have tried this before on #1139.  Is the order change enough to work around the bugs on both Chrome and Firefox?

ExE-Boss:
From my own testing, it works perfectly on Firefox 60 without any issues (although I didn’t try it without #1303).

chicoxyzzy (approved):

PR #1306, opened 2018-06-04
remove invocations of .call and .apply with a host object

Perelandric:
Old IE would not accept host objects in the 'thisArg' position
of Function.prototype.call and ...apply. This commit uses toArray
instead.

One call passed a jQuery object, which should work, but it too
was updated for the sake of consistency.

### DIFF
--- a/master.js
+++ b/master.js
@@ -27,11 +27,15 @@ document.write('<style>td:nth-of-type(2) { outline: #aaf solid 3px; }</style>');
 function makeAsyncPassedFn(className, textContent) {
   return function(rowNum) {
     return function() {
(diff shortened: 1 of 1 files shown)


Perelandric:
I'm going to poke through to see if I can find anything else that would be problematic in old IE, so let's keep this open for a week or so while I add to it.

ljharb (approved):

Perelandric:
Actually, aside from a little CSS incompatibility for IE6, I don't see much else right now. I'd say go ahead whenever you folks are ready, and if I spot something in the meantime, I'll add it and make sure you know.

chicoxyzzy (approved):

PR #1308, opened 2018-06-04
use scheme-less url when fetching jquery

Perelandric:
IE6 seems to have trouble loading jquery from the CDN with 'https'
when the page is 'http'. This changes the 'src' to a scheme-less
url so that it properly loads.

### DIFF
--- a/es5/index.html
+++ b/es5/index.html
@@ -7,7 +7,7 @@
     <title>ECMAScript 5 compatibility table</title>
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,600">
     <link rel="stylesheet" href="../master.css">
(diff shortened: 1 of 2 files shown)


Perelandric:
Don't know if this is the preferred approach to fixing this. Let me know if something else is preferred.

ljharb (approved):

chicoxyzzy (approved):
LGTM

afmenez:
If it works the same, shouldn't this be changed on all the tabs, for consistency?

ljharb:
Yes.

ljharb:
However, separately, why is the page ever http?

Perelandric:
I figured it would be better to be explicitly 'https' in most cases. Since the other tables aren't concerned with IE6 support, I left those alone.

I think there may be some issues with IE6 and encrypted sites, so compatibility may necessitate 'http'.

Issue #1217, opened 2017-11-11
Don't test Uint8ClampedArray in %TypedArray% tests


rakudrama:
IE11 shows 'No' for many %TypedArray% features but the failure is due to missing Uint8ClampedArray, and not the missing feature on other typed arrays.

ljharb:
Then IE 11 doesn't fully support TypedArrays.

I'd support splitting the tests up into subtests when applicable, but < 100% is the same as 0% when asking "do you support this feature".

rakudrama:
I understand what you are saying, but I'm not finding the table helpful for the following situation.
Can I replace the following code that copies part of the array and works on IE 11

```js
new Uint8Array(myUint8Array.subarray(1,5))
```

with the following:

```js
 myUnit8Array.slice(1,5)
```

The table does not help me make that determination.

Is there something to be done to give a better reflection of the partial state of support?

ljharb:
Have you tried clicking the triangle to expand the "typed arrays" section?

rakudrama:
Oh yes. The entry for %TypedArray%.prototype.subarray x IE11 says "No".
It says "No" because the (C) snippet crashes on the subexpression Uint8ClampedArray.prototype.subarray.
That doesn't _quite_ capture that I can use subarray on other typed arrays.

ljharb:
Fair enough. We'd be happy to accept a PR that splits up the `subarray` tests so that it's more clear what works and what doesn't on IE 11!

PR #1312, opened 2018-06-11
Update ES6 results for IE11

afmenez:
After the latest fixes from @Perelandric , it is now possible to run all tests with IE11 (the table layout is still broken, but at least the results are shown).  This patch fixes all results that were wrong.


zloirock:
`Uint8ClampedArray` has been introduced in the build [`KB2929437`](https://support.microsoft.com/en-us/help/2929437/description-of-the-security-update-for-internet-explorer-11-on-windows), it's missing in early IE11 versions. So, at least, first 2 results should not be added.

ljharb (approved):

data-es6.js
```diff
@@ -5323,6 +5323,11 @@ exports.tests = [
         typescript1: typescript.corejs,
         firefox2: false,
         firefox4: true,
+        ie11: {
+          val: true,
```
zloirock, line 6328: Rather `false`. Also, `.subarray` test depends on `Uint8ClampedArray`.

afmenez: Is this value supposed to reflect initial release or the version that people are probably using?  I believe it is the later.

afmenez: Do you mean the note should be added to .subarray too?  Wouldn't that be confusing?  Maybe the test should be changed to avoid the dependency.

ljharb: For IE, it’s the latest in that major line.

zloirock: > Is this value supposed to reflect initial release or the version that people are probably using? I believe it is the later.

Just an example: we will set is as `true`. `babel-preset-env` will load this result. For IE11 on `Uint8ClampedArray` `preset-env` with `useBuiltIns` option will not add a polyfill. But a serious part of users has un-updated versions of IE11.

> Do you mean the note should be added to .subarray too?

I don't care, but with current tests it should be `false`.

> Maybe the test should be changed to avoid the dependency.

IIRC it's already was proposed in another issues.

ljharb: Since the compat table assumes all IE users are on the latest of their major line, that sounds like a flaw in preset-env.

zloirock: @ljharb this approach should work for compilers and unstable releases. For stable engines, we should use data for initial releases, otherwise, developers could have similar problems even without `preset-env`.

ljharb: Then we’d need to keep data for multiple ie 11 builds, but also, preset-env would need to provide a way for uses to specify that. Do they?

zloirock: Makes sense, but I'm not sure that we really need these difficulties because of this. Anyway, at first, it should be solved on `babel` side.

afmenez: So, it is not clear to me what should be changed (if anything).

chicoxyzzy: @zloirock is there an issue in Babel repo?

also for reference https://github.com/kangax/compat-table/issues/1217

afmenez: Hum, I see in #1217 that the issue is way more complicated than this patch can handle (by this patch I mean me).  I'll change both results to false and leave the comment until the tests are fixed.  
I'll keep the `var` change.  Is this OK?

ljharb: If it passes in the latest IE 11, it should be true.

zloirock: @chicoxyzzy seems it will be an issue when `babel` will update `compat-table` version.

zloirock: @afmenez as @ljharb proposed, you can add new positive results as a column for another IE11 build.

ljharb: Then babel can file a separate issue about it; the compat table doesn’t currently provide more than one set of results for each IE major, at “latest”.

afmenez: I believe adding another IE11 column would be confusing, as most users don't know about these "sub-versions".  I'll keep it as false with a note that both tests may pass later builds.

ljharb: Again, it should be true, not false.

zloirock: Again, it should be `false`, not `true` -) Just adding comments, like @afmenez wrote, should work for the first time, later we can add a new column for modern builds with positive results.

ljharb: The compat table has never listed non-latest IE builds - that is the standard we must follow.

Any bugs arising from that should be dealt with separately.

zloirock: > that is the standard we must follow

It was not a standard and I strictly disagree. See the reason above.

If someone wanna support IE he should support early IE builds because a serious part of users disabled auto updates -)

ljharb: That’s not the purpose of the compat table. IE 8, for example, only supports the JSON methods in a later build, but the table shows it as passing.

If you disagree that’s how it should work, that’s fine, but that disagreement is irrelevant to the fact that this is how the table operates, and that will not change in this unrelated PR.

zloirock: > That’s not the purpose of the compat table.

Really? Providing actual data about supported features is the main purpose of the compat table and we should not show a false positive -)

> IE 8, for example, only supports the JSON methods in a later build, but the table shows it as passing.

But the issue from this PR is known from 2014 and the table shows it as not passing. I do not care about the results of `JSON`. I don't see anything related your position in the documentation.

> If you disagree that’s how it should work, that’s fine, but that disagreement is irrelevant to the fact that this is how the table operates, and that will not change in this unrelated PR.

This PR just should not be added in this form.

Adding a new column with positive results makes sense.

afmenez: Updated the patch according to @ljharb orientation.

zloirock (changes requested):
Fix false positive for early IE11 versions.

data-es6.js
ljharb: Whether you care about the result of json or not is also irrelevant. It shows, clearly, that the table has never shown results from non-latest IE major versions. That shouldn’t change here. The bug is that this result sat at false for so long.

zloirock: One more time: if you wanna positive results for IE11 - feel free to add a new column. Changing results of the actual column will break `preset-env` and confuse developers who wanna support all IE11 versions. This precedent could cause much more serious problems in the future.

chicoxyzzy (approved):
I think that notes cover all necessary info so I'll approve

data-es6.js
ljharb: The current column represents "IE 11 latest". If *you* want negative results for non-latest IE 11, then you're welcome to add a column.

zloirock: Me? -) Interesting position. The current column is just [IE11 with release date 2013-10-17](https://github.com/kangax/compat-table/blob/2c10d2d4fa6abee6aae21ef58fac299de6716249/environments.json#L393). It's definitely not "IE 11 latest".

chicoxyzzy: `preset-env` could override results. Also ideally it should use something what fits better for it than `compat-table`, which purpose never was to provide data for `preset-env` or anything else.

zloirock: > `preset-env` could override results.

It's additional special cases. Why?

> Also ideally it should use something what fits better for it than `compat-table`

Definitely. But currently we haven't any good alternative.

chicoxyzzy:
@zloirock are you ok with latest changes?

zloirock:
@chicoxyzzy nope, see my argumentation above - nothing was changed.

data-es6.js
ljharb: a) the release date is for the first version, results always reflect the latest
b) supporting preset-env is ancillary and NOT something that this repo needs to cleave to, especially not when it changes the meaning this project's test columns have had historically.

chicoxyzzy:
Is everyone ok if we'll show these features as unsupported and will add a note that in latest versions of IE these features was added?

ljharb:
No, i want them to show as supported if they’re available in the latest version, per precedent.

mgol:
This PR has been opened a year ago. Can we get a decision here so that it can be landed in whatever version gets accepted?

chicoxyzzy (approved):
LGTM

zloirock:
I wrote my objections above. I could accept only `false` with a note or adding a new version. I don't wanna break the web.

ljharb:
It won't break the web; the columns mean "latest version". Anyone not using the latest version of the thing in the column is *already* broken if they're relying on this data.

zloirock:
`compat-table` is a data source not only about the latest versions of things. [As I wrote above](https://github.com/kangax/compat-table/pull/1312#discussion_r194834967), the current column is just IE11 with release date 2013-10-17. And many people relying on this data as on data about all IE11 versions.

ljharb:
Then it sounds like we should add a new column for the latest IE 11, and mark the current one as very obsolete.

zloirock:
I agree. But just "obsolete" could be better.

afmenez:
I don't believe it is productive to add columns for minor versions.  How about changing the results to "flagged", keeping the notes on the current PR?

chicoxyzzy:
> compat-table is a data source not only about the latest versions of things. As I wrote above, the current column is just IE11 with release date 2013-10-17. And many people relying on this data as on data about all IE11 versions.

So we should create issue / PR against `babel-present-env` to override this as it overrides other things. If any tool uses compat-table data as is, than it's probably already broken

chicoxyzzy:
Everywhere else we indicate latest status. If not, we'd  need to mark a bunch of Safari, Chrome and Firefox features as `false` because of temporary bugs in minor versions and patches.

ljharb:
^ ftr, that matches my intuition as well as babel-preset-env reality, and is why I think @zloirock's position here is objectively wrong.

zloirock:
@chicoxyzzy yes, `compat-table` contains incomplete data, but it would be better just improve this data instead of making it even more broken and required even more special cases. It's not a problem for me since for `core-js@3` I made `core-js-compat`, but it will cause problems in many other cases.

The case from this PR is adding new features, not fixing bugs, so it's a semver minor, not patch. So incorrectly compare it with other browsers, the issue here in a versioning / naming MS browsers.

ljharb:
The best thing to do is take broken use cases - like “not the latest version of IE 11” and make them more broken, precisely so the users wrongly relying on this nonexistent property have more incentive to stop doing so.

It’s unreasonable to insist on maintaining a property that has never existed and does not currently exist.

chicoxyzzy:
I still think that we should merge this.

@existentialism @nicolo-ribaudo @JLHwung how hard it would be to handle this change in `babel-preset-env` if we'll merge this?

nicolo-ribaudo:
It wouldn't be too hard, we can hard code the old data.

ljharb:
Typically, overriding a block isn't something I'd ever support; however there's extenuating circumstances here, and we don't want to have to wait a year or more, so if @chicoxyzzy agrees, I'm ok with that.

data-es6.js
```diff
@@ -6324,9 +6324,14 @@ exports.tests = [
       res: {
         babel6corejs2: babel.corejs,
         typescript1corejs2: typescript.corejs,
+        ie11: {
+          val: true,
+          note_id: "ie-clampedarray",
+          note_html: "A partial implementation of <code>Uint8ClampedArray</code> has been introduced with <a href='https://support.microsoft.com/en-us/help/2929437/description-of-the-security-update-for-internet-explorer-11-on-windows'>KB2929437</a>, so this test fails on early IE11 builds."
```
chicoxyzzy: Can you elaborate what does "partial implementation" mean? It would be great if we can add failing tests for those cases.

afmenez: To be honest, this was so long ago that I don´t remember what I meant.  But I believe that a better description wold be "A buggy implementation shipped with IE10 and early releases of IE11, and it was fixed with the PR".  Would this be enough?
I don't know enough to write a new subtest, but I believe this one already fails on those cases...

chicoxyzzy:
```suggestion
          note_html: "A buggy implementation of <code>Uint8ClampedArray</code> has been fixed with patch <a href='https://support.microsoft.com/en-us/help/2929437/description-of-the-security-update-for-internet-explorer-11-on-windows'>KB2929437</a>"
```

afmenez: Done.

PR #1313, opened 2018-06-11
Update to jQuery 1.12

afmenez:
Updating the jQuery to version 1.12.4.  It is supposed to be fully compatible with the version 1.11.1, which is currently linked.
The change logs mention many IE corrections, but I didn't notice any improvement.  What I did notice was faster initial rendering on Firefox, specially on the ES6 tab.

### DIFF
--- a/es2016plus/index.html
+++ b/es2016plus/index.html
@@ -7,7 +7,7 @@
     <title>ECMAScript 2016+ compatibility table</title>
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:300,600">
     <link rel="stylesheet" href="../master.css">
-    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
     <script src="../master.js"></script>
     <script>
       var __script_executed = {};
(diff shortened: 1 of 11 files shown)


ljharb:
Any reason not to update to a later version?

ljharb (approved):

afmenez:
Both jQuery 1.12.x and 2.x were maintained simultaneously, so there would be no security gains updating to 2.x.  But there are reasons not to upgrade to 2.x:
1. jQuery 2.0 drops compatibility with IE6...IE8;
2. We don't use any of the newer features on 2.x;

OTH there are no breaking changes on jQuery 2.x, so we probably can update to 2.2.3 if we can leave IE6..IE8 behind (I personally don't think so, since we still don't consider them old enough to remove the IE8 results from the tables, and there was recent activity by @Perelandric for improving IE6 compatibility of the tables).  Maybe we can compromise and link to 1.12 on "es5", "es-intl" and "non-standard"; and link to 2.2.x on the newer tabs?

We can even update to jQuery 3.3.1 if someone is willing to test everything with the newer version, as there are some breaking changes in the API.

ljharb:
With jquery 2, i believe there’s a compat plugin that should still work with old IE.

afmenez:
As far as I know the compat plugins are for jQuery API compatibility, not old browser support.

ljharb:
ah, you're right. let's stick to v1 then.

chicoxyzzy (approved):

afmenez:
OK, it is done then.  There are two approvals already, what is missing to merge this one?

afmenez:
Thanks!

Issue #919, opened 2016-09-22
Update TypeScript to 2.0


zloirock:
https://blogs.msdn.microsoft.com/typescript/2016/09/22/announcing-typescript-2-0/

chicoxyzzy:
Should we add TS 2.0 as a new row in a table or updating current row is enough?

ljharb:
I think a new row is better - people won't update immediately, and the historical data is useful.

dead-claudia:
I also think a new row would be better, for the same reasons. It would also fit in line with much of the rest, where the compilers are more of the exception, not the rule, in lacking versions (TypeScript is the only one that has really changed in its support across versions).

I will note that TypeScript does now offer ES2015 through ES2017 as compilation targets, with some features not transpilable down to earlier standards (e.g. async functions were only recently opened up to ES5 and ES6 targets), so I don't know if that would qualify as transpiler support or not.

chicoxyzzy:
cc @DanielRosenwasser who may be interested =)

DanielRosenwasser:
Well I'll point out that the data already reflects 2.0 (trailing commas in arg lists seems green right now).

To put my two cents in here, I don't entirely see why you'd want a new column. The reason browsers have multiple versions is to let developers know when it makes sense to start using certain features because they don't know which browsers their clients might be running. Is this as much a concern with a compiler which your team fixes on a certain version?

That said, it's really whatever you all think is most appropriate. You'd eventually have to do the same thing with Babel and other compilers too though.

ljharb:
@DanielRosenwasser the table is also used as a reference for people who may not be using the newest one.

dead-claudia:
It's also useful as a historical reference. Seeing the progression with browsers, even to ES5, is rather interesting IMHO.

devoto13:
From the implementation point of view it's problematic to have multiple versions of TypeScript (or any library) as a dependency. Does it mean that results from previous versions should be somehow "archived" when adding a new version of compiler?

dead-claudia:
@devoto13 Are browser results cached already?

devoto13:
@isiahmeadows Yes, they are. But the difference is that browser results are always entered manually, while transpilers results are computed automatically. Not saying that it's undoable, just that it will introduce more complexity.

martinsuchan:
TypeScript 2.1 with support for Async Functions and Object Rest & Spread was just released:
[...]

PR #1315, opened 2018-06-15
update TypeScript results with 'downlevelIteration' option

trotyl:
Following https://github.com/kangax/compat-table/issues/919#issuecomment-396485252

Env:

+ TypeScript 3.0.1 (last rebase);
+ IE 9 in VirtualBox;

Note: 

+ Removed note `typescript.asyncawait` as generator is already supported under this flag;
+ Some `eval()` tests in es2016+ async function are left to be `null` (not sure how to test it without compilation);
+ There's a special case (`spreading non-iterables is a runtime error`) works before (accidentally) but not with this flag (not sure how to mark it);

Off-topic question:

Some subclassing tests are also passed (not related to this flag), is there any other requirements for transpilers? (Seems all of them are marked unsupported in current table)

image


data-common.json
```diff
@@ -27,10 +27,15 @@
...
-    "asyncawait": {
+    "downlevelIteration": {
       "val": true,
-      "note_id": "typescript-async-await",
-      "note_html": "TypeScript <code>async</code> / <code>await</code> requires native generators support."
+      "note_id": "typescript-downlevel-iteration",
+      "note_html": "Requires the <code>downlevelIteration</code> compile option."
+    },
+    "notDownlevelIteration": {
+      "val": true,
```
zloirock: I don't care should this value be `true` with a note or `false` with a note. Let's listen to the opinions of the rest maintainers.

ljharb: I would expect false; trues in compilers are for default unflagged things.

trotyl: Done.

chicoxyzzy (approved):

ljharb (approved):

ljharb:
This needs a rebase and a regen of the test data.

trotyl:
Rebase done. (Re-checked on TypeScript 3.0.1, no data change)

chicoxyzzy:
Thank you!

PR #1317, opened 2018-06-18
Make the test for Misc > Proxy get > RegExp#flags less fragile.

### LINKED ISSUE #1208: RegExp.prototype.flags test failing in recent Node versions
The test isn't dynamic enough to account for new flags.

```js
function(){
  // RegExp.prototype.flags -> Get -> [[Get]]
  var get = [];
  var p = new Proxy({}, { get: function(o, k) { get.push(k); return o[k]; }});
  Object.getOwnPropertyDescriptor(RegExp.prototype, 'flags').get.call(p);
  return get + '' === "global,ignoreCase,multiline,unicode,sticky";
}
```

In Node with `--harmony` flag or Node's Nightly- `get` has additional values and yields:
>  global,ignoreCase,multiline,dotAll,unicode,sticky

Initially reported here:
https://github.com/nodejs/node/issues/16449

benblank:
Rather than comparing an array by converting it to a string, tests the array for inclusion of well-known elements.

This allows correct detection of the feature on recent versions of Node (resolves #1208) and Chrome.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -9649,9 +9649,9 @@ exports.tests = [
         var get = [];
         var p = new Proxy({}, { get: function(o, k) { get.push(k); return o[k]; }});
         Object.getOwnPropertyDescriptor(RegExp.prototype, 'flags').get.call(p);
-        return get.indexOf('global') !== -1 &&
-               get.indexOf('ignoreCase') !== -1 &&
-               get.indexOf('multiline') !== -1;
+        return get.indexOf('global') === 0 &&
+               get.indexOf('ignoreCase') === 1 &&
+               get.indexOf('multiline') === 2;
       */},
       res: {
         firefox2: false,
--- a/data-es6.js
+++ b/data-es6.js
@@ -9646,12 +9646,22 @@ exports.tests = [
       name: 'RegExp.prototype.flags',
       exec: function() {/*
         // RegExp.prototype.flags -> Get -> [[Get]]
-        var get = [];
-        var p = new Proxy({}, { get: function(o, k) { get.push(k); return o[k]; }});
+        var expected = [];
+        // Sorted alphabetically by shortname – "gumsuy".
+        if ('global' in RegExp.prototype) expected.push('global');
+        if ('ignoreCase' in RegExp.prototype) expected.push('ignoreCase');
+        if ('multiline' in RegExp.prototype) expected.push('multiline');
+        if ('dotAll' in RegExp.prototype) expected.push('dotAll');
+        if ('unicode' in RegExp.prototype) expected.push('unicode');
+        if ('sticky' in RegExp.prototype) expected.push('sticky');
+        var actual = [];
+        var p = new Proxy({}, { get: function(o, k) { actual.push(k); return o[k]; }});
         Object.getOwnPropertyDescriptor(RegExp.prototype, 'flags').get.call(p);
-        return get.indexOf('global') === 0 &&
-               get.indexOf('ignoreCase') === 1 &&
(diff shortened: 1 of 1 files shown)


data-es6.js
```diff
@@ -9649,7 +9649,9 @@ exports.tests = [
         var get = [];
         var p = new Proxy({}, { get: function(o, k) { get.push(k); return o[k]; }});
         Object.getOwnPropertyDescriptor(RegExp.prototype, 'flags').get.call(p);
-        return get + '' === "global,ignoreCase,multiline,unicode,sticky";
+        return get.indexOf('global') !== -1 &&
+               get.indexOf('ignoreCase') !== -1 &&
+               get.indexOf('multiline') !== -1;
```
ljharb: this now no longer tests for sticky or unicode, and also no longer tests the order - both of which are important.

ljharb (changes requested):

data-es6.js
benblank: Unicode and sticky are both ES6 extensions and not part of the feature being tested.

I wasn't aware that order was significant; I'll update the PR.

ljharb: This is a test for flags; it requires testing all flags.

benblank: This is a test for "Proxy, internal 'get' calls" and how they act on `RegExp.prototype.flags`, not for completeness of the implementation's RegExp support.

If it should not fail when `dotAll` is present (which is the bug in issue #1208), then why should it fail when `unicode` and `sticky` are absent?

ljharb: I think this test can be written such that it does not fail when dotAll is present or not.

data-es6.js
```diff
@@ -9649,7 +9649,9 @@ exports.tests = [
         var get = [];
         var p = new Proxy({}, { get: function(o, k) { get.push(k); return o[k]; }});
         Object.getOwnPropertyDescriptor(RegExp.prototype, 'flags').get.call(p);
-        return get + '' === "global,ignoreCase,multiline,unicode,sticky";
+        return get.indexOf('global') === 0 &&
+               get.indexOf('ignoreCase') === 1 &&
```
ljharb: My suggestion:
```js
var global = get.indexOf(‘global’);
var ignoreCase = get.indexOf(‘ignoreCase’);
var multiline = get.indexOf(‘multiline’);
var unicode = get.indexOf(‘unicode’);
var sticky = get.indexOf(‘sticky’);
return -1 < // this is weird indentation ofc, so indent however
  global && global <
  ignoreCase && ignoreCase <
  multiline && multiline <
  unicode && unicode <
  sticky;
```

benblank: But why fail at all when `unicode` and `sticky` are absent?  That's not what's being tested here.

ljharb: If the engine supports those flags, and the proxy isn’t able to intercept them, then the proxy isn’t functioning correctly.

benblank: The same could be said for `dotAll`, which your suggestion doesn't cover.  Let me put together a version of the test with feature detection and I'll update the PR.

ljharb: that's true, but this is the ES6 page, so it's fine if dotAll is missing (altho it'd be fine, if present, to ensure it's in the proper order)

benblank: Okay, the new version checks whether each flag exists before expecting it to be hit by `#flags`.  Yeah, it's a bit silly to check for the ES3 flags, but it's makes the code more consistent and easier to modify if new flags are ever added.

ljharb (approved):
Thanks, this looks great!

chicoxyzzy (approved):
LGTM!

PR #1319, opened 2018-06-25
Firefox 61 released

afmenez:
It will actually be released tomorrow, but here's the early patch.
No surprises, just updating release dates, adding a new version and retiring ancient ones.


ljharb (approved):

afmenez:
Firefox 61 is actually out now.

PR #1320, opened 2018-06-28
SIMD is no longer in JavaScript (moved to WebAssembly).

brianblakely:
Per TC39 documentation: https://github.com/tc39/ecmascript_simd


ljharb (changes requested):
The proposal is inactive, but the feature still exists. Instead of deleting it, please move it to nonstandard.

data-non-standard.js
```diff
@@ -208,7 +1398,7 @@ exports.tests = [
   exec: function () {
     return (function () {}).arity === 0 &&
       (function (x) { return x; }).arity === 1 &&
-      (function (x, y) { return y; }).arity === 2;
+      (function (y) { return y; }).arity === 2;
```
ljharb: this is incorrect; the arity should be 2 only when there's 2 arguments.

ljharb (changes requested):

chicoxyzzy (changes requested):
yes, please move it to non-standard page

chicoxyzzy (commented):
Oops. Now I see you did it already. Please make changes according to @ljharb's comments

chicoxyzzy (approved):
Oh. I should work little bit less to prevent burnout :)
LGTM

data-non-standard.js
```diff
@@ -208,7 +1398,7 @@ exports.tests = [
   exec: function () {
     return (function () {}).arity === 0 &&
       (function (x) { return x; }).arity === 1 &&
-      (function (x, y) { return y; }).arity === 2;
```
ljharb: this function is fine continuing to return `y`; i'm not sure why this line needs a change at all. can you revert it?

ljharb (changes requested):
One more change, and it'd be great if you could rebase down to one commit :-)

ljharb (approved):

PR #1323, opened 2018-07-06
Add results for GraalVM JavaScript

wirthi:
Hi,

GraalVM (https://www.graalvm.org) is coming with JavaScript (and Node.js) support. This PR adds results for this engine.

GraalVM JavaScript is based on the Graal Java JIT compiler (https://github.com/oracle/graal). The source code is available at https://github.com/graalvm/graaljs

Supporting Node.js, GraalVM can reuse the `node.js` file to generate the results. The commandline to reproduce is `/path/to/graalvm-ce-1.0.0-rc3/bin/node --jvm --shared-array-buffer --intl-402 node.js`. The results are based on GraalVM CE 1.0.0 RC3 (https://github.com/oracle/graal/releases/tag/vm-1.0.0-rc3).

-- Christian

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
     "pretest": "npm run lint",
     "build:compilers": "node build.js compilers",
     "git:clean": "if type git > /dev/null 2>&1; then git stash save --keep-index; fi",
-    "test:no_change": "npm run --silent git:clean && if npm run --silent build | grep 'Write to file'; then echo 'Please run `npm run build` and commit the changes.' && exit 1; fi"
+    "test:no_change": "npm run --silent git:clean && if npm run --silent build | grep 'Write to file'; then echo 'Please run `npm run build` and commit the changes.' && exit 1; fi",
+    "graalvm": "node --jvm --shared-array-buffer --intl-402 node.js"
   },
   "config": {
     "unsafe-perm": true
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -39,6 +39,7 @@ exports.tests = [
           chrome52: true,
           safari10_1: true,
           duktape2_0: true,
+          graalvm: true,
         }
       },
       {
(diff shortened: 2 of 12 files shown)


ljharb:
(the link has an SSL error btw)

ljharb:
LGTM - could you add npm run-script(s) so that the commands to generate the results (and install the needed deps) are discoverable in the repo?

wirthi:
Thanks, fixed the link.

What run scripts do you expect exactly? GraalVM is compatible with Node.js, so all the pieces are in the repo already - you can use the `node.js` file to generate the results with GraalVM as described above.

ljharb:
like, `npm run graal` or something :-) just like https://github.com/kangax/compat-table/blob/gh-pages/package.json#L43

wirthi:
ah, ok, gotcha, will add.

wirthi:
I've added a script so you can do `npm run graalvm`. Note however that this assumes that GraalVM is on your path before Node.js. If this is not the case, you need to explicitly call GraalVM's `node` command as described above.

package.json
```diff
@@ -42,7 +42,8 @@
     "pretest": "npm run lint",
     "build:compilers": "node build.js compilers",
     "git:clean": "if type git > /dev/null 2>&1; then git stash save --keep-index; fi",
-    "test:no_change": "npm run --silent git:clean && if npm run --silent build | grep 'Write to file'; then echo 'Please run `npm run build` and commit the changes.' && exit 1; fi"
+    "test:no_change": "npm run --silent git:clean && if npm run --silent build | grep 'Write to file'; then echo 'Please run `npm run build` and commit the changes.' && exit 1; fi",
+    "graalvm": "node --jvm --shared-array-buffer --intl-402 node.js"
```
ljharb, line 46: will this work in any version of node, or does it have to be special compiled?

ljharb (approved):

pmlopes:
I think the results should state that this was run on the nodejs mode, not embedded mode.

wirthi:
I have added such a comment.

chicoxyzzy (approved):
Thank you, looks good

PR #1331, opened 2018-07-31
Rename spread operator to spread syntax

chicoxyzzy:
Technically there is no such thing as spread operator in specification and spread is part of grammar and runtime semantics. So I think it's better to call it spread syntax as [in MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -1770,7 +1770,7 @@ exports.tests = [
   ],
 },
 {
-  name: 'spread (...) operator',
+  name: 'spread syntax for iterable objects',
   category: 'syntax',
   significance: 'large',
   spec: 'http://www.ecma-international.org/ecma-262/6.0/#sec-argument-lists-runtime-semantics-argumentlistevaluation',


data-es6.js
```diff
@@ -1770,7 +1770,7 @@ exports.tests = [
   ],
 },
 {
-  name: 'spread (...) operator',
+  name: 'spread syntax',
```
ljharb: I'm not convinced here; colloquially this is known as the spread operator.

chicoxyzzy: I've tried to explain this here 
https://twitter.com/chicoxyzzy/status/1024606477963681794

Also see https://stackoverflow.com/a/44934830/1722762

ljharb (approved):

data-es6.js
ljharb: altho this should somehow denote that it’s not object spread, but iterable spread

chicoxyzzy: I'm not sure I understand what do you mean. Do you mean that we also should reword object spread from ES2016+ tab somehow?

chicoxyzzy: Reworded. Please take a look

ljharb (approved):

PR #1340, opened 2018-09-04
Chrome 69 release

chicoxyzzy:


environments.json
```diff
@@ -1753,33 +1753,46 @@
     "family": "V8",
     "short": "CH&nbsp;66,<br>OP&nbsp;53",
     "release": "2018-04-17",
-    "retire": "2018-05-31",
+    "retire": "2018-05-29",
```
afmenez, line 1756: Google Chrome 67 for Android was released on May 31...

environments.json
```diff
@@ -1753,33 +1753,46 @@
...
-    "obsolete": false
+    "retire": "2018-07-24",
+    "obsolete": true
   },
   "chrome68": {
     "full": "Chrome",
     "family": "V8",
     "short": "CH&nbsp;68,<br>OP&nbsp;55",
     "release": "2018-07-24",
+    "retire": "2018-09-04",
```
afmenez, line 1772: Google Chrome for Android was not released yet...

environments.json
```diff
@@ -1753,33 +1753,46 @@
...
     "obsolete": false
   },
   "chrome69": {
-    "full": "Chrome Dev",
+    "full": "Chrome",
     "family": "V8",
     "short": "CH&nbsp;69,<br>OP&nbsp;56",
-    "unstable": true
+    "release": "2018-09-04",
+    "retire": "2018-10-16",
```
afmenez, line 1780: I believe it is to early to set this retire date...

environments.json
```diff
@@ -1753,33 +1753,46 @@
...
+    "obsolete": false
   },
   "chrome70": {
-    "full": "Chrome Canary",
+    "full": "Chrome Dev",
     "family": "V8",
     "short": "CH&nbsp;70,<br>OP&nbsp;57",
+    "release": "2018-10-16",
+    "retire": "2018-12-04",
```
afmenez, line 1788: I believe it is to early to set this retire date...

afmenez (changes requested):
I would hold back on the "retire" changes, but fine otherwise.

ljharb:
agreed; retire dates shouldn't be added until they are passed.

chicoxyzzy:
@afmenez I use release dates from official https://omahaproxy.appspot.com and https://www.chromium.org/developers/calendar. Not sure we should use Android version release dates when versions for all other platforms are the same.
Also I'm confused. Wasn't it too early when you set retire dates for Firefox Nightly 63 [back in June](https://github.com/kangax/compat-table/commit/95f4f84dde1303374eea42920743e2bce1650601#diff-ca0f8d6af0c57ea63bddc9086afc0bc2R976)?

Any reasons why not to set retire dates when they are already known and we know that version is not obsolete?

environments.json
chicoxyzzy: According to Google calendar release date is correct (also it didn't change in this PR)

environments.json
chicoxyzzy: Chrome 68 for Andriod was released in July

afmenez:
I am not disputing the dates, I am just arguing that the first platform to be released does not obsolete the other platforms.
The reason I used to avoid setting future retire dates for Chrome is related, their usually take some days to release all versions, Mozilla releases them all at once.  But on a second thought, they are all estimates anyway, and can bi fixed later if necessary.

chicoxyzzy:
I'll go ahead and merge this as retire date ETAs are known and provided by Chrome team and everyone seems to be OK with other changes. Chrome 69 was released so we should reflect this in table for developers and tools which use compat-table as data source

PR #1356, opened 2018-10-05
Add globalThis proposal, move global to non-standard

chicoxyzzy:
Chrome 70 beta supports globalThis feature so we should update table 

@zloirock does core-js support globalThis?


ljharb (approved):

zloirock:
@chicoxyzzy it's [renamed](https://github.com/zloirock/core-js/commit/3311d50670bcb2fbe9ea49fd9fa743257e08760b) and will be published in the next `core-js@3` release. Now you can mark it as `false`.

chicoxyzzy:
I unchecked core-js support and added Node to non-standard page

zloirock (approved):

PR #1375, opened 2018-11-26
Test the regexp "u" flag for non-BMP Unicode characters on ES6

### LINKED ISSUE #1322: RegExp "u" flag anomaly in Safari
I encountered a problem when testing examples from [You-Dont-Know-JS](https://github.com/getify/You-Dont-Know-JS/blob/master/es6%20%26%20beyond/ch2.md#unicode-flag): 

while 
```js
/^.-clef/u.test( "𝄞-clef" );		// true
```
returned `true` as expected in Chrome and Firefox, it returned `false` in Safari.

Safari is marked as 'passing' for this feature in _compat-table_, based on the following test:
```js
"return 𠮷".match(/^.$/u)[0].length === 2;
```

While Safari 'passes' with this test, it fails to return `3` for the length for the following expression (because it doesn't find a match):
```js
"𠮷x".match(/^.x$/u)[0].length
```
Both Chrome and Firefox return the expected length.

I'm wondering whether the test should be improved and Safari should be marked as 'partially passing'.
BTW, Safari Technology Preview Release 59 (Safari 12.0, WebKit 13606.1.21) also fails.

afmenez:
Fixes #1322.
Can someone confirm the Safari results?

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -5536,6 +5536,29 @@ exports.tests = [
       },
     },
     {
+      name: '"u" flag, non-BMP Unicode characters',
+      exec: function() {/*
+        return "𠮷x".match(/^.x$/u)[0].length === 3;
+      */},
+      res: {
+        tr: true,
+        babel6: true,
+        typescript1: typescript.fallthrough,
+        edge12: true,
+        edge13: true,
+        firefox2: false,
+        firefox46: true,
+        opera10_50: false,
+        chrome50: true,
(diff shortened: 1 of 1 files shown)


ljharb (approved):
Thanks!

chicoxyzzy (approved):

afmenez:
The PR is already merged, I'll let someone with access to Safari to update the results.

PR #1376, opened 2018-11-28
Add Symbol.prototype.description to ES2019

chicoxyzzy:
Moved to stage 4 on November TC39 meeting


ljharb (approved):
I believe edge has it now too.

chicoxyzzy:
Yes, it should appear in Edge 19 Preview, but it seems that v19 wasn't released yet

PR #1378, opened 2018-11-28
Promote Function.prototype.toString revision to ES2019

chicoxyzzy:
As per November TC39 meeting


ljharb (approved):
We should add many more tests here :-)

PR #1391, opened 2018-12-25
[ES2019] Improve `Symbol#description` test case

zloirock:
Safari fails the third test.


ljharb (approved):
I believe edge fails the third case too.

zloirock:
@ljharb it's completely missed in Edge 18 and I haven't more modern versions.

PR #1397, opened 2018-12-26
[ESNext] Add `compositeKey` and `compositeSymbol`

zloirock:
(they should live on built-in modules, but they are still unavailable)

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -2880,10 +2880,10 @@ exports.tests = [
   ]
 },
 {
-  name: 'compositeKey and compositeSymbol',
+  name: 'richer keys',
   category: STAGE1,
   significance: 'small',
-  spec: 'https://github.com/bmeck/proposal-richer-keys/tree/master/compositeKey',
+  spec: 'https://github.com/bmeck/proposal-richer-keys/',
   subtests: [
     {
       name: 'compositeKey',
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -2879,6 +2879,43 @@ exports.tests = [
     },
   ]
 },
+{
+  name: 'compositeKey and compositeSymbol',
+  category: STAGE1,
+  significance: 'small',
+  spec: 'https://github.com/bmeck/proposal-richer-keys/tree/master/compositeKey',
+  subtests: [
+    {
+      name: 'compositeKey',
+      exec: function () {/*
+        var a = ['a'];
+        var b = ['b'];
+        var c = ['c'];
+        var ab = compositeKey(a, b);
+        return typeof ab == 'object'
+          && Object.isFrozen(ab)
+          && ab === compositeKey(a, b)
+          && ab !== compositeKey(a, c);
+      */},
+      res: {
+      },
+    },
+    {
+      name: 'compositeSymbol',
+      exec: function () {/*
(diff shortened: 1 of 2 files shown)


ljharb (changes requested):
I think this one is too premature to add - the repo itself hasn’t even been submitted by the champion to the proposals list, and until builtin modules progresses (if it does), it’d have to supply a global name to access it (which is what these tests would have to use).

zloirock:
@ljharb it's a part of richer keys stage 1 proposal. It's on the list. This proposal contains 2 separate parts, this PR adds the first. For preventing confusion, I can rename this test case by the name of the main proposal.

ljharb:
The proposal doesn’t yet provide enough information to implement it or test for it. This is not uncommon for proposals that agent stage 3.

zloirock:
@ljharb I do not completely understand you. For me, in the repo - readme and polyfill - enough information for basic tests. Missed parts were discussed with the champion when I wrote a polyfill.

ljharb:
The champion needs to submit the repo in a PR to the proposals list; and if it’s goijg to (foolishly, imo) chain itself to builtin modules, then it’s not possible for anything to pass any tests for it until builtin modules exist.

I completely agree that it’s possible to implement a polyfill for the individual functions, and tests for those, but i don’t agree that the proposal provided enough information to test the proposal overall (which is what goes in this table).

zloirock:
@ljharb 
> The champion needs to submit the repo in a PR to the proposals list;

This proposal already in the list and was presented tc39, seems just the link to the repo should be updated:

image

> foolishly, imo

I think the same, [see the issue](https://github.com/bmeck/proposal-richer-keys/issues/1).

> but i don’t agree that the proposal provided enough information to test the proposal overall 

Both parts of the proposal - keys composition and rekey - clearer than a serious part of the rest proposals in the ESNext table, so it's a strange argument

zloirock:
сс @bmeck

bmeck:
@ljharb I do plan on requesting stage 2 in January, perhaps it would be best to wait until after then? The proposal is split into 2 parts but is being pushed in tandem rather than disparate proposals due to overlap of context.

ljharb:
Given https://github.com/bmeck/proposal-richer-keys/issues/1#issuecomment-450198871, i don’t think the table should have a test for compositeKey at all at this time.

bmeck:
@ljharb I disagree with that statement, locations of references are free to change and do so during stage advancement.

ljharb:
That is true - but at the moment, there exists no location. Typically they change from one exiting location to a different one, and the test changes along with them.

The table doesn’t have tests for unobservable things, and at the moment, compositeKey is impossible to observe or access.

bmeck:
@ljharb that just seems to imply that the tests could be amended to allow for such access to be injected rather than preventing any testing. I don't see how that blocks things.

chicoxyzzy (approved):
LGTM

zloirock:
@ljharb you still wanna block it?

ljharb:
@zloirock based on today's TC39, rekey and compositeKey are being split out; i think compositeKey can't possibly have any tests unless the proposal specifies an existing way for it to be reachable: either via syntax, or via global API.

PR #1398, opened 2018-12-27
Separate parsing of test results from report generation

chrisdarroch:
This refactor makes it simpler to consume this repository as a set of tests and a test runner for the various JS environments. The `build.js` module remains responsible for constructing the compatibility table, where the `test-parser.js` module is responsible for prepping individual assertions for checking in a given environment.

I've done what I can to preserve the git history after splitting the files so that `git blame` continues to work in both files.


build.js
```diff
@@ -397,6 +399,25 @@ function handle(options) {
 }
 
 function dataToHtml(skeleton, rawBrowsers, tests, compiler) {
+  var wrapWithScript = function(assertions) {
+    // wrap a single test in an array so we can deal with them consistently
+    if (assertions && !Array.isArray(assertions)) {
```
ljharb: If you’re checking isArray, you could just do `[assertions]`; concat is how you’d do it without the if check at all.

build.js
```diff
@@ -397,6 +399,25 @@ function handle(options) {
 }
 
 function dataToHtml(skeleton, rawBrowsers, tests, compiler) {
+  var wrapWithScript = function(assertions) {
```
ljharb: This function can be hoisted out of dataToHtml

build.js
ljharb: Also, relying on Array.isArray won’t work on the pre-ES5 page.

build.js
```diff
@@ -397,6 +399,25 @@ function handle(options) {
...
+    // wrap a single test in an array so we can deal with them consistently
+    if (assertions && !Array.isArray(assertions)) {
+      assertions = [].concat(assertions);
+    }
+
+    return assertions.filter(function(assertion) {
```
ljharb: filter and map are not available pre-ES5.

test-parser.js
```diff
@@ -0,0 +1,114 @@
...
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+function prepareTest(fn, transformFn, rowNum) {
+  var script, expr;
+  function deindentFunc(fn) {
```
ljharb: this function can be hoisted outside of prepareTest

test-parser.js
```diff
@@ -0,0 +1,114 @@
...
+          if (expr.search(/Function\s*\(|eval\s*\(/) > -1) {
+            throw new Error("This test's code invokes eval() and cannot be compiled.");
+          }
+          expr = transformFn("(function(){" + expr + "})");
+          transformed = true;
+        } catch(e) {
+          expr = "/* Error during compilation: " + e.message + "*/";
+        }
+      }
+      /* jshint unused: true */ // this appears unused, but removing it might break things.
```
ljharb: if removing it allows the linter to pass, let’s remove it; if not, let’s remove the comment

test-parser.js
```diff
@@ -0,0 +1,114 @@
...
+        +  'catch(e){'
+        +    'return false;'
+        +  '}'
+        +'}());';
+      return { script: script, source: expr };
+    }
+  } else {
+    // it's an array of objects like the following:
+    // { type: 'application/javascript;version=1.8', script: function () { ... } }
+    return fn.map(function(script) {
```
ljharb: map is not available pre-ES5

ljharb (changes requested):
Thanks, i like the direction!

build.js
zloirock: @ljharb it's a building tool and it already uses `Array.isArray`.

ljharb: ah, ok great - then scratch all my pre-ES5 comments :-) just the first one in this thread then.

test-parser.js
chrisdarroch: It looks like this variable was never used; it seems like an artefact of [the initial commit](https://github.com/kangax/compat-table/commit/c4a601f1cc949e03ed1d080d2aab33f1682185a2) in which it was introduced. I can remove it.

build.js
chrisdarroch: I agree that it *could*, but there's another five functions that exist below it within the `dataToHtml` closure that don't need to be there, either. I was aiming for consistency. Should I move all of them in a separate refactor, perhaps?

ljharb: I suppose that's fine too.

ljharb (approved):
LGTM, thanks

test-parser.js
```diff
@@ -0,0 +1,111 @@
+// Copyright 2012 Dan Wolff | danwolff.se
```
chicoxyzzy, line 1: Why 2012 and who is copyright owner?

chrisdarroch: I split the original file in two, preserving history. Though github implies this is new content, it is in fact unchanged since the original file was created.

The copyright headers in both files could be revised, though I feel it is a separate cleanup task unrelated to this PR's scope.

chrisdarroch:
Is there anything further I can do to get this PR merged?

build.js
```diff
@@ -396,6 +398,26 @@ function handle(options) {
...
+  }
+
+  return assertions.filter(function(assertion) {
+    return assertion && assertion.script;
+  })
+  .map(function(assertion) {
+    var $ = cheerio.load(''
+        + '<script' + (assertion.type ? ' type="' + assertion.type + '"' : '') + '>'
+        + assertion.script
+        + '\n</script>'
```
zloirock, line 414: `\n` added here, but not after `<script>`. It's inconsistency. It should be in both places or nowhere. The rest LGTM.

zloirock (changes requested):

build.js
ljharb: It’s needed here for single-line comments; it’s not needed after the closing tag. So no matter what, it can’t be nowhere. I’m in favor of including it after as well, tho.

zloirock: I mean after opening tag. At this moment, it's nowhere and it's the reason of changes in generated HTML.

ljharb: I only see one before the closing tag - but sure, fair enough.

zloirock: Apparently, we just did not understand each other. It's added before the closing tag and it's the reason of changes in generated HTML, but not after opening tag. Adding it in both places could make sense.

PR #1399, opened 2018-12-28
Mark `babel@7` as stable

zloirock:


environments.json
```diff
@@ -21,22 +21,22 @@
   },
   "babel6": {
     "full": "Babel 6.5 + core-js 2.5",
-    "short": "Babel 6 +<br><nobr>core-js</nobr>",
+    "short": "Babel 6 +<br><nobr>core-js 2</nobr>",
     "family": "Babel",
     "platformtype": "compiler",
-    "obsolete": false,
+    "obsolete": true,
```
ljharb: I'd prefer not to mark babel 6 as obsolete just yet; it's still in active usage, and it's receiving the same level of maintenance and support as the last year or more before babel 7 was out.

zloirock: It makes sense.

ljharb (approved):

PR #1400, opened 2018-12-30
Add `core-js@3` results

zloirock:


data-esnext.js
```diff
@@ -305,8 +305,8 @@ exports.tests = [
       return typeof globalThis === 'object' && globalThis && globalThis === actualGlobal && !globalThis.lacksGlobalThis && globalThis.__system_global_test__ === 42;
     */},
     res: {
-      babel6: false,
```
ljharb: we should keep these; they represent verified results

zloirock: Ok.

ljharb (approved):

PR #1401, opened 2019-01-03
Rename old compilers shortcuts for consistency with #1400

zloirock:


ljharb (approved):
Do we not have typescript targets that don't include core-js?

zloirock:
@ljharb the builder generates tests for it (and for `babel`), but we haven't it in the table.

ljharb:
We should probably add them.

zloirock:
I'm not sure about it. We have already too many columns for compilers. Features which require `core-js` have notes about it.

ljharb:
Perhaps a redesign where things like babel, core-js, etc, can be toggled on or off on top of the basic runtime-less list?

Fair tho that as long as all the things that require the runtime have a note, it's fine - mainly thinking that we should find a way to have a single column for babel 6, and one for 7, and be able to pick a core-js version to apply on top of both (or none)

zloirock:
Makes sense.

kangax:
Relatedly, sometimes I open compat table to find certain feature but don't remember part of which "year" it is. It'd be nice to redesign it in such a way that it was searchable without toggling ES revisions.

PR #1402, opened 2019-01-09
Update test results for GraalVM 1.0.0 RC10

jirkamarsik:
Hi,

I have updated the test results for GraalVM to reflect the current version (1.0.0 RC10). This accounts for some new positives due to bug fixes and new features added since the last version recorded in the table (1.0.0 RC3).

It also fixes the non-standard table, in which GraalVM was mistakenly reported to support everything outside of SIMD.

As part of running the test suites on the new GraalVM, I have also taken a look at the test runners included in the repo. The `node.js` runner had an issue with using `setTimeout(_, 250)` to wait for tests to finish before producing the report. This led to some asynchronous tests (those which use longer delays and therefore don't have time to finish) to be reported as false negatives, both on the original node.js and on GraalVM. I fixed this by using `process.on('exit', _)`, which waits for the event queue to drain before exiting.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -599,7 +599,7 @@ exports.tests = [
           opera10_50: false,
           safari10_1: true,
           duktape2_0: false,
-          graalvm: false,
+          graalvm: true,
         }
       },
       {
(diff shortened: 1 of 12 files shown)


zloirock (approved):

chicoxyzzy (approved):
Thank you!

PR #1404, opened 2019-01-10
Add "JSON superset" tests to the ES2016+ tab

afmenez:
This is the only Stage 4 proposal that has no tests yet.  I ported both subtests from test262, so I believe they are correct.
I haven't run the tests on Safari, got the results from the release notes.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -3069,6 +3069,46 @@ exports.tests = [
       },
     }]
   },
+  {
+    name: 'JSON superset',
+    spec: 'https://github.com/tc39/proposal-json-superset',
+    category: '2019 misc',
+    significance: 'small',
+    subtests: [
+      {
+        name: 'LINE SEPARATOR can appear in string literals',
+        exec: function(){/*
+          return eval("'\u2028'") === "\u2028";
+        */},
+        res : {
+          ie11: false,
+          firefox2: false,
+          firefox61: false,
+          firefox62: true,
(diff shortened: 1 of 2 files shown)


afmenez:
I believe it works on Babel 7 too, but now I don't know how to add the results with the core2/core3 complication.

ljharb (approved):
Seems useful to maybe also add a test that utilizes JSON functions? LGTM otherwise

afmenez:
Yes, it seemed strange to me that the test262 tests didn't mention any JSON. :)
The [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON#JavaScript_and_JSON_differences) says that `JSON.parse()` works on all browsers, but just newer browsers support `eval()`, so I assumed that was the important test.  But I don't know how to add JSON functions if they won't fail for old browsers.

afmenez:
Please note [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1435828#c15) on the Firefox bug:
"extend ECMA-262 syntax into a superset of JSON" is a somewhat obscure way to describe the *effect* of this change.  I would instead suggest:

"JavaScript string literals may now directly contain `U+2028 LINE SEPARATOR` and `U+2029 PARAGRAPH SEPARATOR`.  As a consequence, JSON syntax is now a subset of JavaScript literal syntax."

chicoxyzzy (approved):

PR #1406, opened 2019-01-20
Update Closure Compiler to v20190106

teppeis:
No updates for compat-table in recent two versions.
It has passed Symbol.prototype.description undefined test for a long time.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -2917,6 +2917,7 @@ exports.tests = [
         res : {
           babel6corejs2: false,
           babel7corejs3: babel.corejs,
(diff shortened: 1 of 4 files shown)


data-es2016plus.js
```diff
@@ -2917,6 +2917,7 @@ exports.tests = [
         res : {
           babel6corejs2: false,
           babel7corejs3: babel.corejs,
+          closure: true,
```
zloirock, line 2920: Seems it's my mistake in the code of this test. I'll fix it.

zloirock (approved):

PR #1407, opened 2019-01-21
Update Node.js -> Chrome mapping

zloirock:
per https://nodejs.org/dist/index.tab


ljharb (approved):

chicoxyzzy:
We can use https://github.com/chicoxyzzy/node-releases to autogenerate this

PR #1415, opened 2019-02-01
Chrome 74 Canary has Static Private Fields under flag

chicoxyzzy:


ljharb (approved):
I would expect the entirety of the class fields proposal to be shipped under a flag; is this not the case?

chicoxyzzy:
> I would expect the entirety of the class fields proposal to be shipped under a flag; is this not the case?

I can't test in Chrome beta 73, but Chrome 72 doesn't have this feature even with `--harmony` flag

ljharb:
I think maybe we should be testing these things before updating the results and not just relying on release announcements. Browsers claiming they have a feature has never been a reliable indicator that they built it right.

chicoxyzzy:
I thought that we are already doing this... At least I do :) All results for class fields except Canary 74 look valid. I think that private static fields was added in Canary in one of latest patches after first 74 release. Although I could miss testing with `--harmony` flag for PR with Chrome 72 / Chrome 74 Canary releases, not sure TBH.
image

ljharb:
ah i just misunderstood your comment :-)

zloirock (approved):

PR #1417, opened 2019-02-03
[ESNext] Add `Promise.any`

zloirock:


ljharb (approved):
This is fine, but I’d like to see more tests that verify execution order.

zloirock:
@ljharb I think makes sense to add them after writing the spec text / advances to the next stages.

Issue #1419, opened 2019-02-11
Babel Can't find private variable @derivedConstructor (Safari iOS 10.3.1)


adrienharnay:
Hey,

Related Babel issue: https://github.com/babel/babel/issues/9465

Basically, async private methods get transpiled wrongly by Babel, resulting in something that looks like this:

```js
class SomeClass_SomeClass extends react_default.a.Component {
  constructor() {
    super(...arguments);

    _defineProperty(this, "state", getInitialState(this.props));

    _defineProperty(this, "someFunction", async (props, state) => {
      ...
    }
  }
}
```

Safari 10.3.1 doesn't seem to like it, throwing this error:
`Can't find private variable @derivedConstructor` on the `async` keyword.

I'm not sure how compat-table is used by Babel, but please tell me if I can help solving this!

Thanks in advance

ljharb:
Seems like we need a test on async functions in classes specifically, that covers that browser - won’t to make a PR?

adrienharnay:
Sure! Could I ask for a starting point to understand how I should proceed? Thanks!

ljharb:
Async functions are tested here: https://github.com/kangax/compat-table/blob/c28c86b928408920ed3650c40a88a72f9742d243/data-es2016plus.js#L448-L897

You'd want to add another object to that array, that tested this case, and included test results.

adrienharnay:
I see some code that looks like what I'm doing: https://github.com/kangax/compat-table/blob/c28c86b928408920ed3650c40a88a72f9742d243/data-es2016plus.js#L782-L813, isn't it already covered then?

ljharb:
Ah, if so, then you may just need to make a PR that adds the `false` test results for the browser that has a bug?

adrienharnay:
I created a PR, but as I'm said I'm not really sure what I'm doing so I hope I'm right! https://github.com/kangax/compat-table/pull/1420 Thanks for your guidance

PR #1420, opened 2019-02-11
Fix Can't find private variable @derivedConstructor (Safari iOS 10.1 to 10.3.1)

adrienharnay:
Async arrow functions as class private methods don't seem to work in Safari 10.1 to 10.3.1.

I'm very new to compat-table so please review this carefuly.

Context:
- https://github.com/kangax/compat-table/issues/1419
- https://github.com/babel/babel/issues/9465

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -2356,7 +2356,11 @@ exports.tests = [
         res: {
           babel6corejs2: true,
           closure: true,
-          closure20181028: false,
+          closure20181028: {
+            val: false,
+            note_id: 'closure-object-assign',
+            note_html: 'Requires native support for <code>Object.assign</code>',
+          },
           typescript2_1corejs2: true,
           jsx: true,
           ie11: false,
@@ -2383,6 +2387,10 @@ exports.tests = [
         res: {
           babel6corejs2: true,
           closure: true,
+          closure20190121: {
+            val: false,
+            note_id: 'closure-object-assign',
+          },
           jsx: true,
           ie11: false,
           firefox2: false,
@@ -2778,6 +2786,7 @@ exports.tests = [
         */},
         res: {
           babel7corejs2: true,
+          closure20190215: true,
           typescript2_5corejs2: true,
           ie11: false,
           firefox2: false,
@@ -2807,6 +2816,7 @@ exports.tests = [
         */},
         res: {
           babel7corejs2: true,
+          closure20190215: true,
           typescript2_5corejs2: true,
           ie11: false,
           firefox2: false,
@@ -2841,6 +2851,7 @@ exports.tests = [
         */},
         res: {
           babel7corejs2: true,
+          closure20190215: true,
           typescript2_5corejs2: true,
           ie11: false,
           firefox2: false,
@@ -3083,6 +3094,7 @@ exports.tests = [
           return eval("'\u2028'") === "\u2028";
         */},
         res : {
+          closure20190215: true,
           ie11: false,
           firefox2: false,
           firefox61: false,
@@ -3099,6 +3111,7 @@ exports.tests = [
           return eval("'\u2029'") === "\u2029";
         */},
         res : {
+          closure20190215: true,
           ie11: false,
           firefox2: false,
           firefox61: false,
--- a/data-es6.js
+++ b/data-es6.js
@@ -18659,6 +18659,7 @@ exports.tests = [
       res: {
         ejs: true,
         babel6corejs2: babel.corejs,
+        closure20190215: true,
         typescript1corejs2: typescript.corejs,
         tr: true,
         es6shim: true,
--- a/environments.json
+++ b/environments.json
@@ -73,7 +73,7 @@
     "short": "Closure 2018.03",
     "family": "Closure",
     "platformtype": "compiler",
-    "obsolete": true,
+    "obsolete": "very",
     "test_suites": [
       "es6",
       "es2016plus",
@@ -85,7 +85,7 @@
     "short": "Closure 2018.04",
     "family": "Closure",
     "platformtype": "compiler",
-    "obsolete": true,
+    "obsolete": "very",
     "test_suites": [
       "es6",
       "es2016plus",
@@ -97,7 +97,7 @@
     "short": "Closure 2018.05",
     "family": "Closure",
     "platformtype": "compiler",
-    "obsolete": true,
+    "obsolete": "very",
     "test_suites": [
       "es6",
       "es2016plus",
@@ -109,7 +109,7 @@
     "short": "Closure 2018.06",
     "family": "Closure",
     "platformtype": "compiler",
-    "obsolete": true,
+    "obsolete": "very",
     "test_suites": [
       "es6",
       "es2016plus",
@@ -121,7 +121,7 @@
     "short": "Closure 2018.07",
     "family": "Closure",
     "platformtype": "compiler",
-    "obsolete": true,
+    "obsolete": "very",
     "test_suites": [
       "es6",
       "es2016plus",
@@ -133,7 +133,7 @@
     "short": "Closure 2018.08",
     "family": "Closure",
     "platformtype": "compiler",
-    "obsolete": true,
+    "obsolete": "very",
     "test_suites": [
       "es6",
       "es2016plus",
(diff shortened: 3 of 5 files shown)


data-es2016plus.js
```diff
@@ -806,7 +806,7 @@ exports.tests = [
           firefox2: false,
           firefox52: true,
           opera10_50: false,
-          safari10_1: true,
+          safari10_1: false,
```
ljharb: Do they work in Safari 11? if so, let's add `safari11: true`. Also, you said to 10.3.1 - is there a 10.3.2 or 10.4 that it does work in? If so, let's add it to environments.json and set *it* to `true` as well.

ljharb (changes requested):

data-es2016plus.js
adrienharnay: I will test Safari 10, 10.1, 10.2 and so on tomorrow and update the table accordingly 👍

adrienharnay: After a few tests, both on Safari Mobile and Safari Webview, here are the results:

```
NOT WORKING:
Safari Webview iOS 10.1
Safari Webview iOS 10.2
Safari Webview iOS 10.3.1
Safari Mobile iOS 10.3.1
Safari Webview iOS 11

WORKING:
Safari Mobile iOS 9.3
Safari Mobile iOS 10.1
Safari Mobile iOS 10.2
Safari Mobile iOS 11
```
(couldn't test Safari Webview iOS 9.3 yet)
(Thanks to @titozzz who helped me a lot with the tests ❤️)

I'm not sure what to do with this? Since for the same version of iOS, it sometimes works on Mobile but not on Webview.

adrienharnay: Hey, any update on this?

ljharb: basically, any result that can’t be represented right now will necessitate the addition of browsers to environments.json.

adrienharnay: Alright, we will do this with @Titozzz tomorrow

adrienharnay: @Titozzz just tested everything again, and I was mistaken.

Only Safari 10.3 (both webview and mobile) has the bug. Will update the PR

existentialism:
@adrienharnay @Titozzz nice work! I'll get env updated when this lands!

data-es2016plus.js
adrienharnay: Update:

Only **iOS 10.3** has the bug. This is Safari 10.1.

iOS 10, 10.1 and 10.2 have Safari 10.0.

I have update the PR, I think this is good this time, please let me know if not! Thanks

data-es2016plus.js
```diff
@@ -806,7 +806,7 @@ exports.tests = [
           firefox2: false,
           firefox52: true,
           opera10_50: false,
-          safari10_1: true,
+          safari11: true,
```
ljharb: i think we need a safari10_3 to mark false here, then?

Titozzz:
Hello, just to add some clarifications, because the version numbers are very annoying and confusing 😠!

So we tested this on:

iOS 9.X - Safari Mobile 9.? - Webkit version 601.1.46
iOS 10.0 - Safari Mobile 10.0 - Webkit version 602.1.50
iOS 10.1 - Safari Mobile 10.0.1 - Webkit version 602.2.14
iOS 10.2 - Safari Mobile 10.0.2 - Webkit version 602.3.12
**iOS 10.3.X - Safari Mobile 10.1 - Webkit version 603.1.30** - *broken*
iOS 11.0 - Safari Mobile 11.? - Webkit version 604.2.10
iOS 11.1 - Safari Mobile 11.? - Webkit version 604.3.5

After testing all these on both webview and safari directly, only one version is broken.

What we tried to achieve in this PR, is to say that async arrow function only work in 11 and above, so we moved it from 10.1 to 11. I think it works below 10.1 because the compat only says it's enabled after 10.1.

Is this the right way to proceed @ljharb?

ljharb:
@Titozzz I understand that your goal is to inform babel-preset-env, but that's not the primary purpose of the compat table. If the feature works in 10.2 and 11 but is broken in 10.3, the table needs to convey that.

adrienharnay:
Wasn't our intention at all, we just didn't know if we were supposed to specify the oldest working version or what you said! I'm updating the PR

ljharb:
@adrienharnay since it works in 10.1, i think we need to add an entirely new browser entry for 10.3, so that can be marked false.

adrienharnay:
Safari 10.3 doesn’t exist, please look at the list @Titozzz posted earlier. This is very confusing, but basically:
- iOS 10, 10.1, 10.2 = Safari 10
- iOS 10.3 = Safari 10.1 (broken)

ljharb:
Right - I’m saying we need to add an entry for it.

ljharb:
Or are you saying that desktop 10.1 is already associated with mobile 10.3, and it’s broken on both?

adrienharnay:
I am saying that Safari 10.3 dosn’t exist _at all_. Safari 10.1 exists, has the bug, and is installed on iOS 10.3. So the table is good as it is now. Sorry if this isn’t very clear, took us some time to figure out

ljharb:
ahhhh got it, thanks for clearing up my confusion :-)

ljharb (approved):

ljharb:
You'll still need to run `npm run build` before CI will pass :-)

zloirock (changes requested):
I'm not sure how this test related private class methods.

Seems you should add a new test for private fields and add it to `preset-env` dependencies.

Or it's not related private fields?

ljharb:
@zloirock it's not supposed to; the error message safari gives has the word "private" in it, but it's not related to class fields.

zloirock:
@ljharb anyway, this test passed in iOS 10.3:
image

Titozzz:
Are we sure safari support async arrow functions in 10 ? 
https://caniuse.com/#feat=async-functions
I would revert 13b74c9 but I'm not sure anymore 🤔, I'll talk to @adrienharnay tommorrow about that.

data-es2016plus.js
```diff
@@ -806,6 +806,40 @@ exports.tests = [
...
+          safari10_1: true,
+          duktape2_0: false,
+          graalvm: true,
+        }
+      },
+      {
+        name: 'async arrow functions, classes',
+        exec: function () {/*
+          class C {
+            a = async () => { return await Promise.resolve("foo"); }
```
ljharb: this syntax is class fields, which is stage 3, and as far as i know, only shipped in chrome.

A test for async arrow functions shouldn't need class fields; if it's a class fields bug, the test should go there instead.

adrienharnay: Trying to figure this out, will come back to you soon :+1:

Titozzz: @ljharb just so you know, this is the bug we are trying to get rid of: https://bugs.webkit.org/show_bug.cgi?id=166879

Titozzz:
So we kept digging, the issue does not appear on iOS devices below 10.3 because we are using esmodules (https://babeljs.io/docs/en/babel-preset-env#targetsesmodules).

So finally the issue is that **iOS 10.3 (safari mobile 10.1)** support esmodules but crashes when using async arrow class methods.

@existentialism Would you know what needs to be updated to fix this?

data-es2016plus.js
zloirock: @Titozzz so just use something like the first example from the tracker - it does not use class fields.

Titozzz: You mean inside the compat test @zloirock?

zloirock: @Titozzz I mean that you can use something like [this](https://bugs.webkit.org/show_bug.cgi?id=166879#c0) here, in async functions tests, without usage class fields which can't be used here.

adrienharnay: I have updated the PR (but not 100% sure about my test), could you please tell me what you think?

zloirock: If it can detect this bug - LGTM.

adrienharnay: We just copy-pasted the code in each browser, and it does detect the bug when needed :+1:

ljharb:
Can you explain what ESM has to do with this? Classes work in Scripts too.

ljharb:
(and to correct terminology; there’s no such thing as an arrow method; arrow functions can only be properties of objects)

adrienharnay:
@ljharb I just ran `npm run build`, and the PR should be good :+1:

adrienharnay:
> Can you explain what ESM has to do with this? Classes work in Scripts too.

It means that Safari 10.3 loads the `<script type="module">` (.mjs), with the syntax that causes the problem here.

Safari < 10.3 loads the `<script nomodule>` (.js), which is transpiled with @babel/preset-env / browserlist and therefore doesn't have the problematic syntax.

More info here: https://caniuse.com/#feat=es6-module

ljharb:
ah. You could still make a test page tho with always loading the script with the untranspiled syntax :-)

zloirock (approved):

zloirock:
@adrienharnay please, rebase it.

ljharb (approved):

adrienharnay:
Well, I messed up 😅Re-opened a PR: https://github.com/kangax/compat-table/pull/1423

PR #1424, opened 2019-02-26
Add test for TemplateStrings caching

jridgewell:
Safari 12 has a [bug](https://bugs.webkit.org/show_bug.cgi?id=190756) that breaks caching of `TemplateStrings`.

Test case: https://jsbin.com/leyusas/10/edit?js,console

Unfortunately, it's a memory GC bug, so it's very difficult to write a test that would catch this.

Re: https://github.com/babel/babel/pull/9584

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -5461,6 +5461,11 @@ exports.tests = [
     {
       name: 'TemplateStrings call site caching',
       exec: function () {/*
+        // Safari 12 sometimes garbage collects the cached TemplateStrings,
+        // even when the call site is still alive.
+        // This is almost impossible to write a test case for, but we can test
+        // the general semantics.
+        // https://bugs.webkit.org/show_bug.cgi?id=190756
         function strings(array) {
           return array;
         }
--- a/data-es6.js
+++ b/data-es6.js
@@ -5458,6 +5458,43 @@ exports.tests = [
         graalvm: true,
       },
     },
+    {
+      name: 'TemplateStrings call site caching',
+      exec: function () {/*
+        function strings(array) {
+          return array;
+        }
+        function getStrings() {
+          return strings`foo`;
+        }
+        var original = getStrings();
+        var other = strings`foo`;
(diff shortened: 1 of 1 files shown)


ljharb (approved):
LGTM but maybe we could add a note to the test linking to the bug, and explaining why the test might be tricky to verify?

jridgewell:
Done.

zloirock (changes requested):
Please, rebuild it one more time.

jridgewell:
Done.

zloirock (approved):

moderndeveloperllc:
@jridgewell Just come upon this when updating to the babel 7.4 (which includes the `compat-table` build that fixes this). One of the things this fix means is that any build that targets `'last 2 Safari major versions',` via browserlist syntax is now including the `babelHelpers.taggedTemplateLiteral()` where previously it was not. This can cause build pipeline issues with projects using a module/nomodule split as this is not normally needed for ES6 module-compatible browsers.

What did this bug cause to happen? Before I go and refudge my config, I was hoping to figure out if I really need to serve Safari 12 the helper-wrapped version or not. Thanks.

jridgewell:
The bug causes:

```js
function strings(array) {
  return array;
}
function getStrings() {
  return strings`foo`;
}

const ref = getStrings();

// This should always be true
// It may not be in Safari
ref === getStrings();
```

Many libraries use the "always the same reference" property to do caching (via `WeakMap`s) of expensive work. If the references are not the same, then the caching is broken and the expensive work has to be redone. In some libraries, it can cause serious bugs.

moderndeveloperllc:
@jridgewell Thanks for the explanation. It's unfortunate that babel preset-env [currently can't handle](https://github.com/babel/babel/issues/9739) the case of a feature having upper bounds in compat-table. BTW, you may want to add this issue to [core-js](https://github.com/zloirock/core-js) as that is what babel is transitioning to it appears. There does not appear to be a template literals test in the list there.

PR #1427, opened 2019-02-28
Update flat/flatMap @@unscopables for Chrome and Firefox

chicoxyzzy:
\+ template strings for Safari


ljharb (approved):

data-es6.js
```diff
@@ -5491,6 +5491,7 @@ exports.tests = [
         chrome41: true,
         safari9: true,
         safari12: false,
+        safari12_1: true,
```
jridgewell: Unless 12.1 is newer than TP 76, this is incorrect.
screen shot 2019-02-28 at 3 15 18 pm

jridgewell (changes requested):

data-es6.js
chicoxyzzy: It's Version 12.1 (14607.1.35.1)
Also my Safari TP is Release 76 (Safari 12.2, WebKit 14608.1.6.2)

chicoxyzzy: image

jridgewell: It's a memory bug that's impossible to write a test for. The general semantics are all that is tested.

Actual bug: https://bugs.webkit.org/show_bug.cgi?id=190756
Actual reproduction: https://jsbin.com/peniburaho/1/edit?html,console (you'll have to wait, possibly use other tabs)

chicoxyzzy: Can't reproduce on SF 12.1
image

chicoxyzzy: I left a comment in WebKit Bugzilla

chicoxyzzy: Hmm I was able to reproduce it once
image
After that it works again

ljharb: That suggests it's not fixed.

Either way maybe let's just remove it from this PR, which per the title it's out of scope for, and we can discuss it in a separate PR?

jridgewell: It's definitely not fixed, I can reproduce it most times in TP. If we could reproduce with 100%, I would have written the test better. 😉

chicoxyzzy: I was able to catch it second time
image

PR #1433, opened 2019-03-13
Instance class fields are supported under flag in Firefox 67

chicoxyzzy:
https://bugzilla.mozilla.org/show_bug.cgi?id=1529758

image


chicoxyzzy:
It seems that in fact private fields are exposed as public

ljharb:
That should get it a “no” for support, not a flag. Privacy is the primary thing that matters for private fields.

ljharb (approved):

PR #1435, opened 2019-03-21
Mark `core-js@3` as stable

zloirock:
https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md

### DIFF
--- a/build.js
+++ b/build.js
@@ -132,7 +132,7 @@ process.nextTick(function () {
       name: 'babel + core-js',
       url: 'https://babeljs.io/',
       target_file: 'es6/compilers/babel-core-js.html',
-      polyfills: ['node_modules/babel-polyfill/browser.js'],
+      polyfills: ['node_modules/core-js-bundle/index.js', 'node_modules/regenerator-runtime/runtime.js'],
       compiler: function(code) {
         return babel.transform(code, {presets: ['es2015']}).code;
       },
@@ -257,7 +257,7 @@ process.nextTick(function () {
       name: 'TypeScript + core-js',
       url: 'https://www.typescriptlang.org/',
       target_file: 'es6/compilers/typescript-core-js.html',
-      polyfills: ["node_modules/core-js/client/core.js"],
+      polyfills: ["node_modules/core-js-bundle/index.js"],
       compiler: function (code) {
         return ts.transpile(code, { downlevelIteration: true });
       }
@@ -293,7 +293,7 @@ process.nextTick(function () {
       name: 'babel + core-js',
       url: 'https://babeljs.io/',
       target_file: 'esnext/compilers/babel-core-js.html',
-      polyfills: ['node_modules/babel-polyfill/browser.js'],
+      polyfills: ['node_modules/core-js-bundle/index.js', 'node_modules/regenerator-runtime/runtime.js'],
       compiler: function(code) {
         return babel.transform(code, {presets: ['es2015', 'es2016', 'es2017', 'babel-preset-stage-0']}).code;
       },
@@ -319,7 +319,7 @@ process.nextTick(function () {
       name: 'TypeScript + core-js',
       url: 'https://www.typescriptlang.org/',
       target_file: 'esnext/compilers/typescript-core-js.html',
-      polyfills: ["node_modules/core-js/client/core.js"],
+      polyfills: ["node_modules/core-js-bundle/index.js"],
       compiler: ts.transpile
     },
   ].forEach(function(e){
@@ -333,7 +333,7 @@ process.nextTick(function () {
       name: 'babel + core-js',
       url: 'https://babeljs.io/',
       target_file: 'es2016plus/compilers/babel-core-js.html',
-      polyfills: ['node_modules/babel-polyfill/browser.js'],
+      polyfills: ['node_modules/core-js-bundle/index.js', 'node_modules/regenerator-runtime/runtime.js'],
       compiler: function(code) {
         return babel.transform(code, {presets: ['es2015', 'es2016', 'es2017']}).code;
       },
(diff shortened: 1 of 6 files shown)


build.js
```diff
@@ -257,11 +257,12 @@ process.nextTick(function () {
       name: 'TypeScript + core-js',
       url: 'https://www.typescriptlang.org/',
       target_file: 'es6/compilers/typescript-core-js.html',
-      polyfills: ["node_modules/core-js/client/core.js"],
+      polyfills: ["node_modules/core-js-bundle/index.js"],
       compiler: function (code) {
         return ts.transpile(code, { downlevelIteration: true });
       }
     },
+    /*
```
ljharb, line 265: This should be removed?

build.js
```diff
@@ -27,9 +27,9 @@ var environments = require('./environments');
 
 var fs = require('fs');
 var path = require('path');
-var os = require('os');
+// var os = require('os');
```
ljharb, line 30: Why are these commented?

package.json
```diff
@@ -13,14 +13,13 @@
...
     "babel-core": "latest",
-    "babel-polyfill": "latest",
     "babel-preset-es2015": "latest",
     "babel-preset-es2016": "latest",
     "babel-preset-es2017": "latest",
     "babel-preset-stage-0": "latest",
     "chalk": "^1.1.3",
     "cheerio": "^0.20.0",
-    "core-js": "latest",
+    "core-js-bundle": "^3.0.0",
```
ljharb, line 22: if the package name is changing, shouldn’t that be communicated somehow?

package.json
```diff
@@ -33,6 +32,7 @@
     "jstransform": "latest",
     "lodash.pickby": "^4.6.0",
     "object.assign": "^4.0.4",
+    "regenerator-runtime": "^0.13.2",
```
ljharb, line 35: Why is this needed at the top level?

build.js
zloirock: Maybe, it can be fixed in the future.

build.js
zloirock: Because it's required for Closure Compiler tests which now can't be built.

build.js
ljharb: That seems like something that needs to be fixed separately. cc @teppeis ?

package.json
zloirock: It's the package only with bundled version. Communicated - what do you mean? It's renamed in cases of usage.

package.json
zloirock: For replacement of deprecated and obsolete `babel-polyfill`.

package.json
ljharb: why do we need the bundled version here?

package.json
ljharb: what about `@babel/polyfill`?

ljharb: either way everything is labeled as “with core-is”, not “with regenerator-runtime”. Should we remov this and update the test results?

package.json
zloirock: For generation of pages with test cases? See the source code of `build.js`.

package.json
zloirock: `@babel/polyfill` is dead. `regenerator` transform is a part of `babel`, I don't think that any additional labels required.

build.js
zloirock: Seems since @teppeis provide data not only for ES6, he has an alternative tests runner.

teppeis: @ljharb @zloirock I'm using my own build scripts for closure compiler: https://github.com/teppeis/closure-compiler-es-compat-table
Because it's too slow to run tests one by one in Java, I use [nailgun](https://github.com/facebook/nailgun) to reduce the overhead of starting JVM and add some convenient features.

So it is not a problem for me to delete this closure part of build.js, but I do not know if it should be deleted for compat-table project.

ljharb:
I don’t see any approvals on this yet.

zloirock:
Hmm... You did not request any changes, I answered your questions, there was enough time. If you are against something - feel free to open a new PR or issue and we will discuss it.

ljharb:
The lack of a block isn’t approval; I’ll consistently block in the future.

zloirock:
Can you mark any points in this pull request you don't agree with?

ljharb:
Everything i commented. For one, if the label is going to say “core-js” then regenerator-runtime and it’s results should not be included (and shouldn’t have been previously). If it’s going to include regenerator-runtime, then the label should include that. I also don’t think we should have committed the code commenting out the broken closure compiler bit - i think that should stay broken in master til it’s fixed (and should have CI that exercises it so that can’t happen again). I also think that if we’re using core-js-bundle then that might be misleading to users who would likely use core-js; i think we should use core-js itself (and do our own bundling if needed).

zloirock:
> Everything i commented.

And I answered. You had 17 hours to say that you disagree.

> For one, if the label is going to say “core-js” then regenerator-runtime and it’s results should not be included (and shouldn’t have been previously).

Really?  As I wrote, `babel` contains `regenerator` transform, `regenerator-runtime` is a runtime for that. We haven't any labels about runtimes for other engines.

>  I also don’t think we should have committed the code commenting out the broken closure compiler bit - i think that should stay broken in master til it’s fixed (and should have CI that exercises it so that can’t happen again).

This code is broken on `master` for a long time. And it prevents adding the new result for compilers for a long time. Maybe because of this we have outdated results for almost all of them.

> I also think that if we’re using core-js-bundle then that might be misleading to users who would likely use core-js; i think we should use core-js itself (and do our own bundling if needed).

Previously, `compat-table` used bundled `core-js` version from `core-js` package. Now it's just moved to a separate package, so it's a strange issue. Feel free to rewrite tests builder if you think that it's a good idea, I haven't any time for that.

PR #1438, opened 2019-03-23
BigInt is enabled by default in Firefox Nightly

chicoxyzzy:
https://bugzilla.mozilla.org/show_bug.cgi?id=1531293

I don't like how results look with this PR though.

image

image

### DIFF
--- a/data-common.json
+++ b/data-common.json
@@ -52,17 +52,17 @@
       "note_html": "The feature have to be enabled via \"javascript.options.shared_memory\" setting under <code>about:config</code>.  It is enabled by default in Firefox Developer Edition and Firefox Nightly builds."
     },
     "nomob": {
-      "val": true,
+      "val": "flagged",
       "note_id": "firefox-nomob",
       "note_html": "The feature is available on desktop versions only, it is not available on mobile versions yet."
     },
     "developer": {
-      "val": false,
+      "val": "flagged",
       "note_id": "firefox-developer",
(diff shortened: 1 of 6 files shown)


zloirock:
Maybe makes sense change the value of `firefox.nightly` to `flagged`?

afmenez:
`firefox.nightly` normally means that the feature is only available on nightly builds.  `BigInt` is also available on beta with the flag enabled, so it should be `flagged` on 57, and `true` on 58.

zloirock:
@afmenez it will cause a problem with marking previously nightly versions as stable.

afmenez:
I don't agree with the latest patch.  This is not actually a "nightly" feature.  All the features previously marked as "nightly" were only available on nightly builds (and most of them always on), and they were removed when the version was promoted to beta.  BigInt on the other hand is flagged on beta and on by default on nightly, so why not just set firefox68 to true?

zloirock (approved):

chicoxyzzy:
@afmenez because as @zloirock said, the feature is not in Firefox 68 (which will be stable and beta and not necessarily will have feature support in such versions), but just Firefox 68 Nightly

chicoxyzzy:
@ljharb PTAL at [last commit](https://github.com/kangax/compat-table/pull/1438/commits/505fc1a8870b60f3270b7c356640ef2f1f2ff512) when you'll have time

ljharb (approved):

Issue #1437, opened 2019-03-21
Add more class fields tests


chicoxyzzy:
Currently tests don't cover computed fields
https://bugzilla.mozilla.org/show_bug.cgi?id=1535166

afmenez:
Wasn't this fixed by #1440?

chicoxyzzy:
Not all cases were covered by #1440 
For example
```js
class C {
    2 = 2;
    0b01 = 2;
    "hello" = 2;
}
```
and presence of property x in instance of
```js
class C {
    x;
}
```
are still worth testing IMO

chicoxyzzy:
@evseevdev would you be open to add more tests? If so, you may ping me in Telegram again if you need more information.

PR #1440, opened 2019-03-25
[ESNext] Add computed class fields tests

evseevdev:
This PR covers instance and static class fields. Issue: https://github.com/kangax/compat-table/issues/1437


chicoxyzzy (approved):
Awesome, thanks!

zloirock (approved):

PR #1449, opened 2019-04-17
Fix BigInt results for Firefox 68

### LINKED ISSUE #1448: BigInt support should be "true" for Firefox 68
`BigInt` support is marked "Nightly only" on #1438, but Firefox 68 actually has it enabled by default.  The "nightly only" part was previously removed from Firefox 67 with bug [1522436](https://bugzilla.mozilla.org/show_bug.cgi?id=1522436).

I'll post a PR fixing this later.

afmenez:
Fix BigInt results for Firefox 68 (fixes #1448).
Also revert changes to the Firefox.nightly, Firefox.developer and Firefox.nomob flags.
Finally, fix the formatting of some of the notes for Firefox.

### DIFF
--- a/data-common.json
+++ b/data-common.json
@@ -52,17 +52,17 @@
       "note_html": "The feature have to be enabled via <code>javascript.options.shared_memory</code> setting under <code>about:config</code>.  It is enabled by default in Firefox Developer Edition and Firefox Nightly builds."
     },
     "nomob": {
-      "val": true,
+      "val": "flagged",
       "note_id": "firefox-nomob",
       "note_html": "The feature is available on desktop versions only, it is not available on mobile versions yet."
     },
     "developer": {
-      "val": false,
+      "val": "flagged",
       "note_id": "firefox-developer",
       "note_html": "The feature is available only in Firefox Developer Edition and Firefox Nightly builds."
     },
     "nightly": {
-      "val": false,
+      "val": "flagged",
       "note_id": "firefox-nightly",
       "note_html": "The feature is available only in Firefox Nightly builds."
     },
--- a/data-common.json
+++ b/data-common.json
@@ -49,37 +49,37 @@
     "sharedmem": {
       "val": "flagged",
       "note_id": "firefox-sharedmem",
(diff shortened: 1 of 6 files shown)


ljharb (approved):

zloirock (approved):

data-common.json
```diff
@@ -49,37 +49,37 @@
...
     },
     "developer": {
-      "val": "flagged",
+      "val": false,
       "note_id": "firefox-developer",
       "note_html": "The feature is available only in Firefox Developer Edition and Firefox Nightly builds."
     },
     "nightly": {
-      "val": "flagged",
+      "val": false,
```
zloirock: I'm not sure that it's a good idea.

zloirock (changes requested):

data-common.json
afmenez: It depends on what "flagged" means.  To me it means that the feature is not available by default, but it can be turned on by the user (either by changing the browser configuration or adding a command line parameter).  This is not what nightly means: it is available by default on nightly builds (via compilation flags), and not available at all on other builds.  Changing this to "flagged" marked lots of features as optionally available on old builds, which is not true.

chicoxyzzy (changes requested):
I think it'll be better to move changes from `data-common.json` to another issue / PR

afmenez:
With this PR, I am trying to fix the changes from #1438, including the changes to `data-common.json`.

chicoxyzzy:
Why those changes need a fix? Actually they fix another issue which I described in first comment of #1448 and we can face it again in future.

chicoxyzzy:
Another solution may be renaming `"flagged"` to something else, but I'm still think that it's a separate issue.

afmenez:
They need a fix because they incurred in wrong results for every feature marked "nightly" previously.  The change was done in #1438 assuming the implementation was still "nightly only", but the conditional compilation was already removed by then.

chicoxyzzy:
I don't think that results are wrong. You should interpret them as `available under flag` or `available in nightly`. Notes provide all necessary additional information.

afmenez:
I reverted the flag changes, will resubmit them in a separate PR.

zloirock (approved):

chicoxyzzy (approved):
Thanks! LGTM

PR #1450, opened 2019-04-22
Automaticly Add Results To File (for Duktape)

TheBrokenRail:


duktape.js
```diff
@@ -139,3 +144,44 @@ fs.readdirSync('.').forEach(function (filename) {
...
+console.log("Writing New Results To File...");
+for (var i = 0; i < fixTests.length; i++) {
+    var path = fixTests[i].path;
+    var success = fixTests[i].success;
+    var data = fs.readFileSync(path[0] + ".js", "utf8").split("\n");
+    var k = 1;
+    var done = false;
+    var line = 0;
+    while (!done) {
+        if (data[line].includes("'" + path[k].replace(new RegExp("'", "g"), "\\'") + "'")) {
```
ljharb: Please use a regex literal whenever possible

duktape.js
```diff
@@ -139,3 +144,44 @@ fs.readdirSync('.').forEach(function (filename) {
...
+        line++;
+    }
+    var indentStr = "";
+    for (var x = 0; x < path.length + 1; x++) {
+        indentStr = indentStr + "  ";
+    }
+    if (keyExists) {
+        data[line] = indentStr + dukKey + ": " + success + ",";
+    } else {
+        data.splice(line, 0, indentStr + dukKey + ": " + success + ",");
```
ljharb: is there no way to avoid using splice here?

data-es2016plus.js
```diff
@@ -3007,6 +3007,7 @@ exports.tests = [
         chrome66: true,
         duktape2_0: false,
         graalvm: true,
+        duktape2_3: false,
```
afmenez: It would be nice to keep all duktape entries together.

data-es2016plus.js
```diff
@@ -3187,6 +3188,7 @@ exports.tests = [
       graalvm: false,
       chrome73: true,
       chrome74: true,
+        duktape2_3: false,
```
afmenez: Indentation is wrong.

data-esnext.js
```diff
@@ -2193,6 +2202,7 @@ exports.tests = [
     ie11: false,
     firefox52: false,
     graalvm: false,
+      duktape2_3: false,
```
afmenez: Indentation is wrong.

data-esnext.js
```diff
@@ -2856,6 +2903,7 @@ exports.tests = [
     typescript1corejs2: typescript.fallthrough,
     typescript3_2corejs3: typescript.corejs,
     firefox68: true,
+      duktape2_3: false,
```
afmenez: Indentation is wrong.

data-esnext.js
```diff
@@ -2879,6 +2927,7 @@ exports.tests = [
     babel7corejs3: babel.corejs,
     typescript1corejs2: typescript.fallthrough,
     typescript3_2corejs3: typescript.corejs,
+      duktape2_3: false,
```
afmenez: Indentation is wrong.

data-esnext.js
```diff
@@ -2928,6 +2979,7 @@ exports.tests = [
...
```
afmenez: Indentation is wrong.

afmenez (commented):
Besides the cosmetic issues, I noticed you added 2.3=false to a lot of tests.  I believe they are also false to previous versions, are you able to test it with a previous version (2.0 for example) to have greater version coverage?

duktape.js
```diff
@@ -139,3 +144,44 @@ fs.readdirSync('.').forEach(function (filename) {
...
+            done = true;
+        } else {
+            line++;
+        }
+    }
+    while (!data[line].includes("res: {")) {
+        line++;
+    }
+    var keyExists = false;
+    while (!data[line].includes("}")) {
```
ljharb: This algorithm seems to be highly coupled to unenforced conventions - which seems brittle.

ljharb (changes requested):
It seems like it would be better to move the test results to a json file, rather than trying to programmatically edit a .js file.

data-es2016plus.js
TheBrokenRail: Fixed!

data-es2016plus.js
TheBrokenRail: Fixed!

data-esnext.js
TheBrokenRail: Fixed!

data-esnext.js
TheBrokenRail: Fixes!

data-esnext.js
TheBrokenRail: Fixed!

duktape.js
```diff
@@ -139,3 +144,76 @@ fs.readdirSync('.').forEach(function (filename) {
...
+    var path = fixTests[i].path;
+    var success = fixTests[i].success;
+    var data = fs.readFileSync(path[0] + ".js", "utf8").split("\n");
+    var k = 1;
+    var done = false;
+    var line = 0;
+
+    // Find Test
+    while (!done) {
+        if (data[line].includes("'" + path[k].replace(new RegExp(/'/g/, "g"), "\\'") + "'")) {
```
ljharb: sorry i wasn't clear; i mean that use of `new RegExp` is bad and should be avoided whenever possible in javascript. This can just be:
```suggestion
        if (data[line].includes("'" + path[k].replace(/'/g, "\\'") + "'")) {
```

data-esnext.js
TheBrokenRail: Fixed!

duktape.js
TheBrokenRail: Fixed!

duktape.js
TheBrokenRail: It is needed to insert data into an array instead of replacing it.

duktape.js
TheBrokenRail: Already Done!

duktape.js
ljharb: Can we replace it instead, and avoid the mutation?

duktape.js
ljharb: It's not done in your latest version; perhaps there's a typo?

duktape.js
```diff
@@ -139,3 +144,76 @@ fs.readdirSync('.').forEach(function (filename) {
...
+    var path = fixTests[i].path;
+    var success = fixTests[i].success;
+    var data = fs.readFileSync(path[0] + ".js", "utf8").split("\n");
+    var k = 1;
+    var done = false;
+    var line = 0;
+
+    // Find Test
+    while (!done) {
+        if (data[line].includes("'" + path[k].replace(/'/g/, "\\'") + "'")) {
```
ljharb: There's no slash after the flag in a JS regex literal.
```suggestion
        if (data[line].includes("'" + path[k].replace(/'/g, "\\'") + "'")) {
```

duktape.js
TheBrokenRail: Fixed!

PR #1453, opened 2019-04-24
Fixing the result values for Firefox features marked as "nomob"

afmenez:
"flagged" means that the feature is not available by default, but it can be turned on by the user (either by changing the browser configuration or adding a command line parameter).  This is not what the "nomob" nightly means: the feature is available on desktop builds only, and there's no way to enable it on mobile builds.

This PR reverts a change made in #1438, but had no relation with the issue being fixed.

### DIFF
--- a/data-common.json
+++ b/data-common.json
@@ -52,9 +52,9 @@
       "note_html": "The feature have to be enabled via \"javascript.options.shared_memory\" setting under <code>about:config</code>.  It is enabled by default in Firefox Developer Edition and Firefox Nightly builds."
     },
     "nomob": {
-      "val": "flagged",
+      "val": true,
       "note_id": "firefox-nomob",
-      "note_html": "The feature is available on desktop versions only, it is not available on mobile versions yet."
+      "note_html": "The feature is available on desktop builds only, it is not available on mobile builds."
     },
     "developer": {
       "val": "flagged",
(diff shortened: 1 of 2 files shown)


zloirock (changes requested):
I think that it would be better to rename `flagged` to something else what will include controversial cases.

afmenez:
I don't think "flagged" should go, we perhaps need a new value then, for features partially supported.  Notice that Firefox is the only browser where the features not available on all platforms are marked, the results for the other browsers simply assume the Desktop results are the ones that matter.

zloirock:
However, I think that `nomob` / `nightly` / `developer` should be shown in the table like `flagged` now - as partially supported with `false` result.

zloirock:
We need some kind of value which will mean "supported in some cases".

afmenez:
My point exactly.

afmenez:
Actually, most Firefox users are on Desktop.  By most I mean something like > 98%.  If the feature is available for >98% of the users, the result should be true, with a note saying where it is not.  As it was before, and as this PR reverts.

zloirock:
"> 98%" does not mean "all" which by my meaning required for `true`. At least, it could cause problems for tools like `preset-env`.

afmenez:
OK, in that case I believe "flagged" is good enough.  I am dropping this one.

PR #1454, opened 2019-04-24
Fixing the result values for Firefox features marked as "nightly" and "developer"

afmenez:
"flagged" means that the feature is not available by default, but it can be turned on by the user (either by changing the browser configuration or adding a command line parameter).
This is not what "nightly" means: it is available by default on nightly builds (via compilation flags), and not available at all on other builds. Changing this to "flagged" marked lots of features as optionally available on older versions, which is not true.

This PR reverts a change made in #1438, which was trying to mark BigInt as a "flagged" on nightly builds for Firefox 68, but the feature is not #ifdef as nightly anymore (see #1448 and #1449).

### DIFF
--- a/data-common.json
+++ b/data-common.json
@@ -57,12 +57,12 @@
       "note_html": "The feature is available on desktop versions only, it is not available on mobile versions yet."
     },
     "developer": {
-      "val": "flagged",
+      "val": false,
       "note_id": "firefox-developer",
       "note_html": "The feature is available only in Firefox Developer Edition and Firefox Nightly builds."
     },
     "nightly": {
-      "val": "flagged",
+      "val": false,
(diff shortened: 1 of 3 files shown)


zloirock (changes requested):
Similarly https://github.com/kangax/compat-table/pull/1453#pullrequestreview-230396239

afmenez:
I don't think the situation is the same as the "nomob" patch, and that's why I did a separate patch.  For the "nomob", the feature is partially available on all the development stages.  For "nigthly" the feature is only available on the latest version, and not available when it progresses to more stable stages (there are currently 11 FF versions on the table, the result would only be correct for one).

afmenez:
I changed my view of the "flagged" state.

PR #1466, opened 2019-06-08
Updating links to the current version of the respective proposal

afmenez:
Ditto.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -2206,7 +2206,7 @@ exports.tests = [
   }
 },
 {
-  name: 'Getting last element from array',
+  name: 'Getting last item from array',
   spec: 'https://github.com/keithamus/proposal-array-last',
   category: STAGE1,
   significance: 'small',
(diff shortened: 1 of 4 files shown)


data-esnext.js
```diff
@@ -2206,7 +2206,7 @@ exports.tests = [
   }
 },
 {
-  name: 'getting last item from array',
+  name: 'Getting last element from array',
```
zloirock: In the case of `.lastIndex`, I don't think that "element" better.

zloirock: I think it would be better just revert it for consistency with the header of the proposal repo.

ljharb (approved):

data-esnext.js
afmenez: I copied the title from here: https://github.com/tc39/proposals/blob/master/stage-1-proposals.md
Fixed.

chicoxyzzy (approved):

PR #1473, opened 2019-06-24
WeakReferences to Stage 3

afmenez:
The WeakReferences proposal advanced to Stage 3.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -503,6 +503,23 @@ exports.tests = [
         duktape2_0: false,
         graalvm: false,
       }
+    },
+    {
+      name: 'Finalizers minimal support',
+      spec: 'https://github.com/tc39/proposal-weakrefs#finalizers',
+      exec: function(){/*
+        var fg = new FinalizationGroup(function() {});
+        return Object.getPrototypeOf(fg) === FinalizationGroup.prototype;
+      */},
+      res : {
+        ie11: false,
+        firefox2: false,
+        opera10_50: false,
+        chrome65: false,
+        chrome74: {val: 'flagged', note_id: "chrome-weakrefs"},
+        duktape2_0: false,
+        graalvm: false,
+      }
     }
   ]
 },
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -481,25 +481,31 @@ exports.tests = [
   }
 },
 {
-  name: 'weak references',
+  name: 'WeakReferences',
   spec: 'https://github.com/tc39/proposal-weakrefs',
   category: STAGE3,
   significance: 'large',
-  exec: function(){/*
-    var O = {};
-    var weakref = System.makeWeakRef(O);
-    var works = weakref.get() === O;
-    weakref.clear();
-    return works && weakref.get() === undefined;
-  */},
-  res : {
-    ie11: false,
-    firefox2: false,
-    opera10_50: false,
-    chrome65: false,
-    duktape2_0: false,
-    graalvm: false,
-  }
+  subtests: [
+    {
+      name: 'Weak references',
+      spec: 'https://github.com/tc39/proposal-weakrefs#weak-references',
+      exec: function(){/*
+        var O = {};
+        var weakref = new WeakRef(O);
+        var works = weakref.deref() === O;
+        weakref.clear();
+        return works && weakref.deref() === undefined;
+      */},
+      res : {
+        ie11: false,
+        firefox2: false,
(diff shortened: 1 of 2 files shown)


ljharb (changes requested):
The stage 3 proposal no longer has System at all; we should update the tests before/while advancing the stage.

afmenez:
Sorry, I am not sure I know enough, and there's no implementation of the spec to test it.  Would this be the correct update of the test?
```
    var O = {};
    var weakref = new WeakRef(O);
    var works = weakref.deref() === O;
    weakref.clear();
    return works && weakref.deref() === undefined;
```
I noticed that there are no tests for the [`Finalizers`](https://github.com/tc39/proposal-weakrefs#finalizers) part of the proposal.  I'll move the test above to a subtest and open a new issue for the missing one.

ljharb:
I think that looks good - at the worst it's much closer than the current test :-)

ljharb (approved):

afmenez:
The proposal says in the end that it is implemented on latest chrome builds, so I have marked it as such.

data-esnext.js
```diff
@@ -481,25 +481,32 @@ exports.tests = [
...
-  }
+  subtests: [
+    {
+      name: 'Weak references',
+      spec: 'https://github.com/tc39/proposal-weakrefs#weak-references',
+      exec: function(){/*
+        var O = {};
+        var weakref = new WeakRef(O);
+        var works = weakref.deref() === O;
+        weakref.clear();
```
chicoxyzzy: There is no such method as `WeakRef.prototype.clear`

chicoxyzzy (changes requested):

data-esnext.js
```diff
@@ -481,25 +481,32 @@ exports.tests = [
...
+        return works && weakref.deref() === undefined;
+      */},
+      res : {
+        ie11: false,
+        firefox2: false,
+        opera10_50: false,
+        chrome65: false,
+        chrome77: {val: 'flagged', note_id: "chrome-weakrefs", note_html: "Available behind the <a href='https://bugs.chromium.org/p/v8/issues/detail?id=8179'><code>--harmony-weak-refs</code></a> flag in V8."},
```
chicoxyzzy: It's implemented in Chrome 74 with flag `--js-flags=--harmony-weak-refs --expose-gc`

chicoxyzzy:
I think we can try to adopt example from [here](https://github.com/JSKongress/JS-Kongress-Munich-Deep-Track/issues/20#issuecomment-477750040). There is also https://github.com/tc39/test262/pull/2192

data-esnext.js
afmenez: I'll fix it.

data-esnext.js
afmenez: This is in the current test, I missed it in my revision.  How about splitting this test in two, one testing only the WeakRef() creation and other testing the persistence after a gc()?

afmenez:
There are A LOT of tests there, and they are too much for me.  How about I fix the remaining issues with the current test, and someone with more knowledge can write the appropriate tests in a separate issue? There's even #1476 already open.

chicoxyzzy:
Sure, you don't need to move all tests from test262 (there could be some some small tests we can take though). I think that fixing current example could work as a good starting point for now.

data-esnext.js
afmenez: Done.

data-esnext.js
afmenez: I couldn't make the gc() part work in Chrome, so I left it out.  I'll let someone more familiar with it write additional tests.

afmenez:
OK, I fixed the test and added a minimal test for the Finalizers part.  I believe this is ready for merging, but not enough to close #1476 yet.

data-esnext.js
```diff
@@ -484,24 +484,39 @@ exports.tests = [
   name: 'WeakReferences',
   spec: 'https://github.com/tc39/proposal-weakrefs',
   category: STAGE3,
-  significance: 'large',
+  significance: 'medium',
```
ljharb: why the downgrade? this is a fundamentally new capability.

ljharb (approved):

data-esnext.js
afmenez: No good reason.  I'll change it back.

chicoxyzzy (approved):
Tested in Chrome Canary with WeakRef flags

LGTM

PR #1456, opened 2019-05-06
Update TemplateStrings bug test

jridgewell:
https://github.com/kangax/compat-table/pull/1424 for context.

The Safari bug revealed that `getStrings() !== new getStrings()`, making
this easily testable. I've separated the original test into two:

1. Test TemplateStrings call site [revision](https://github.com/tc39/ecma262/pull/890)
2. Test the buggy Safari behavior explicitly

- - -

Is there some way to run old JS environments to see when the browsers changed behaviors?

PR #1477, opened 2019-06-27
Update Safari Template Strings Permanent Caching results

### LINKED ISSUE #1472: TemplateStrings permanent caching is fixed in Safari?
Tests are passing in Safari 13 beta and bug [seems to be fixed in WebKit Bugzilla](https://bugs.webkit.org/show_bug.cgi?id=190756), though version mentioned there is Safari 12. I don't have Safari 12 nor 12.1 on my devices. Can someone check if tests are passing for these versions?
ref: https://github.com/kangax/compat-table/pull/1424

chicoxyzzy:
[Bug](https://github.com/kangax/compat-table/pull/1456) was fixed in Safari 13 Beta

closes #1472

### DIFF
--- a/data-common.json
+++ b/data-common.json
@@ -114,6 +114,13 @@
       "note_html": "The feature is considered unstable, but can be enabled via <code>--js-flags=\"--harmony-simd\"</code> flag"
     }
   },
+  "safari": {
+    "templateStringsPermanentCaching": {
+      "val": false,
+      "note_id": "safari-templ-cahing",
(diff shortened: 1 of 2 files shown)


ljharb (approved):

chicoxyzzy:
cc @jridgewell

afmenez:
When a note is used only once, does it make sense to put it on data-common.json?  Why not put it inside the test?

jridgewell (approved):

chicoxyzzy:
@afmenez I don't think that it's really matters where to put the note.

@jridgewell thank you for review!

PR #1481, opened 2019-07-08
Firefox 68 released

afmenez:
Firefox 68 will be released tomorrow.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -814,7 +814,7 @@
     "obsolete": "very"
   },
   "firefox10": {
-    "full": "Firefox 10 ESR",
+    "full": "Firefox 10 Extended Support Release",
     "family": "SpiderMonkey",
     "short": "FF 10 ESR",
     "release": "2012-01-31",
@@ -870,7 +870,7 @@
     "obsolete": "very"
   },
   "firefox17": {
-    "full": "Firefox 17 ESR",
+    "full": "Firefox 17 Extended Support Release",
     "family": "SpiderMonkey",
     "short": "FF 17 ESR",
     "release": "2012-11-20",
@@ -926,7 +926,7 @@
     "obsolete": "very"
   },
   "firefox24": {
-    "full": "Firefox 24 ESR",
+    "full": "Firefox 24 Extended Support Release",
     "family": "SpiderMonkey",
     "short": "FF 24 ESR",
     "release": "2013-09-17",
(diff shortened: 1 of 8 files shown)


chicoxyzzy (commented):
Numeric Separators and Promise.allSettled are not available by default in Firefox 68 and Firefox 69

environments.json
```diff
@@ -1278,21 +1278,28 @@
     "obsolete": false
   },
   "firefox68": {
-    "full": "Firefox 68 Beta",
+    "full": "Firefox 68",
```
chicoxyzzy:
```suggestion
    "full": "Firefox 68 ESR",
```

environments.json
```diff
@@ -1278,21 +1278,28 @@
...
     "family": "SpiderMonkey",
-    "short": "FF 68 Beta",
+    "short": "FF 68",
```
chicoxyzzy:
```suggestion
    "short": "FF 68 ESR",
```

chicoxyzzy (changes requested):

environments.json
afmenez: 68 ESR wasn't released yet.

environments.json
afmenez: 68 ESR wasn't released yet.

afmenez:
You are correct, both features  seem to be "beta and nightly only" for now.  I'll update the patch later today.

environments.json
chicoxyzzy: https://www.mozilla.org/en-US/firefox/68.0esr/releasenotes/

chicoxyzzy:
> both features seem to be "beta and nightly only" for now

Both are disabled in Beta and Developer Edition and enabled in Nightly only

afmenez:
Fixed all the wrong results (found another one on the non-standard page).  The 68 ESR builds are not released yet, we can update this later (in a separate PR).

chicoxyzzy:
ESR was released see https://github.com/kangax/compat-table/pull/1481#discussion_r301681697

afmenez:
Yep, the release notes for all versions is usually available in advance, but the installer it is not available on the [main download page](https://www.mozilla.org/en-US/firefox/organizations/) yet.
Also, current 60 ESR are not being upgraded yet.
The ESR is  usually released only after the main release is already "unthrottled" and is considered trusted enough.  It takes a few days.

chicoxyzzy:
I don't see any reasons why we shouldn't add ESR now.

> Also, current 60 ESR are not being upgraded yet.

It will get updates until 70 / 68.2 ESR will be released according to calendar https://wiki.mozilla.org/Release_Management/Calendar

chicoxyzzy:
Also you can download 68 ESR here: https://www.mozilla.org/en-US/firefox/organizations/all/?q=English%20(US)

environments.json
afmenez: Done.

chicoxyzzy (approved):

PR #1484, opened 2019-07-26
Promise.allSettled to Stage 4

afmenez:
Just moving the test.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -3486,6 +3486,35 @@ exports.tests = [
       graalvm: true,
     }
   },
+  {
+    name: 'Promise.allSettled',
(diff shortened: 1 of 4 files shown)


afmenez:
Actually, I am just following what was done for all previous Stage 4 proposals, BUT I believe it would be more correct to create a Stage4 category on the ESNext tab and keep the proposals there until ES2020 is actually released.

ljharb:
Why? It’s a living standard; whatever’s on master is in the language. The yearly edition is just an immediately obsolete snapshot.

ljharb (approved):

chicoxyzzy (approved):

PR #1494, opened 2019-08-10
Store Test Results In JSON

### LINKED ISSUE #1493: Store Test Results In JSON
This would allow test runners to automatically save their data, instead of requiring someone to manually input it.

TheBrokenRail:
Fixes #1493


build.js
```diff
@@ -548,7 +560,7 @@ function dataToHtml(skeleton, rawBrowsers, tests, compiler) {
       .append($('<td></td>')
         .attr('id', getHtmlId(id))
         .append('<span><a class="anchor" href="#' + getHtmlId(id) + '">&sect;</a>' + name + footnoteHTML(t) + '</span></td>')
-        .append(testScript(t.exec, compiler, rowNum++))
+        .append(testScript(eval(t.exec), compiler, rowNum++))
```
ljharb: we reallllly shouldn't use `eval`. why is this needed?

ljharb (changes requested):
The tests themselves should be in .js, since they're code. Only the *data* should (maybe) be in json

build.js
TheBrokenRail: Because ```testScript``` expects a function.

ljharb: That’s why the function should be stored in a .js file - functions aren’t data, and json is for data.

TheBrokenRail: It now stores the tests in JS and the results in JSON.

zloirock:
Now:
image
This PR:
image

zloirock:
@TheBrokenRail any progress?

PR #1495, opened 2019-08-29
Disable WeakRefs by default in Safari

chicoxyzzy:
WeakRefs were disabled by default in Safari 13 and Safari TP. There is a flag to turn them on probably, but I have no information about any yet.


ljharb (approved):
any background on why?

chicoxyzzy:
@ljharb perhaps `WeakRef` constructor was enabled by default in some WebKit build by mistake, but I don't know to be honest

PR #1507, opened 2019-10-04
Non-standard Array generics removed from Firefox 71

afmenez:
Non-standard Array generics have been removed from Firefox 71:
https://www.fxsitecompat.dev/en-CA/docs/2019/non-standard-array-generics-have-been-removed/


ljharb (approved):
about time

PR #1510, opened 2019-10-07
Promise.any moved to Stage 3

afmenez:
Promise.any moved to Stage 3 last week.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -3643,6 +3643,7 @@ exports.tests = [
       typescript1corejs2: typescript.fallthrough,
       typescript3_2corejs3: typescript.corejs,
       firefox68: firefox.nightly,
+      firefox71: true,
       chrome76: true,
       safari13: true,
     }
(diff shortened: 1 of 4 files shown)


ljharb (approved):

chicoxyzzy (approved):

data-es2016plus.js
```diff
@@ -3643,6 +3643,7 @@ exports.tests = [
       typescript1corejs2: typescript.fallthrough,
       typescript3_2corejs3: typescript.corejs,
       firefox68: firefox.nightly,
+      firefox71: true,
```
chicoxyzzy, line 3646: Wait, I'm not sure this is true

chicoxyzzy (changes requested):

chicoxyzzy:
`Promise.any` is not implemented in Firefox yet
https://bugzilla.mozilla.org/show_bug.cgi?id=1568903

data-es2016plus.js
afmenez: Yes, it is, but should be a separate PR.  My mistake, I based this one on the other instead of the master.  I'll revert it.

afmenez:
Bad branching, sorry.  Dropping this and sending a new one.

PR #1514, opened 2019-10-09
TypeScript 3.7 Beta with ?. and ?? support

afmenez:
TypeScript 3.7 will support both ?. and ?? proposals:
https://devblogs.microsoft.com/typescript/announcing-typescript-3-7-beta/

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -486,7 +486,19 @@
     "platformtype": "compiler",
     "release": "2019-08-28",
     "obsolete": false,
-    "unstable": false,
+    "test_suites": [
+      "es6",
+      "es2016plus",
+      "esnext"
+    ]
+  },
+  "typescript3_7corejs3": {
+    "full": "TypeScript 3.7 Beta + core-js 3",
+    "short": "Type-<br />Script +<br /><nobr>core-js 3</nobr>",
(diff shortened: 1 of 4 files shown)


environments.json
```diff
@@ -486,7 +486,19 @@
...
     "obsolete": false,
-    "unstable": false,
+    "test_suites": [
+      "es6",
+      "es2016plus",
+      "esnext"
+    ]
+  },
+  "typescript3_7corejs3": {
+    "full": "TypeScript 3.7 Beta + core-js 3",
```
ljharb, line 496: why is core-js required for these features?

afmenez: I have no idea.  It has ever been like this.  Maybe because it is needed to run the test scripts?

ljharb: it's needed for some features, ofc, but not for syntax.

zloirock: I can't understand the claim. It's the name of the column, which also contains `core-js` data, but not the field with a new syntax only.

ljharb: i'm just confused why the column needs to include core-js, rather than the individual features requiring it being flagged "yes, with core-js". doesn't have to block this PR, ofc.

PR #1521, opened 2019-10-21
Mark Node 6 as obsolete

zloirock:


ljharb (approved):
I still don't believe that EOL has any relationship to something being obsolete (#1288), but it's been out of maintenance since April (https://github.com/nodejs/Release/blob/750a443~1/schedule.svg) so this is fine.

chicoxyzzy (approved):

PR #1523, opened 2019-10-22
Firefox 70 released

afmenez:
Firefox 70 was released today.  No surprises.
The PR also adds some missing test results for older versions of Firefox.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -2693,6 +2693,7 @@ exports.tests = [
     babel7corejs3: babel.corejs,
     typescript1corejs2: typescript.fallthrough,
     typescript3_2corejs3: typescript.corejs,
+    firefox60: false,
   }
 },
 {
@@ -2711,6 +2712,7 @@ exports.tests = [
         babel7corejs3: babel.corejs,
         typescript1corejs2: typescript.fallthrough,
         typescript3_2corejs3: typescript.corejs,
+        firefox60: false,
       },
     },
     {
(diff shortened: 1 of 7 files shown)


data-esnext.js
```diff
@@ -2693,6 +2693,7 @@ exports.tests = [
     babel7corejs3: babel.corejs,
     typescript1corejs2: typescript.fallthrough,
     typescript3_2corejs3: typescript.corejs,
+    firefox60: false,
```
zloirock, line 2696: Why FF60?

afmenez: I tested with 60 because is was the oldest still present in the table.

zloirock: In the table, but not in historical data. If it's a problem of testing on real hardware, you can test it, for example, with saucelabs you can test it on FF4+, with browserstack - on FF3+. They are free for FOSS.

zloirock: It was just a question, not a blocker for this PR.

zloirock (approved):

data-esnext.js
afmenez: I actually have every version since 2.0 installed, but older versions don't work with the table (there's an old bug #1127 about this still open).  I'll run all the tests in older versions in a separate PR.

afmenez: Also notice that until #1131 is commited, the change doesn't affect the displayed results.

chicoxyzzy (approved):

data-esnext.js
zloirock: Yep, I know about #1131. But theoretically some third-party tool could do something with this data too.

PR #1527, opened 2019-10-23
Adding Chrome 77 results

afmenez:
There are a lot of tests with no Chrome results, this PR fills them.  Since the engine already considers missing results as 'false', there is no change in the tables.


ljharb:
@afmenez confirming that you personally verified these failing results? (`true` or `false` means a human checked it, "missing" means we don't actually know for sure)

afmenez:
Yep, I personally verified all the results.

zloirock (approved):

PR #1528, opened 2019-10-23
Adding Firefox 10 results

afmenez:
I tried Firefox 2, 3, 3.6 and 4, they all failed to display the results (probably due to #1127).  The earliest version I was able to run was Firefox 10 ESR.


data-es6.js
```diff
@@ -7009,6 +7009,8 @@ exports.tests = [
         es6shim: true,
         ie11: true,
         firefox2: false,
+        firefox10: true,
```
zloirock: Hmmm... Even in MDN I see that `Map` is available only from FF13... I'll check it.

zloirock: image

zloirock (changes requested):

data-es6.js
afmenez: Yep, and the the test fails on FF17.

zloirock: It's FF10, result - `false`.

afmenez: I have rerun the tests now and indeed it is false.  I'll fix the PR.

afmenez: I see now what I did wrong, I mixed Map and WeakMap results.  I'll rerun everything and update the PR.

afmenez: Fixed.

zloirock (approved):

PR #1532, opened 2019-10-31
Add JerryScript

### LINKED ISSUE #1479: Add JerryScript?
Would you be interested in adding the JerryScript engine (https://github.com/jerryscript-project/jerryscript) to the project? I saw that Duktape, another JS engine for embedded systems, is already included. I've prototyped a test runner for JerryScript based on the existing duktape.js runner (https://github.com/akosthekiss/compat-table/commits/jerryscript). Let me know if this could be integrated.

ossy-szeged:
Fixes #1479

### DIFF
--- a/data-es5.js
+++ b/data-es5.js
@@ -31,6 +31,7 @@ exports.tests = [
       nashorn1_8: true,
       nashorn9: true,
       graalvm: true,
+      jerryscript1_0: true,
     }
   },
   {
@@ -56,6 +57,7 @@ exports.tests = [
       nashorn1_8: true,
       nashorn9: true,
       graalvm: true,
+      jerryscript1_0: true,
     },
   },
   {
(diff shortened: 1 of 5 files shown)


zloirock (changes requested):
Mabe makes no sense to show it in `es2016+`, `esnext` and `intl` (not sure about this table) tables since here zero support and just some percents of `es6`?

ossy-szeged:
> Mabe makes no sense to show it in `es2016+`, `esnext` and `intl` (not sure about this table) tables since here zero support and just some percents of `es6`?

I agree with you about omitting es2016+, esnext and intl. But JerryScript already has progressing ES6 support, latest release has 34% (DUK 2.3 has 26%, but included in the table), and we are working on adding more ES6 features. So I would like to include ES5 and ES6.

PR will be updated immediately.

zloirock (approved):

PR #1537, opened 2019-11-08
Firefox now implements String.prototype.replaceAll

afmenez:
Firefox 72 implements `String.prototype.replaceAll`:
https://bugzilla.mozilla.org/show_bug.cgi?id=1540021

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -2164,6 +2164,8 @@ exports.tests = [
       ie11: false,
       firefox10: false,
       firefox52: false,
+      firefox71: false,
+      firefox72: true,
       chrome77: false,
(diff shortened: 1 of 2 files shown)


ljharb:
That’s just been merged, with a target milestone - seems a bit early to mark it as true. Did you try it out in an actual build of v72?

afmenez:
I did, it works in today's build.

ljharb (approved):

chicoxyzzy:
This change is incorrect. The feature was enabled for Nightly versions only.

afmenez:
True, I haven't noticed the #ifdef on the patch.

PR #1544, opened 2019-11-21
Safari TP 76 release

chicoxyzzy:


ljharb (changes requested):
seems like there's an extra commit in here?

ljharb (approved):

PR #1547, opened 2019-11-24
Mark as obsolete early Node 12

zloirock:


environments.json
```diff
@@ -2714,6 +2714,7 @@
     "note_id": "harmony-flag",
     "equals": "chrome62",
     "release": "2018-03-07",
+    "retire": "2019-12-31",
```
ljharb, line 2717: generally we don’t put future dates in here; I’d prefer to wait until after something is retire before marking it as such

ljharb (changes requested):

environments.json
zloirock: It's official dates from [Node releases schedule](https://github.com/nodejs/Release/blob/master/schedule.json). Those fields are not used in the generation of tables. So it's just a reminder when they should be marked as obsolete if we'll forget it.

zloirock: About "generally we don’t put future dates in here" - take a look, for example, at FF ESR.

ljharb: fair enough, if none of the tools read it then it doesn't matter that it will drift.

ljharb (approved):

PR #1558, opened 2019-12-08
Revert "Revert "Improve historical Node data""

zloirock:
Reverts kangax/compat-table#1557


ljharb:
At this point we’re just going to go back and forth until one or both of us gets removed as collaborators.

Nothing lands in master that any collab blocks, this included.

zloirock:
Sorry, but you don't do anything for this project, only exhibit destructive activity, so your opinion does not matter here.

ljharb:
That’s not your decision to make - we’re all collaborators so all of our opinion matters equally, and without consensus, *nothing happens*.

PR #1559, opened 2019-12-08
String.prototype.matchAll fixed on Firefox 73

afmenez:
String.prototype.matchAll now throws on non-global regEx.


zloirock (approved):

chicoxyzzy:
The first two commits in this PR are not related to `String.prototype.matchAll`

afmenez:
Yep, and they are not even mine.  Don't know how they appeared here.

zloirock:
Seems @ljharb forced pushed to `gh-pages` :-D

ljharb:
indeed; those commits shouldn't be on master.

zloirock:
They should be.

PR #1560, opened 2019-12-09
Mark Chrome 79 as stable

zloirock:


ljharb (approved):

chicoxyzzy:
Chrome 79 stable is not released yet https://omahaproxy.appspot.com

ljharb:
lol, guess this should be reverted then

zloirock:
Yep, I don't know why @ljharb merged it too early. Anyway, it's not a problem since it will be released today. I don't think that any actions required.

ljharb:
Fair, i assumed i could trust the PR title, and that a stable PR wouldn’t go up until it was actually stable. I’ll be more careful in the future

PR #1561, opened 2019-12-10
Feature/sticky header improvement

everdimension:
Use CSS for sticky header.
Remove jquery plugin.

I made screen recordings to compare before and after, but files are too large for github ¯\_(ツ)_/¯


zloirock:
(http://raw.githack.com/everdimension/compat-table/feature/sticky-header-improvement/es6/index.html)

zloirock:
IE11:
image

The header is not sticky and overlaps the top rows of the table.

zloirock (changes requested):
^^^

everdimension:
@zloirock Right, thanks for noticing.
Fixed the overlapping.

I believe it's alright to lose the sticky behavior in ie11 at this time.

zloirock:
Seems not fixed, I don't see any changes in IE11 -/

everdimension:
@zloirock Perhaps you're seeing cached results

Here's what I see in IE11 after the update:
image

zloirock:
I checked it on another VM. Ok, let's check it again...

zloirock:
Ok, fixed, it seems it was cached on github / githack side.

zloirock (approved):

master.js
```diff
@@ -95,19 +77,24 @@ $(function() {
...
   };
 
-
-
-  table.on("floatThead", function (evt, isFloating, $container) {
-    $container.toggleClass("floating-header", isFloating);
-  });
-
+  /** Sticky header */
+  var platformTypesHeader = document.querySelector(
```
zloirock: Since it will not work in ancient engines, maybe it would be better to use jQuery like in other parts of this script?

everdimension: I don't think there are browsers out there which don't support this. The latest browser that doesn't is IE7: https://caniuse.com/#feat=queryselector

And since compat-table is primarily a developer-facing tool, I believe we don't have to support this.

The only argument for using jquery here would be consistency, but since this piece of code is quite isolated, I don't think it's an issue. What do you think?

zloirock: I think that it would be better to use jQuery because of both reasons.

ljharb: I agree; jquery plugins are best avoided, but jquery itself is quite useful.

everdimension: 🤷‍♀

chicoxyzzy (approved):

sedghi:
should the `feature name` row be sticky? since it is not right now

PR #1566, opened 2020-01-07
fix Firefox 74 version number

chicoxyzzy:
Before that PR I somehow made a commit to upstream gh-pages. Sorry for that. I think that was because I've tried to track upstream branch after rebases on it. Remote branch is fixed now so I add a small fix in this PR.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -1351,9 +1351,9 @@
     "unstable": true
   },
   "firefox74": {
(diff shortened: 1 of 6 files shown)


chicoxyzzy:
Previous commit https://github.com/kangax/compat-table/commit/a1bed140e00afebff0b698158f69790e2344f31c

ljharb (approved):

PR #1568, opened 2020-01-09
[ESNext] Add `Object` iteration proposal

zloirock:


ljharb:
tbh it seems premature for a proposal earlier than stage 1 to have tests and any `true` values, since it’s not until stage 2 that the shape of the solution is finalized.

zloirock:
Seems you forgot that `compat-table` contains tests for features from all stages, not only for 2+. Moreover, tests for stage 1 features more than others. It's the obvious behavior for proposed methods. If you wanna tests only for stable feature - it's `test262`.

ljharb:
True; really what I'm suggesting is to remove tests for everything prior to stage 2 since they're utterly meaningless, but obviously then there'd be no point in the compat table listing stage < 2 proposals (which it probably shouldn't, but is a much larger discussion than this PR).

ljharb (approved):

PR #1569, opened 2020-01-15
Edge 79 release

chicoxyzzy:
https://blogs.windows.com/windowsexperience/2020/01/15/new-year-new-browser-the-new-microsoft-edge-is-out-of-preview-and-now-available-for-download/

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -698,55 +698,6 @@
       "non-standard"
     ]
   },
-  "edge12": {
-    "full": "Microsoft Edge 12",
-    "family": "Chakra",
-    "short": "Edge 12",
-    "release": "2015-07-15",
-    "obsolete": "very"
-  },
-  "edge13": {
-    "full": "Microsoft Edge 13",
-    "family": "Chakra",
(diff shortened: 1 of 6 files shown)


chicoxyzzy:
I moved all Edge versions below Chrome to make `equals` field work

ljharb:
Have you downloaded it and verified that it passes all the chrome 79 tests?

chicoxyzzy:
@ljharb yep. Tested on macOS 10.15

ljharb (approved):

mgol:
What's the reason to keep Chrome & Opera grouped & Edge 79+ separate when they are all based on Chromium?

chicoxyzzy:
I also thought about that. Do you think it'll be better to group Edge also?

mgol:
For me it makes sense to keep all Chromium browsers in one entry as long as none of them does modifications to the JS engine by itself and AFAIK that hasn't happened so far. Obviously, Edge 18 & older need to stay separate.

PR #1570, opened 2020-01-15
Rename "destructuring, parameters / defaults, arrow function"

nicolo-ribaudo:
It only tests shorthand defaults in object destructuring inside parameters, so
"shorthand defaults, arrow function" describes it more closely.

I propose to rename this test instead of also testing for `x: y = z` because the
latter works in edge15-17, while the shorthand version doesn't.


ljharb:
Do we then need a separate test for the shorthand version?

nicolo-ribaudo:
Good idea, but I don't have any old browser to test it.

ljharb:
I'm happy to load up old edge versions in browserstack to add some results if you make the test first :-)

nicolo-ribaudo:
:warning: The data pushed in the last commit is not manually verified. I copied engine support from the other tests, except for Edge. I don't know if the first working version was 14 or 15.

ljharb:
hmm, looks like I only have edge 15 and 11 available ("aliased defaults, arrow function" works in 15, syntax errors in 11). Let's clear out the 13/14 results for now.

nicolo-ribaudo:
Is it possible to mark `12-14` as unknown? Or should I mark `14` as not working?

ljharb:
We currently just leave them unspecified (ie, delete the entries)

PR #1573, opened 2020-01-20
Safari test for Safari bug with block scoped variables

nicolo-ribaudo:
```js
function f(e) {
  for (let e in {}) e
}
```

This code throws in old Safari versions about `e` being redeclared.
We will soon use compat data about this test in Babel, for https://github.com/babel/preset-modules/blob/master/src/plugins/transform-safari-for-shadowing/index.js.

I only added data for engines I was sure about:
- Safari was verified by @developit.
- Firefox 2 and Opera 10.50 haven't been verified, but I noticed that every other test for `let` included them, and they don't support `let` anyway.
- Everything else was manually tested by me.


chicoxyzzy (approved):

ljharb:
I checked; this does not throw in Firefox 43 or 44. It also does not throw in v40, v30, v20, v10, v5, or v4, but it's a syntax error in v3.6 and v3.0. Can you update to add those results?

ljharb (approved):

PR #1572, opened 2020-01-20
Firefox 74 supports WeakRefs

chicoxyzzy:
Firefox 74 Nightly got WeakRefs support under flag

PR #1574, opened 2020-01-21
Firefox 74 implements the optional chaining operator (?.)

afmenez:
Ditto.


ljharb:
I probably saw the same comment as you indicating that it *would* be in FF 74, but 74 isn't released yet. Should we wait until then to merge this? (#1572 is under a flag, so i don't think that needed to wait)

afmenez:
This feature was [already merged](https://bugzilla.mozilla.org/show_bug.cgi?id=1566143), and it is already working on today's nightly build.
What's the point of having "unstable" versions if we have to wait until release to update their features?

chicoxyzzy (approved):

chicoxyzzy:
@ljharb I think you are talking about [this tweet](https://twitter.com/nicolaschevobbe/status/1219517536645066752). It mentions that the feature will be shipped tonight. And it was shipped =) AFAICT it is shipped without any flags, so changes in this PR LGTM

ljharb (approved):

PR #1576, opened 2020-01-22
Move IE engine right before Edge

### LINKED ISSUE #1575: [regression] Pre-chromium edge doesn't show correct data for let/const
Ref: https://github.com/babel/babel/pull/11039#discussion_r369237851

Old comapt-table versions correctly showed Edge 16 as supporting `let`/`const`: https://web.archive.org/web/20191113231919/https://kangax.github.io/compat-table/es6/
The current version doesn't.

According to `git bisect` the regression has been introduced by 44c16dd2e3b2d16ad8280a5edd23fc48def434c1 (@existentialism You were right :stuck_out_tongue:)

I'm investigating a fix.

nicolo-ribaudo:
Compatibility data are only inherited if the previous version of an engine is right before the current version.
Edge 12 should default to what IE 11 supports, so they must be next to each other in the environments list.

Fixes #1575

This is the relevant code in the build script: https://github.com/kangax/compat-table/blob/d6680df01bc90bdca73e7c56a7588acd7fe103bf/build.js#L434-L444

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -2174,7 +2174,9 @@
     "family": "Chakra",
     "short": "Edge 12",
     "release": "2015-07-15",
-    "obsolete": "very"
+    "obsolete": "very",
+    "#": "IE11 and Edge12 are not equal, but we can use IE11 as a baseline",
+    "equals": "ie11"
   },
   "edge13": {
     "full": "Microsoft Edge 13",
--- a/es2016plus/index.html
+++ b/es2016plus/index.html
@@ -7162,9 +7162,9 @@ return (function(){
 <td class="tally" data-browser="chrome79" data-tally="1">16/16</td>
 <td class="tally unstable" data-browser="chrome80" data-tally="1">16/16</td>
 <td class="tally unstable" data-browser="chrome81" data-tally="1">16/16</td>
-<td class="tally obsolete" data-browser="edge15" data-tally="0.25" style="background-color:hsl(30,75%,50%)">4/16</td>
-<td class="tally obsolete" data-browser="edge17" data-tally="0.375" style="background-color:hsl(45,69%,50%)">6/16</td>
-<td class="tally" data-browser="edge18" data-tally="0.375" style="background-color:hsl(45,69%,50%)">6/16</td>
+<td class="tally obsolete" data-browser="edge15" data-tally="0.75" style="background-color:hsl(90,53%,50%)">12/16</td>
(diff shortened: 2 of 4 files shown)


ljharb:
hm, does this change the visible order of engines in the rendered table?

ljharb:
Perhaps it'd be better to update the build script to not care about the order of engines in an object, which is an inherently unsorted thing anyways?

nicolo-ribaudo:
I pushed an alternative fix that doesn't rely on Edge being right after IE.

---

If you prefer, I can try to remove _all_ the ordering constraints in the list, also for subsequent versions of the same browser, or I can refactor the engines list to use an array so that the ordering is explicit.

Something like this:
```json
{
  "chrome": [
    {
      "id": "chrome1",
      "full": "Chrome 1",
      "family": "V8",
      "short": "CH 1",
      "release": "2008-12-11",
      "retire": "2009-05-24",
      "obsolete": "very"
    },
    {
      "id": "chrome2",
      "full": "Chrome 2",
      "family": "V8",
      "short": "CH 2",
      "release": "2009-05-24",
      "retire": "2009-10-12",
      "obsolete": "very"
    },
  ],
  "firefox": [
    /* ... */
  ]
}
```

If it should be done, I would prefer to do it in a separate PR, because integrating it in Babel would take more time.

chicoxyzzy:
I have noticed that I mistakenly marked Edge 15 as `"obsolete": true` and Edge 16 as `"obsolete": "very"`. Should be the other way around. Could you please fix this here?

ljharb:
@nicolo-ribaudo i mean, ideally i should be able to put any entry in any order and it shouldn't matter :-)

nicolo-ribaudo:
Would analyzing the engine names to automatically find the previous version be ok, or should they explicitly "link" to the previous one? Currently this is possible by adding `equals` everywhere, which could be renamed to `inherits`.

ljharb:
("inherits" or "extends" is a better word than "equals", yes)

I don't think it's necessary for "chrome39" to explicitly link to "chrome38", but I don't think the order of the engines should matter for it to be able to find "chrome38", or if there's a gap, "chrome17" or whatever.

chicoxyzzy (approved):

chicoxyzzy:
I think we can merge these changes and make all further improvements in next PRs

ljharb (approved):
fair enough

nicolo-ribaudo:
I will prepare another PR :+1:

PR #1578, opened 2020-01-28
Add sticky 'feature name' column

davidsharp:
Makes the feature name column sticky for better readability when scrolling the page horizontally

### DIFF
--- a/master.css
+++ b/master.css
@@ -495,7 +495,7 @@ body.hide-platformtype sup.num-features {
     position: -webkit-sticky;
     position: sticky;
     top: 0;
-    z-index: 1;
+    z-index: 2;
   }
 
   #table-wrapper thead tr:nth-child(2) th {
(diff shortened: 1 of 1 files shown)


ljharb:
This doesn't seem to make the "feature name" header stick; can that be added?

davidsharp:
I wasn't sure whether the feature name should stick or not, but I've added that now :+1:
(Optional features were also getting left behind, so that's fixed too)

ljharb (approved):
Thanks, it still looks a little weird because the gray bar on top of "feature name" is not sticky, but this still seems like an improvement.

chicoxyzzy (approved):
I'd prefer if the whole grey column (with the "Current browser" subcolumn) would be sticky to compare results easily, but I'm happy with current changes also!

Thank you @davidsharp!

ljharb:
@davidsharp in that case, let's make the entire grey column including "current browser" sticky, i'll verify manually, and then i'll merge this. thanks!

davidsharp:
I've added the current browser column on this separate branch: https://github.com/davidsharp/compat-table/tree/davidsharp-patch-1-1

It's not my proudest work, as the nature of the table's structure makes it difficult to target the necessary elements (and due to the compatibility header, the current browser columns spacer element is sticky too), but if you're happy with it then I can add it to this branch.

ljharb:
The spacer actually looks decent to me; it helps differentiate "current" from "all the others". I'm content to update the PR to that.

ljharb:
Thanks!

PR #1585, opened 2020-02-13
Add optional function call to optional chaining

ilogico:
I noticed the syntax for the optional function call wasn't being tested `f?.()`.
I also slightly changed the test for optional method call to better guarantee `this` is correct.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -3874,9 +3874,34 @@ exports.tests = [
       {
         name: 'optional method call',
         exec: function(){/*
-          var foo = { baz: function () { return 42; } };
+          var foo = { baz: function () { return this.value; }, value: 42 };
           var bar = null;
           return foo?.baz() === 42 && bar?.baz() === undefined;
+        */},
+        res : {
+          babel7corejs2: true,
+          typescript3_7corejs3: true,
+          ie11: false,
+          firefox10: false,
+          firefox52: false,
+          firefox73: false,
+          firefox74: true,
+          chrome77: false,
+          chrome78: {val: 'flagged', note_id: "chrome-optional-chaining"},
+          chrome80: true,
+          safari13_1: true,
+          safaritp: true,
+          graalvm: false,
+        }
+      },
+      {
+        name: 'optional function call',
+        exec: function(){/*
+          var foo = { baz: function () { return 42; } };
+          var bar = {};
+          function baz() { return 42; };
+          var n;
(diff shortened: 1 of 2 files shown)


ljharb (approved):
Pending confirmation that all the results here have been manually verified.

ilogico:
I didn't confirm the results manually, I assumed they were the same as the rest.
I did confirm them on the latest Firefox, Chrome and TypeScript.

ljharb:
In that case, please remove all the results you didn't confirm manually :-) future PRs can add the results as they're tested.

ilogico:
I removed the results I couldn't confirmed.
I kept the false results which are guaranteed to be false (ie11 and firefox < 73)

chicoxyzzy (approved):
Both optional method call and optional function call tests return `true` in Safari 13.1 Beta and Firefox 74 Beta. Though we can add these results in the following PR.

data-es2016plus.js
```diff
@@ -3874,23 +3874,38 @@ exports.tests = [
...
         res : {
-          babel7corejs2: true,
           typescript3_7corejs3: true,
           ie11: false,
           firefox10: false,
           firefox52: false,
           firefox73: false,
-          firefox74: true,
           chrome77: false,
-          chrome78: {val: 'flagged', note_id: "chrome-optional-chaining"},
```
chicoxyzzy: I think that this result is still valid, we should re-check it

data-es2016plus.js
```diff
@@ -3874,23 +3874,38 @@ exports.tests = [
       {
         name: 'optional method call',
         exec: function(){/*
-          var foo = { baz: function () { return 42; } };
+          var foo = { baz: function () { return this.value; }, value: 42 };
           var bar = null;
           return foo?.baz() === 42 && bar?.baz() === undefined;
         */},
         res : {
-          babel7corejs2: true,
```
chicoxyzzy: This result is still valid

chicoxyzzy (changes requested):
Thinking again, all removed test results from optional method call tests should be added again. I've rechecked them all manually.

ilogico:
Should I add them again?

chicoxyzzy:
@ilogico yes please

ilogico:
Done

chicoxyzzy:
Oh, sorry, I mentioned only optional method call test results, not optional function call. But that's ok, I'll check them too. Thanks!

chicoxyzzy (approved):
Optional function call results are also correct. Thank you again!

PR #1590, opened 2020-03-02
Add JerryScript 2.2.0 results

ossy-szeged:


asyncTestPassedHead.js
```diff
@@ -0,0 +1,16 @@
+/* jshint esversion: 6, unused: false, undef: false */
```
ljharb: why do we need these extra files littering the root of the project?

ljharb (changes requested):

asyncTestPassedHead.js
ossy-szeged: > why do we need these extra files littering the root of the project?

We don't need to add extra files, I added their content to jerryscript.js.

ljharb (approved):

chicoxyzzy (approved):

PR #1591, opened 2020-03-04
Update GraalVM to version 20.0.0

wirthi:
I have updated the data for GraalVM 20.0.0
Also clarified this is run with the --js.ecmascript-version=2021 flag to get all latest features included.


chicoxyzzy (approved):

ljharb (changes requested):
If it requires a flag, it can’t be `true` - it has to indicate that it requires the flag.

wirthi:
> If it requires a flag, it can’t be `true` - it has to indicate that it requires the flag.

@ljharb thanks, can you please advice what I should do? We don't need a flag for any feature that is "ECMAScript 2019" or below, as this is our default. We need the `--js.ecmascript-version=2020` (or `=2021`) for features expected to land in those spec versions, and `--intl-402` for the `Intl` package.

ljharb:
There’s other examples of things that use a string, instead of `true`, to indicate that it’s only supported via a flag - I’m mobile rn so i can’t link to a specific one, but grep for “flagged”

wirthi:
@ljharb it took me a while to grasp what you meant. Setting an object with `val: 'flagged'` instead of the boolean `true` in the data-files. I had thought you meant the entry in `environment.json`.

I will fix that.

It's unfortunate that this system seemingly penalizes GraalVM to be ECMAScript-compatible by default. We are 2019 compatible, all the features that might (or might not) land in 2020 are flagged now and let us seem less compatible. We might have to tweak our defaults just to look more compatible on this table.

ljharb:
It correctly “penalizes” anything that’s not supported by default, yes. Flags are fine but they’re not the same as being a default.

Everything that’s in master in GitHub will land in 2020 when i cut the branch this month, so there’s not much point in waiting once a feature is merged :-)

environments.json
```diff
@@ -3233,13 +3233,13 @@
     ]
   },
   "graalvm": {
-    "full": "GraalVM JavaScript 19.0.0",
+    "full": "GraalVM JavaScript 20.0.0",
```
ljharb, line 3262: can we keep the results for v19?

wirthi: I've separated the GraalVM 19.3.1 and GraalVM 20.0.0 results. Not much difference at the moment, but we will make GraalVM 20.1.0 by default ES 2020, then the 19 and 20 results will differ more significantly.

ljharb (approved):
Thanks for bearing with me!

wirthi:
Thank you for helping me! @ljharb

PR #1593, opened 2020-03-20
Public static class fields supported on Firefox 75

afmenez:
Public static class fields will be supported on Firefox 75:
https://wiki.developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/75#JavaScript


data-esnext.js
```diff
@@ -986,7 +987,8 @@ exports.tests = [
       */},
       res: {
         firefox2: false,
-        firefox75: firefox.developer,
+        firefox74:false,
```
ljharb:
```suggestion
        firefox74: false,
```

data-esnext.js
```diff
@@ -946,7 +946,8 @@ exports.tests = [
         tr: true,
         typescript1corejs2: true,
         firefox2: false,
-        firefox75: firefox.developer,
+        firefox74:false,
```
ljharb:
```suggestion
        firefox74: false,
```

ljharb (changes requested):
Seems like this shouldn’t merge until April 7th?

afmenez:
Why the wait?

ljharb:
In case it doesn’t end up being released on that date. We generally don’t include things in advance of their being released, i think?

afmenez:
Yes, we do.  We have the "unstable" tab for that.  There are 2 unstable
versions of Firefox.  And Chrome.  And Safari.

ljharb (approved):

ljharb (approved):

chicoxyzzy (approved):

PR #1595, opened 2020-03-24
Add Safari 13.1 release

chicoxyzzy:
It was released along with iOS 13.4 and macOS 10.15.4 today

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -2401,7 +2401,7 @@
     "family": "JavaScriptCore",
     "short": "SF 11",
     "release": "2017-09-19",
-    "obsolete": true
+    "obsolete": "very"
   },
   "safari11_1": {
     "full": "Safari 11.1",
@@ -2422,7 +2422,7 @@
     "family": "JavaScriptCore",
     "short": "SF&nbsp;12.1",
     "release": "2019-03-25",
-    "obsolete": false
+    "obsolete": true
   },
   "safari13": {
     "full": "Safari 13",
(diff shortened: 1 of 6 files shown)


environments.json
```diff
@@ -2422,7 +2422,7 @@
     "family": "JavaScriptCore",
     "short": "SF&nbsp;12.1",
     "release": "2019-03-25",
-    "obsolete": false
+    "obsolete": true
```
ljharb, line 2425: i'd think we usually consider 2 majors not to be obsolete; meaning, 12.1 isn't obsolete until 14 is out.

chicoxyzzy: Two not obsolete versions in this PR are 13 and 13.1

ljharb: sure but those are minors, not majors.

chicoxyzzy: Browserslist also shows minor Safari updates in https://browserl.ist/?q=last+2+versions. As far as I know, Safari minor versions are actually could be treated as a majors, because they just include all latest Safari TP changes. Major version number is coupled to macOS major version.

ljharb: Right - which means that if we only show 13.1 and 13.0 as not obsolete, then people on a slightly older OS won’t see any relevant results.

chicoxyzzy: usually new versions are available on all of versions of macOS which are supported at the moment of release (2 latest versions of macOS)

chicoxyzzy: https://developer.apple.com/documentation/safari_release_notes/safari_13_1_beta_release_notes

chicoxyzzy: 13.1 is available even on three last OS versions

> Safari 13.1 ships with iOS 13.4 and macOS 10.15.4. It’s also available for macOS 10.14.6 and 10.13.6.

ljharb: Fair enough.

ljharb (approved):

PR #1597, opened 2020-04-07
Firefox 75 release

chicoxyzzy:
I've found no changes


ljharb (approved):
Is there anything that should be adjusted here based on #1598?

chicoxyzzy:
Do we consider version as very obsolete when its release date is over a year or when its retire date was more than a year ago? I don't remember.

ljharb:
I'd say retire date, to be safer.

chicoxyzzy:
fixed :)

PR #1600, opened 2020-04-15
Do not rely on envs ordering when determining browsers inheritance

nicolo-ribaudo:
Ref: https://github.com/kangax/compat-table/pull/1576#issuecomment-577084923

This will prevent regressions in case the order of the table columns changes.

There are some changes in the generated HTML: I don't understand why, but they all seem correct to me according to the data we have.

### DIFF
--- a/esnext/index.html
+++ b/esnext/index.html
@@ -2025,7 +2025,7 @@ Promise.any([
 <td class="tally" data-browser="firefox75" data-tally="1">2/2</td>
 <td class="tally unstable" data-browser="firefox76" data-tally="1">2/2</td>
 <td class="tally unstable" data-browser="firefox77" data-tally="1">2/2</td>
-<td class="tally obsolete" data-browser="opera12_10" data-tally="0">0/2</td>
+<td class="tally obsolete" data-browser="opera12_10" data-tally="1">2/2</td>
 <td class="tally obsolete" data-browser="chrome74" data-tally="1">2/2</td>
 <td class="tally obsolete" data-browser="chrome75" data-tally="1">2/2</td>
 <td class="tally obsolete" data-browser="chrome76" data-tally="1">2/2</td>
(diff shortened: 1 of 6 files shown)


esnext/index.html
```diff
@@ -2156,9 +2156,9 @@
 <td class="yes" data-browser="chrome81">Yes</td>
 <td class="yes unstable" data-browser="chrome83">Yes</td>
 <td class="yes unstable" data-browser="chrome84">Yes</td>
-<td class="no obsolete" data-browser="edge15">No</td>
-<td class="no obsolete" data-browser="edge17">No</td>
-<td class="no" data-browser="edge18">No</td>
+<td class="yes obsolete" data-browser="edge15">Yes</td>
+<td class="yes obsolete" data-browser="edge17">Yes</td>
+<td class="yes" data-browser="edge18">Yes</td>
```
nicolo-ribaudo, line 2161: For example, this is supported by IE7 and we don't have any other data that says that newer versions of Chakra don't support it.

scripts/build-environments-tree.js
```diff
@@ -0,0 +1,86 @@
...
+  var result = {};
+  // Pre-initialize the "result" object to make it have the same order as "data"
+  Object.keys(data).forEach(function (id) { result[id] = null; });
+
+  var envs = Object.create(null);
+
+  Object.keys(data).forEach(function (id) {
+    var parsed = parseEnvId(id);
+    if (!envs[parsed.name]) envs[parsed.name] = [];
+    envs[parsed.name].push(parsed);
```
chicoxyzzy, line 16: it seems that `result[id] = null;` from line 12 could be moved here to avoid second iteration over `data` object keys

chicoxyzzy (approved):
I think that changes in HTML are correct. Probably there was a bug that omitted nested results for browsers with `obsolete: "very"`

scripts/build-environments-tree.js
nicolo-ribaudo: Actually, I think I can just remove it since the code is not relying on the ordering of `result` (it did in a previous attempt).

ljharb (approved):
LGTM

esnext/index.html
afmenez: Isn't this issue #1118?

ljharb: Hm, possibly? If so, i guess we can close it :-)

PR #1601, opened 2020-04-15
Fix opera 30 & 33 info

nicolo-ribaudo:
I found these errors while trying to unify Babel's tests results analysis logic with `compat-data`'s.

It didn't cause any change in the generated HTML.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -2572,8 +2572,8 @@
     "short": "OP 29",
     "obsolete": "very"
   },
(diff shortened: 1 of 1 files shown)


ljharb (approved):
Rebased; the build had no changes

afmenez (approved):
My bad, should have double-checked!

chicoxyzzy (approved):
oh wow...

PR #1602, opened 2020-04-16
add private-class-methods test results

### LINKED ISSUE #1496: Tests missing for static and instance methods
The esnext tab has some tests for instance class fields and static class fields, but the proposals also contains methods which currently have no tests:
- [Public static methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_elements#Public_static_methods)
- [Private static methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_elements#Private_static_methods)
- [Public instance methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_elements#Public_instance_methods)
- [Private instance methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_elements#Private_instance_methods)

JLHwung:
Resolves #1496 

Note that `public static methods` and `public instance methods` are included in `data-es6.js`: 

https://github.com/kangax/compat-table/blob/737e0ae1d24bbb33b5f01ee6fc4072389d4f3ba7/data-es6.js#L16659

and 
https://github.com/kangax/compat-table/blob/737e0ae1d24bbb33b5f01ee6fc4072389d4f3ba7/data-es6.js#L16633

respectively.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1001,6 +1001,108 @@ exports.tests = [
   ]
 },
 {
+  name: 'private class methods',
+  category: STAGE3,
+  significance: 'medium',
+  spec: 'https://github.com/tc39/proposal-private-methods',
+  subtests: [
+    {
+      name: 'private instance methods',
+      exec: function () {/*
+        class C {
+          #x() { return 42; }
+          x() {
+            return this.#x();
+          }
+        }
+        return new C().x() === 42;
+      */},
+      res: {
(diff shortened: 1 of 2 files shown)


ljharb (approved):
rebased

chicoxyzzy (approved):

PR #1603, opened 2020-04-16
Implement the Logical Assignment tests

afmenez:
Implementing the Logical Assignment tests.  
- I believe I tested all important cases.
- Added the results for Chrome and Firefox
- I am never sure about the significance field.  Is 'small' OK for this one?

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -3990,7 +3990,7 @@ exports.tests = [
   spec: 'https://github.com/tc39/proposal-logical-assignment/',
   subtests: [
     {
-      name: '||=',
+      name: '||= basic support',
       exec: function () {/*
         let a;
         let b = 0;
@@ -4001,6 +4001,7 @@ exports.tests = [
         return a === 2 && b === 2 && c === 1;
       */},
       res: {
+        firefox60: false,
         firefox76: false,
         firefox77: firefox.nightly,
         chrome80: false,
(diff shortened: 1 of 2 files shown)


data-esnext.js
```diff
@@ -3881,6 +3881,65 @@ exports.tests = [
...
+        let c = 1;
+        a ||= 2;
+        b ||= 2;
+        c ||= 2;
+        return (a == 2) && (b == 2) && (c == 1);
+      */},
+      res: {
+        firefox76: false,
+        firefox77: firefox.nightly,
+        chrome80: false,
```
JLHwung, line 4007: Can confirm both `safari13` and `safaritp` does not support it either.

data-esnext.js
```diff
@@ -3881,6 +3881,65 @@ exports.tests = [
...
+    {
+      name: '||=',
+      exec: function () {/*
+        let a;
+        let b = 0;
+        let c = 1;
+        a ||= 2;
+        b ||= 2;
+        c ||= 2;
+        return (a == 2) && (b == 2) && (c == 1);
```
JLHwung: nit: use `===` is preferred.

ljharb (approved):
"small" seems right to me. rebased and fixed the nits.

JLHwung (approved):

nicolo-ribaudo (commented):
Depending on how much the tests should be specific, we could also test two differences these operators have from all the other assignment operators:
1. The short-circuiting behaviour (`true ||= ++i`) shouldn't increment `i`
2. Setters shouldn't be unnecessarily invoked:
    ```js
    var obj = { get x() { return true }, set x() { i++; } }
    obj.x ||= 2; // shouldn't increment i
    ```

nicolo-ribaudo (approved):

data-esnext.js
afmenez: added safari13 results, safaritp is inherited.

data-esnext.js
afmenez: Done by @ljharb

chicoxyzzy (approved):

PR #1605, opened 2020-04-17
Extract data interoplation logic to a separate file

nicolo-ribaudo:
This PR only moves code around, without adding any feature or fixing any bug.

I extracted all the logic used to interoplate envs/tests data to a separate file, so that consumers of `compat-table` (i.e. Babel) can `require()` that file without needing to re-implement all the logic.

You can see the advantage of this in Babel's `compat-data` build script: it allowed me to delete more than 50% of the code, while also finding some bugs in the old generated data: [diff](https://github.com/babel/babel/compare/master...nicolo-ribaudo:update-compat-table-script#diff-b0170cca6f037004c76ccb91b5798200)

Note that since `compat-table` is not actually meant to be consumed by other programs, this new file shouldn't introduce any backward-compatibility concern, and it can be refactored whenever needed.

Also, I renamed `scripts/build-environments-tree.js` (introduced by #1600) to `build-utils.js`, since the `scripts` folder was inconsistent with all the other files (that are in the top-level folder).

### DIFF
--- a/build-utils/compare-versions.js
+++ b/build-utils/compare-versions.js
@@ -1,0 +1,21 @@
+module.exports = function compareVersions(a, b) {
+  var vA = a.version;
+  var vB = b.version;
+
+  if (!vA && !vB) return 0;
+  if (!vB) return 1;
+  if (!vA) return -1;
+
+  for (var i = 0; i < vA.length && i < vB.length; i++) {
+    if (vA[i] === "tp" && vB[i] === "tp") continue;
+    if (vA[i] === "tp") return 1;
+    if (vB[i] === "tp") return -1;
+    if (vA[i] > vB[i]) return 1;
+    if (vA[i] < vB[i]) return -1;
+  }
+
+  if (vA.length > vB.length) return 1;
+  if (vA.length < vB.length) return -1;
+
+  return 0;
+};
--- a/build-utils/interpolate-all-results.js
+++ b/build-utils/interpolate-all-results.js
@@ -1,0 +1,45 @@
+var parseEnvsVersions = require("./parse-envs-versions");
+
+module.exports = function interpolateAllResults(tests, envs) {
+  var envsTree = buildEnvsTree(envs);
+
+  tests.forEach(function (t) {
+    // Calculate the result totals for tests which consist solely of subtests.
+    [].concat(t.subtests || t).forEach(function (e) {
+      interpolateTestResults(e.res, envsTree);
+    });
+  });
+};
+
+function interpolateTestResults(res, envsTree) {
+  var bid, prevBid;
+  for (bid in envsTree) {
+    if (res[bid] !== undefined) continue;
+
+    prevBid = bid;
+    do {
+      prevBid = envsTree[prevBid];
+    } while (prevBid !== null && res[prevBid] === undefined);
+
+    if (prevBid !== null) res[bid] = res[prevBid];
+  }
+}
+
+// This function returns an object mapping "browser id" -> "parent browser id"
+function buildEnvsTree(data) {
+  var result = {};
+  var envs = parseEnvsVersions(data);
+
+  Object.keys(envs).forEach(function (name) {
+    var id = envs[name][0].id;
+    result[id] = data[id].equals || null;
+
+    for (var i = 1; i < envs[name].length; i++) {
+      id = envs[name][i].id;
+      if (data[id].equals === false) result[id] = null;
+      else result[id] = data[id].equals || envs[name][i - 1].id;
+    }
+  });
+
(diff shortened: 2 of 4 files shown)


build-utils.js
```diff
@@ -3,22 +3,16 @@ var RE = {
   version: /^(\d{4})(\d{2})(\d{2})$|^(\d+)(?:_(\d+))?(?:_(\d+))?$/
 };
 
-module.exports = buildEnvsTree;
+exports.parseEnvsVersions = parseEnvsVersions;
```
ljharb: tbh i'd much rather move these functions to separate files than to pretend that CJS has more than a default export.

nicolo-ribaudo: Back to the `scripts` folder then?

ljharb: i'm ok with that

chicoxyzzy (approved):

build-utils/compare-versions.js
```diff
@@ -0,0 +1,21 @@
+module.exports = function compareVersions(a, b) {
```
ljharb, line 1: i wonder if this would be simpler if we used one of `semver`'s comparison functions (altho it wouldn't handle `tp`)

build-utils/interpolate-all-results.js
```diff
@@ -0,0 +1,49 @@
+var parseEnvsVersions = require("./parse-envs-versions");
+
+module.exports = function interoplateAllResults(tests, envs) {
+  var envsTree = buildEnvsTree(envs);
+
+  tests.forEach(function (t) {
+    // Calculate the result totals for tests which consist solely of subtests.
+    if ("subtests" in t) {
```
ljharb: should this use `Object.prototype.hasOwnProperty.call(t, 'subtests')` rather than `in`?

build-utils/interpolate-all-results.js
```diff
@@ -0,0 +1,49 @@
...
+      t.subtests.forEach(function(e) {
```
ljharb:
```suggestion
      t.subtests.forEach(function (e) {
```

build-utils/interpolate-all-results.js
```diff
@@ -0,0 +1,49 @@
...
+        interpolateTestResults(e.res, envsTree);
+      });
+    } else {
+      interpolateTestResults(t.res, envsTree);
+    }
+  });
```
ljharb: this might be a bit simpler, and won't repeat the interpolate call?
```suggestion
    [].concat(t.subtests || t.res).forEach(function(e) {
      interpolateTestResults(e.res, envsTree);
    });
  });
```

build.js
```diff
@@ -24,7 +24,7 @@ require('object.assign').shim();
 var pickBy = require('lodash.pickby');
 
 var environments = require('./environments');
-var buildEnvironmentsTree = require('./scripts/build-environments-tree.js');
+var interoplateAllResults = require('./build-utils/interpolate-all-results');
```
ljharb:
```suggestion
var interpolateAllResults = require('./build-utils/interpolate-all-results');
```

build.js
```diff
@@ -501,15 +485,10 @@ function dataToHtml(skeleton, rawBrowsers, tests, compiler) {
     );
   });
 
+  interoplateAllResults(tests, environments);
```
ljharb:
```suggestion
  interpolateAllResults(tests, environments);
```

ljharb (commented):
None of these are blockers, but maybe worth fixing before merging this?

build-utils/compare-versions.js
nicolo-ribaudo: I would prefer to keep these files dependencies-free if possible, to make it easier to re-use them by simply cloning this repository.

build-utils/interpolate-all-results.js
nicolo-ribaudo: Yes!
I used `in` because it's used in other parts of the codebase, but I'm happy to remove it.

build.js
nicolo-ribaudo: :see_no_evil:

build-utils/compare-versions.js
ljharb: I'm fine with that in this case, but I think that's flat out a bad goal to have in general. If we want this stuff to be usable by outside tools we should publish it as a separate package :-/

ljharb (approved):

build-utils/compare-versions.js
nicolo-ribaudo: Yeah, "zero deps" shouldn't be a goal. However, in this case it was easy enough to keep it with the custom implementation.

PR #1606, opened 2020-04-17
Update babel support data

JLHwung:
This PR should be reviewed by commits.

The first commit updates the test page to Babel 7 so I can check the results on my local browsers.

The second commit updates the test results of babel 7 according to the test page results and my assessment on some `eval` test cases. (See explainers below).

Update rule:
If the transpiled code passes the test without any `corejs` helpers, set `babel7corejs2: true`, otherwise set `babel7corejs3: babel.corejs` since `corejs2` is obsolete and the value of support is marginal.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -578,6 +578,7 @@ exports.tests = [
         res: {
           tr: null,
           babel6corejs2: null,
+          babel7corejs2: babel.regenerator,
           closure: true,
           typescript1corejs2: typescript.downlevelIteration,
           chrome52: null,
@@ -707,6 +708,7 @@ exports.tests = [
         res: {
           tr: null,
           babel6corejs2: null,
+          babel7corejs2: babel.regenerator,
           closure: true,
           typescript1corejs2: typescript.downlevelIteration,
           chrome52: null,
@@ -768,6 +770,7 @@ exports.tests = [
         res: {
           tr: null,
           babel6corejs2: null,
+          babel7corejs2: babel.regenerator,
           closure: true,
           typescript1corejs2: typescript.downlevelIteration,
           chrome52: null,
@@ -835,6 +838,7 @@ exports.tests = [
         res: {
           tr: null,
           babel6corejs2: null,
+          babel7corejs2: babel.regenerator,
           closure: true,
           typescript1corejs2: typescript.downlevelIteration,
           chrome52: null,
@@ -873,6 +877,7 @@ exports.tests = [
         res: {
           tr: null,
           babel6corejs2: null,
+          babel7corejs2: babel.regenerator,
           closure: true,
           typescript1corejs2: typescript.downlevelIteration,
           chrome52: null,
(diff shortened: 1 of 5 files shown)


data-es2016plus.js
```diff
@@ -3287,6 +3294,7 @@ exports.tests = [
         */},
         res : {
           closure20190215: true,
+          babel7corejs2: true,
```
JLHwung, line 3707: The JSON superset is supported by [@babel/proposal-json-strings](https://babeljs.io/docs/en/babel-plugin-proposal-json-strings) and included in `preset-env`, although by default the `eval`-ed strings will not be transpiled.

Note that we could transpile them with the help of https://github.com/babel/babel-test262-runner/blob/master/lib/run-tests/plugins/transform-eval.js but I feel like it's a bit overkill here especially the compiler test page is only one of reference.

nicolo-ribaudo:
Should we have a new `babel7_9corejs3` entry? Many of those results don't reflect all the Babel 7 versions.

ljharb:
I think for things like transpilers and polyfills, we can probably just assume that any major version "n" is the latest available in that version? Meaning, all the babel7 results should be with latest babel.

(Unless there's value in keeping them separate, for babel-preset-env, or something?)

nicolo-ribaudo:
I proposed to specify the minor version because:
- I don't think that it adds too much complexity/maintenance, but it increases the granularity (and thus quality) of the data;
- Many people don't use older minors, so having data for them might be useful;
- For TypeScript, we already have data for every minor release. Also, we have many versions of the Closure Compiler (but since they are date-based I don't know if they would constitute a precedent).

> Unless there's value in keeping them separate, for babel-preset-env, or something?

Not that I know of, `@babel/preset-env` only needs the browsers data.

JLHwung:
I agree with @ljharb on that it suffices to have major version data. And maybe we should apply the same rule for closure compiler as well as typescript.

As a dev tool, any team can and is recommended, to update to the latest minors, unlike node.js versions/browser versions which they may not control. The `compat-table` should serve as a reference to showcase the gap between your code and the engines, and to which extent the polyfills/tranpilers can help fill in the gap. That said if you see something does not work as stated, the first thing you should check is whether you have updated to the latest versions.

Another ergonomic point: when there are many versions arranged in a big table (clicked the "show obsolete platforms" button), the scrolling behaviour is really wonky.

chicoxyzzy (approved):

ljharb:
TS doesn't use semver, so we *must* keep minors data for TS.

nicolo-ribaudo:
Ok, those are all fair points 👍

JLHwung:
@ljharb Rebased. Please let me know if there is any other questions.

nicolo-ribaudo:
@JLHwung could you rebase this and @ljharb could you review? :pray:

JLHwung:
@ljharb @chicoxyzzy Rebased.

nicolo-ribaudo (approved):

ljharb:
Does this effect the ability of the es5 page to run in older browsers, or is this only at deploy time?

JLHwung:
I think the changes in `build.js` are targeted to how we test if Babel supports certain features. It seems not related to how the website is built.

ljharb (approved):

chicoxyzzy (approved):

PR #1608, opened 2020-04-22
fix: render edge79+ as V8

JLHwung:
The Edge79+ should be rendered as V8. Here we whitelist edge12-18 as Chakra and the rest as V8.

### DIFF
--- a/master.js
+++ b/master.js
@@ -326,27 +326,27 @@ $(function() {
   // browser engine color stripes
   function getBrowserColour(name) {
     /* Chakra */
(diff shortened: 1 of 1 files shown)


ljharb (approved):

chicoxyzzy (approved):

afmenez:
Shouldn't Opera > 12 also be rendered as V8 then?

ljharb:
@afmenez i believe Opera 15 was the first Blink-based version.

PR #1609, opened 2020-04-22
feat: add hashbang grammar support data

JLHwung:


data-esnext.js
```diff
@@ -4144,6 +4144,34 @@ exports.tests = [
...
+  */},
+  res: {
+    chrome1: false,
+    chrome74: true,
+    firefox2: false,
+    firefox67: true,
+    ie11: false,
+    opera10_50: false,
+    edge18: false,
+    safari13: null,
```
JLHwung: I don't have `safari13` environments in hands, so I will be appreciated if you can suggest `safari13` support data. The other major browser support data is copied from https://github.com/tc39/proposal-hashbang#status

data-esnext.js
```diff
@@ -4144,6 +4144,34 @@ exports.tests = [
     },
   ]
 },
+{
+  name: 'Hashbang Grammar',
+  category: STAGE3,
+  significance: 'tiny',
```
JLHwung, line 4150: Mark it as `tiny` because it is a simple grammar extension.

data-esnext.js
existentialism: Locally on Safari@13.0.5, `safari13` returns `false`

data-esnext.js
```diff
@@ -4144,6 +4144,36 @@ exports.tests = [
...
+    }
+  */},
+  res: {
+    chrome1: false,
+    chrome74: true,
+    firefox2: false,
+    firefox67: true,
+    ie11: false,
+    opera10_50: false,
+    edge18: false,
```
chicoxyzzy, line 4167: Could you please add `edge79: true`?

chicoxyzzy (approved):

data-esnext.js
JLHwung: I think `edge78+` maps to `chrome 78+` if support data is missing. So `chrome74: true` should implies `edge78: true`. If it turns out that it doesn't work on `edge 78`, then we should mark as unsupported, otherwise I am good to leave it as-is.

chicoxyzzy: oh yes indeed :)

ljharb (approved):

PR #1610, opened 2020-04-24
Test optional catch on generator throw

jridgewell:
It seems the current test is actually just testing that a normal throw will be caught by an optional-binding catch. I think it should be testing that resuming the generator with a throw will be properly caught, like how the `await` test is using a rejected promise's throw.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -2994,7 +2994,6 @@ exports.tests = [
           function *foo() {
             try {
               yield;
-              throw new Error();
             }
             catch {
               return true;
@@ -3003,7 +3002,7 @@ exports.tests = [
 
           var it = foo();
           it.next();
-          return it.next().value;
+          return it.throw().value;
         */},
         res: {
           babel7corejs2: true,
@@ -3013,15 +3012,12 @@ exports.tests = [
           firefox2: false,
           firefox57: false,
           firefox58: true,
-          opera10_50: false,
           chrome65: chrome.harmony,
           chrome66: true,
           safari11_1: true,
           safaritp: true,
           webkit: true,
           duktape2_2: false,
-          graalvm19: true,
-          graalvm20: true,
         }
       }
     ]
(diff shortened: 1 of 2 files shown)


ljharb:
Seems like if we're updating a test, we'd need to invalidate and recheck all results of that test.

jridgewell:
How do I do that?

ljharb:
The simplest is just to delete all the results for that test; the best but most work is to manually check the results of your updated test in each of the engines, and update the result accordingly.

jridgewell:
I removed everything I wasn't able to verify.

ljharb (approved):

ljharb:
@chicoxyzzy once you've stamped this, i'll rebase and merge it :-)

chicoxyzzy (commented):
Test returns true in Safari 13.1. Also we have a [test runner for Ducktape](https://github.com/kangax/compat-table/blob/gh-pages/duktape.js).

existentialism:
Test returns true for Safari 13.0.5 as well!

jridgewell:
Tested in safari11_1 and duktape2_2, both remained the same.

data-es2016plus.js
```diff
@@ -3013,15 +3012,12 @@ exports.tests = [
           firefox2: false,
           firefox57: false,
           firefox58: true,
-          opera10_50: false,
           chrome65: chrome.harmony,
           chrome66: true,
           safari11_1: true,
           safaritp: true,
           webkit: true,
           duktape2_2: false,
```
chicoxyzzy, line 3020: This should be `true` according to @jridgewell's [comment](https://github.com/kangax/compat-table/pull/1610#issuecomment-619465950)

jridgewell: Sorry, I misspoke. I confirmed that this failed, as the original value did.

chicoxyzzy (approved):

PR #1613, opened 2020-04-30
Add optional deep chain with private access

jridgewell:
This was also recently approved in TC39. I collected the results by taking the maximum of the optional chain and private instance fields results, then manually verifying the supported browsers. Only Chrome has it so far.

Re: https://github.com/kangax/compat-table/pull/1612

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -930,6 +930,32 @@ exports.tests = [
       }
     },
     {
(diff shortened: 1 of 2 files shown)


ljharb (approved):

ljharb:
(you'll need to `npm run build` to get the tests passing)

chicoxyzzy (approved):

PR #1622, opened 2020-05-07
Add Object Iteration results for Firefox and Chrome

afmenez:
Ditto.


ljharb (approved):
These are technically correct, but it's stage 1 - it should be false for basically everything, since it doesn't even have a final solution shape yet.

I'm very much growing my opinion that we should just remove stage 0 and 1 proposals from the table entirely.

afmenez:
Rebased after Stage 0 removal.  I believe it is now ready to be merged.

ljharb (approved):

chicoxyzzy (approved):

PR #1627, opened 2020-05-18
Split Unicode Escape tests

jridgewell:
Babel is [concerned](https://github.com/babel/babel/pull/11377) with a few unique cases:

1. Escapes in strings
2. Escapes in bare identifiers
3. Escapes in the property keys

These require a slightly different transform for each, and only strings and property keys are actually transformable. If we encounter an escape in a bare identifier, we currently error out.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -17723,8 +17723,8 @@ exports.tests = [
     {
       name: 'in identifiers',
       exec: function(){/*
-        var \u{102C0} = { \u{102C0} : 2 };
-        return \u{102C0}['\ud800\udec0'] === 2;
+        var \u{102C0} = 2;
+        return \u{102C0} === 2;
       */},
       res: {
         ejs: true,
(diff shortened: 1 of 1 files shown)


data-es6.js
```diff
@@ -17273,6 +17273,45 @@ exports.tests = [
...
       }
     },
+    {
+      name: 'in property key definitions',
+      exec: function(){/*
+        var o = { \u{102C0} : 2 };
+        return o['\ud800\udec0'] === 2;
+      */},
+      res: {
+        ejs: true,
```
jridgewell, line 17754: I reused the same test results, since this was split from the previous test. I imagine everything will pass them both individually.

data-es6.js
```diff
@@ -17273,6 +17273,45 @@ exports.tests = [
...
+      }
+    },
+    {
+      name: 'in property key accesses',
+      exec: function(){/*
+        var o = { '\ud800\udec0' : 2 };
+        return o.\u{102C0} === 2;
+      */},
+      res: {
+        edge12: true,
```
jridgewell, line 17777: These are the only implementations I could manually verify.

ljharb (approved):
As long as results are all manually verified, this LGTM

chicoxyzzy (approved):

PR #1631, opened 2020-05-20
Remove stage -1 and stage 0 tests, make stage 1 tests non-significant

chicoxyzzy:
- Stage -1 and stage 0 shouldn't be presented in the compatibility table as they are not proposals TC39 is working on.
- Stage 1 proposals shouldn't be taken into account when we count score since they don't have spec text, so they can't be tested


ljharb (approved):
This is great. We should even look into removing stage 1 (in a separate PR), as nobody should be implementing those yet either :-)

PR #1634, opened 2020-05-27
Update to GraalVM 20.1.0

wirthi:
Updating the values from GraalVM 20.0.0 to 20.1.0.


data-esnext.js
```diff
@@ -1585,7 +1585,7 @@ exports.tests = [
     safari13: true,
     safaritp: true,
     graalvm19: false,
-    graalvm20: {val: 'flagged', note_id: "graalvm-es2020", note_html: "The feature have to be enabled via <code>--js.ecmascript-version=2020</code> flag"},
```
ljharb, line 1588: it seems better to me to keep this as-is, and add a new value for graalvm 20.1.

chicoxyzzy (commented):
I think that it would be better to keep graalvm20 and add graalvm20_1

wirthi:
Why would you keep the old version, what's the advantage? From our point of view, 20.0 is superseded, no-one should be using it any longer. At the end of the year, we would have 19.3.x, 20.0, 20.1, 20.2, and 20.3 - and add four more next year. When do we delete them again?

IMHO, there should be our 19.3 LTS version, and there should be the most recent version of 20 (20.1 currently).

ljharb:
@wirthi people use old versions of everything, what they "should" be using isn't the point. You could certainly mark 20.0 as "very obsolete", but that doesn't mean the test data for it should be erased.

In other words, I don't think test results for any release should be deleted *ever* - 100 years from now, I should be able to determine exactly how much v20.0 complied with, or violated, the spec.

wirthi:
Ok, I am not convinced this is necessary/reasonable, but if that's how you do things here, I'll set it up like that.

Is it possible to NOT duplicate all the values? I see there is an `equals` entry in `environments.json` - can I use that for this case, i.e. have the data in `graalvm20` and only specify exceptions for `graalvm20_1`?

-- Christian

ljharb:
@wirthi yep, exactly! you'd set up each new graal version as basically inheriting from the previous one, and then you'd only have to add the results that differ.

chicoxyzzy (approved):

wirthi:
I have rebased to the current state of gh-pages. Ready to be merged from my side.

chicoxyzzy:
LGTM, thanks!

PR #1639, opened 2020-06-02
Firefox 77

chicoxyzzy:
77 is ~~not released yet, but Beta, Dev Edition and Nightly are bumped.~~ released now

SAB doesn't work in 78+ even with `shared_memory` flag enabled


ljharb (approved):

chicoxyzzy:
small note: SAB was intentionally disabled in SM because 78 will be an ESR

PR #1667, opened 2020-08-17
Add `String.prototype.split` and `.item()`

ljharb:


ljharb:
updated to include a bunch of manual results

chicoxyzzy (approved):

PR #1668, opened 2020-08-24
TypeScript 4 released

afmenez:
Ditto.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -519,7 +519,6 @@
     "family": "TypeScript",
     "platformtype": "compiler",
     "obsolete": "very",
-    "obsolete": true,
     "test_suites": [
       "es6",
       "es2016plus",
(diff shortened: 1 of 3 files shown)


environments.json
```diff
@@ -584,6 +583,19 @@
     "family": "TypeScript",
     "platformtype": "compiler",
     "release": "2020-05-12",
+    "obsolete": true,
```
ljharb, line 586: i'm not sure we should declare 3.9 obsolete yet; i think we should have both the latest v3, and v4, always shown at the same time.

ljharb (commented):
LGTM pending comment

environments.json
nicolo-ribaudo: TypeScript doesn't follow semver, and the new version is 4.0 just because it's what comes after 3.9 using decimal numbers.

ljharb: Yes, I am aware of that, but people will still be on v3 for a long time (there's been plenty of people on v2 for a long time after v3 came out)

chicoxyzzy: Right now we have only latest versions of compilers and polyfills as non-obsolete versions, so I think we can mark 3.9 as obsolete here

chicoxyzzy (approved):

PR #1677, opened 2020-09-21
Firefox 81 released

afmenez:
Firefox 81 will be released tomorrow, but it is already available on the ftp servers.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -1333,7 +1333,7 @@
     "short": "FF 68 ESR",
     "release": "2019-07-09",
     "retire": "2020-09-22",
-    "obsolete": false
+    "obsolete": true
   },
   "firefox69": {
     "full": "Firefox 69",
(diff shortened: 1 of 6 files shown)


environments.json
```diff
@@ -1269,7 +1269,7 @@
     "short": "FF 60 ESR",
     "release": "2018-05-09",
     "retire": "2019-10-22",
-    "obsolete": "very"
+    "obsolete": true
```
chicoxyzzy: why?

afmenez: The "rule" I was using was "1 year after retire date".  This will be true until the next release.

ljharb (approved):
Assuming the release went out as scheduled.

environments.json
chicoxyzzy: I could be wrong, but the rule was about release dates. Anyway, I don't think we should add 60 ESR to the table

afmenez: FF60 removed.

chicoxyzzy (approved):

PR #1679, opened 2020-09-30
CSS: Not let the table overflow the whole page

Jack-Works:


ljharb:
Thanks, can you provide some screenshots that show the difference?

Jack-Works:
Before (the whole page overflowed)



After (the page itself no longer overflowed. The table does.)

ljharb:
This doesn't seem like an improvement - it seems better to me when the only scrollbars are on the window.

PR #1680, opened 2020-10-07
.item() method promoted to Stage 3, renamed to .at(), implemented by Firefox 85

afmenez:
Ditto.


ljharb (changes requested):
We should hold off on landing this due to https://github.com/tc39/proposal-item-method/issues/31 and https://github.com/tc39/proposal-item-method/issues/28

afmenez:
Is the proposal moving back to Stage 2?

ljharb:
It’s unclear; we’ll surely discuss it at the November meeting.

afmenez:
@ljharb, is it OK now?

ljharb (approved):

zloirock (changes requested):
https://github.com/zloirock/core-js#relative-indexing-method

ljharb:
I don’t understand the reason for the block, can you elaborate?

zloirock:
`core-js` implementation was not added. Any reason to do it in another PR?

ljharb:
ah, thanks. while i don't think we should be blocking proposal changes based on adding test results (as long as the result won't be false positives in the interim), it seems fine to add core-js here.

afmenez:
I am not familiar with corejs, and I don't know which versions are and are not compatible.  Do you mind adding it in a separate patch?

chicoxyzzy (approved):

zloirock:
@afmenez just add
```js
babel7corejs3: babel.corejs,
typescript4corejs3: typescript.corejs,
```
to `Array#at` and `%TypedArray%#at` results.

zloirock:
Oops

data-esnext.js
```diff
@@ -1680,17 +1688,20 @@ exports.tests = [
              return true;
            }
            var arr = new Constructor([1, 2, 3]);
```
zloirock, line 1690: `new BigInt64Array([1, 2, 3]) -> TypeError`

data-esnext.js
```diff
@@ -1680,17 +1688,20 @@ exports.tests = [
              return true;
```
chicoxyzzy: I think it should be `false` here

chicoxyzzy (changes requested):
there are a few issues in %TypedArray% test which need to be addressed

data-esnext.js
afmenez: Fixed

ljharb: Let's fix this by adding a separate BigInt test rather than just not testing them at all?

zloirock (approved):

data-esnext.js
afmenez: Fixed.

data-esnext.js
afmenez: Please feel free to add it in a separate patch.  This one is already 10 times the work I originally intended to do.

chicoxyzzy (approved):
Removing BigInt typed arrays tests is a downgrade, but given that old test wasn't working before, I think that we can add them later (but we still need to, because if I'm not mistaken, there is a WIP patch for JavaScriptCore and JSC doesn't support BigInt typed arrays yet)

PR #1694, opened 2021-01-20
Add GraalVM 21.0.0 results

wirthi:
GraalVM 21.0.0. has been released yesterday (January 19th), those are the results for this release.

Also updated the version numbers for the 19 and 20 branches.


data-esnext.js
```diff
@@ -1655,6 +1656,7 @@ exports.tests = [
         safaritp: true,
         babel7corejs3: babel.corejs,
         typescript4corejs3: typescript.corejs,
+        graalvm21: graalvm.es2022flag,
```
ljharb, line 1659: just a comment - being stage 3 now does not guarantee it will be stage 4 in ES2022 - what happens in a year if it hadn’t advanced yet?

ljharb (approved):

data-esnext.js
wirthi: Thanks @ljharb, that's an excellent question. We'll have to monitor the decisions of TC39 and adapt our choices then.

GraalVM is currently built mostly around ECMAScript revisions. You can select whether you want to be (5, 2015, ...) 2019, or 2020, or 2021 compatible. For future-looking revisions, that involves some guesswork. We try to figure out if something will likely land in 2021 or 2022 (or later). But of course, such guesses won't be perfect.

chicoxyzzy (approved):

data-esnext.js
ljharb: You wouldn’t need guesswork if users selected proposed features individually - which is why babel moved to that model.

PR #1695, opened 2021-01-25
Add JerryScript 2.4.0

ossy-szeged:
Additional minor changes:
- Run asnyc tests in function, because some of them contains return statement
- Add historical results for new es5 and es2016plus tests


jerryscript.js
```diff
@@ -146,7 +146,7 @@ function runTest(parents, test, sublevel) {
         }
 
         if (src.includes('asyncTestPassed')) {
-            script += asyncTestHelperHead + m[1] + asyncTestHelperTail;
+            script += asyncTestHelperHead + '(function test() {' + m[1] + '})();' + asyncTestHelperTail;
```
ljharb, line 149: Why this change?

ossy-szeged: Because 6 async tests contain return statement, that's why they needed to run in a function:
es2016plus -> async functions -> return
es2016plus -> async functions -> throw
es2016plus -> async functions -> async methods, object literals
es2016plus -> async functions -> async methods, classes
es2016plus -> async functions -> async arrow functions
es2016plus -> async functions -> async function constructor

We already run non-async tests inside a function, but it was in a bug for async test cases.
But good point, this fix makes 4 tests pass on JerryScript 2.3.0 too. I'm going to update its results too
and update the commit log accordingly.

ljharb: Is this only a bug in the Jerryscript file? Meaning, are there other places it needs fixing?

Does it change any other test results besides v2.4 and v2.3?

ossy-szeged: > Is this only a bug in the Jerryscript file? Meaning, are there other places it needs fixing?

Good point, it seems hermes.js has the same code, maybe it needs to be fixed too. But I don't have hermes build to be able to verify it and update their results.
 
> Does it change any other test results besides v2.4 and v2.3?

No, only jerry 2.3 and 2.4 are affected, because older jerry doesn't have async support.

ljharb (approved):

chicoxyzzy (approved):

PR #1697, opened 2021-01-27
Add Safari TP 119 results

chicoxyzzy:
https://webkit.org/blog/11525/release-notes-for-safari-technology-preview-119/

### DIFF
--- a/data-es5.js
+++ b/data-es5.js
@@ -1606,6 +1606,7 @@ exports.tests = [
       edge13: true,
       firefox2: true,
       safari3_1: false,
+      safaritp: true,
       chrome5: false,
       chrome13: false,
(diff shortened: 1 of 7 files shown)


data-es5.js
```diff
@@ -1606,6 +1606,7 @@ exports.tests = [
       edge13: true,
       firefox2: true,
       safari3_1: false,
+      safaritp: true,
```
ljharb, line 1609: wow, this is an old one

ljharb (approved):

data-es5.js
chicoxyzzy: yes, it's great to see that JSC team has started to focus on compatibility recently!

PR #1706, opened 2021-03-12
Retiring ancient IE and old Edge versions

afmenez:
Retiring ancient IE and old Edge versions, according to the [MS IE & Edge Lifecycle FAQ](https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge).


ljharb:
Duplicated by #1707.

PR #1707, opened 2021-03-12
Retiring ancient IE and old Edge versions

afmenez:
Retiring ancient IE and old Edge versions, according to the [MS IE & Edge Lifecycle FAQ](https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge).

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -748,7 +748,7 @@
     "short": "IE 8",
     "release": "2009-03-19",
     "retire": "2016-01-12",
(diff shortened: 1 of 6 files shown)


ljharb (commented):
I don't mind the Edge ones, but many people still need to support IE 9, and IE 8 is a useful pivot point to display (IE 10 users have mostly migrated to IE 11 or beyond).

PR #1708, opened 2021-03-14
Add test for spread parameters after optional chaining

nicolo-ribaudo:
Ref: https://github.com/babel/babel/issues/13001

I couldn't manually verify the Firefox and Safari data, but @jridgewell wrote that they work in https://github.com/babel/babel/issues/13001 so I copied the same versions for the other optional chaining features.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -4591,6 +4591,28 @@ exports.tests = [
           hermes0_7_0: true,
         }
       },
+      {
+        name: 'spread parameters after optional chaining',
+        exec: function(){/*
+          var fn = null;
+          var n = null;
+          var o = {};
+
+          return fn?.(...[], 1) === void undefined && fn?.(...[], ...[]) === void undefined && o.method?.(...[], 1) === void undefined && n?.method(...[], 1) === void undefined;
+        */},
+        res : {
+          ie11: false,
+          firefox10: false,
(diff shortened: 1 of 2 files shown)


data-es2016plus.js
```diff
@@ -4591,6 +4591,26 @@ exports.tests = [
...
+
+          return fn?.(...[], 1) === void undefined && fn?.(...[], ...[]) === void undefined && o.method?.(...[], 1) === void undefined && n?.method(...[], 1) === void undefined;
+        */},
+        res : {
+          ie11: false,
+          firefox10: false,
+          firefox52: false,
+          firefox73: false,
+          firefox74: true,
+          chrome89: false,
```
ljharb, line 4611: is there a version of chrome where this ever worked?

ljharb (approved):

data-es2016plus.js
jridgewell: I downloaded an old Chrome 80 build, and it failed there, too. I don't think it's ever been supported.

nicolo-ribaudo: I'll add this so that older versions are marked as "no" and not as "unknown"

chicoxyzzy (approved):

nicolo-ribaudo:
Rebased to fix the merge conflict

PR #1726, opened 2021-05-26
Add `Object.hasOwn`

nicolo-ribaudo:
https://github.com/tc39/proposal-accessible-object-hasownproperty

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1407,6 +1407,28 @@ exports.tests = [
     opera10_50: false,
     safari12: false,
   }
+},
+{
+  name: 'Object.hasOwn',
+  category: STAGE3,
+  significance: 'small',
+  spec: 'https://github.com/tc39/proposal-accessible-object-hasownproperty',
+  exec: function () {/*
+    return Object.hasOwn({ x: 2 }, "x") === true;
+  */},
+  res: {
+    babel7corejs3: babel.corejs,
+    typescript3_2corejs3: typescript.corejs,
+    ie11: false,
+    chrome1: false,
+    chrome90: false,
+    edge18: false,
+    firefox2: false,
+    firefox90: false,
+    ie11: false,
+    opera10_50: false,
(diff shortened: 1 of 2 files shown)


data-esnext.js
```diff
@@ -1407,6 +1407,27 @@ exports.tests = [
...
     safari12: false,
   }
+},
+{
+  name: 'Object.hasOwn',
+  category: STAGE3,
+  significance: 'small',
+  spec: 'https://github.com/tc39/proposal-accessible-object-hasownproperty',
+  exec: function () {/*
+    return Object.hasOwn({ x: 2 }, "x") === true;
```
ljharb: This should also test that ToObject is called before ToPropertyKey, since it’s highly likely that impls will just copy over Object.prototype.hasOwnProperty and forget to swap the step ordering.

nicolo-ribaudo: Should I add it as a separate test or in the same one?

ljharb: i guess we could do subtests, then implementations get partial credit if it works and only has the wrong validation order

nicolo-ribaudo: Done :+1:

data-esnext.js
```diff
@@ -1407,6 +1407,56 @@ exports.tests = [
...
+        safari12: false,
+      }
+    },
+    {
+      name: "ToObject called before ToPropertyKey",
+      exec: function () {/*
+        let ok = !!Object.hasOwn;
+        try {
+          Object.hasOwn(null, { toString() { ok = false } });
+          return false;
```
ljharb:
```suggestion
          Object.hasOwn(null, { toString: function () { ok = false; } });
          return false;
```
let's not have the test be reliant on modern syntax, since this can be polyfilled

ljharb (commented):
LGTM pending comment

ljharb (approved):

zloirock (approved):

PR #1740, opened 2021-07-17
Move "Ergonomic brand checks for private fields" to ES2022

nicolo-ribaudo:
They reached Stage 4 in the last meeting

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -5716,6 +5716,35 @@ exports.tests = [
       }
     ]
   },
+  {
+    name: 'Ergonomic brand checks for private fields',
+    category: '2022 features',
+    significance: 'small',
+    spec: 'https://github.com/tc39/proposal-private-fields-in-in',
+    exec: function () {/*
+      class A {
+        #x;
+        static check(obj) {
+          return #x in obj;
+        }
+      }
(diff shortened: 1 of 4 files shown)


ljharb (approved):

data-es2016plus.js
```diff
@@ -5714,6 +5714,36 @@ exports.tests = [
...
+    name: 'Ergonomic brand checks for private fields',
+    category: '2022 features',
+    significance: 'small',
+    spec: 'https://github.com/tc39/proposal-private-fields-in-in',
+    exec: function () {/*
+      class A {
+        #x;
+        static check(obj) {
+          return #x in obj;
+      rhino1_7_13: false
```
nicolo-ribaudo: Whops, this is wrong (it was also there before this PR)

```suggestion
```

ljharb: nice catch; it'd have been a noop at least, so it wouldn't break anything.

nicolo-ribaudo: Oh right, I didn't even realize it's a valid labeled statement.

chicoxyzzy (approved):
This needs rebase now. Sorry for that. Otherwise it looks good

nicolo-ribaudo:
Rebased!

PR #1741, opened 2021-07-21
Update Chrome and Safari data and add Error.cause test

chicoxyzzy:
Also see comments

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -5742,6 +5742,7 @@ exports.tests = [
       ie11: false,
       opera10_50: false,
       safari12: false,
+      safari15: true,
       rhino1_7_13: false
     }
   },
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1317,6 +1317,7 @@ exports.tests = [
         firefox85: firefox.nightly,
         firefox90: true,
         chrome85: false,
+        chrome92: true,
         safari14_1: {
           val: 'flagged',
           note_id: 'safari-at-method',
@@ -1351,6 +1352,7 @@ exports.tests = [
         firefox85: firefox.nightly,
         firefox90: true,
         chrome85: false,
+        chrome92: true,
         safari14_1: {
           val: 'flagged',
           note_id: 'safari-at-method'
@@ -1398,6 +1400,7 @@ exports.tests = [
         firefox85: firefox.nightly,
         firefox90: true,
         chrome85: false,
+        chrome92: true,
         safari14_1: {
           val: 'flagged',
           note_id: 'safari-at-method'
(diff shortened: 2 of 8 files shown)


data-esnext.js
```diff
@@ -1431,7 +1434,7 @@ exports.tests = [
     babel7corejs2: true,
     babel7corejs3: true,
     chrome1: false,
-    chrome91: true,
+    chrome92: false,
```
chicoxyzzy: This one was added mistakenly probably

data-es2016plus.js
```diff
@@ -5742,6 +5742,7 @@ exports.tests = [
       ie11: false,
       opera10_50: false,
       safari12: false,
+      safari15: true,
```
chicoxyzzy, line 5745: Tested both on macOS Monterey and iOS 15

data-esnext.js
```diff
@@ -1543,6 +1548,26 @@ exports.tests = [
...
   ]
+},
+{
+  name: 'Error.cause property',
+  category: STAGE3,
+  significance: 'small',
+  spec: 'https://github.com/tc39/proposal-error-cause',
+  exec: function () {/*
+      var error = new Error('error', { cause: 'cause' })
+      return error.hasOwnProperty('cause') && error.cause === 'cause';
```
ljharb, line 1566: we may want to test all 8 of the error constructors instead of just this 1

ljharb (approved):

data-esnext.js
nicolo-ribaudo: This code returns `true` for me in Chromium 90 :thinking: 

```js
(() => {
      let ok = false;
    class A {
      static { ok = true; }
    }
    return ok;
 })();
```

nicolo-ribaudo: Oh I had the "Experimental JavaScript" flag enabled.

chicoxyzzy: Thanks, I will add it under flag

chicoxyzzy: @nicolo-ribaudo are you sure it's Chrome 90? I see Chrome 91 mentioned [here](https://v8.dev/features/class-static-initializer-blocks). I have only Chrome 92 installed on this machine, so can't check

data-esnext.js
chicoxyzzy: done

ljharb (approved):

data-esnext.js
chicoxyzzy: It was 91 in https://github.com/kangax/compat-table/pull/1716, so I'm going to merge current version for now, we can change to 90 later if necessary

nicolo-ribaudo: I use Brave (which is Chromium-based), and the `chrome://version` page says `Chromium: 90.0.4430.212` and `V8 9.0.257.29`.

Probably it's flagged under Chrome 90 and unflagged in Chrome 91.

chicoxyzzy: It's not available in Chrome 92 to 94 by default

PR #1746, opened 2021-08-02
Class static initialization blocks support in Chrome 94

chicoxyzzy:
https://bugs.chromium.org/p/v8/issues/detail?id=11375#c4

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1439,6 +1439,7 @@ exports.tests = [
       note_id: 'ch-static-init-blocks',
       note_html: 'The feature has to be enabled via <code>harmony_class_static_blocks</code> flag.'
     },
+    chrome94: true,
     safari13: false,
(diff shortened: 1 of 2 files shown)


ljharb (approved):
i mean i guess, but it seems premature before it's actually released - things get pulled from the release train all the time.

chicoxyzzy:
It works on Canary Version 94.0.4595.3 (Official Build) canary (x86_64) for me

ljharb:
Sure, but things can still be pulled from canary too. There’s no rush really, is there?

chicoxyzzy:
The proposal is enabled in Chrome 94 Dev and Canary is 95 now, so I think we can merge this now

ljharb:
It’s fine, but I’d intended it to wait until release. Dev/canary don’t really count.

PR #1753, opened 2021-09-01
Move some stage 3 features to ES2022

chicoxyzzy:
...per Aug-Sep 2021 TC39 meeting

Bonus: actualization of Chrome and Safari TP versions


ljharb (approved):

data-es2016plus.js
```diff
@@ -5759,6 +5759,228 @@ exports.tests = [
...
+          safari14_1: {
+            val: 'flagged',
+            note_id: 'safari-at-method'
+          },
+          safaritp: {
+            val: 'flagged',
+            note_id: 'safari-at-method'
+          },
+          graalvm21: graalvm.es2022flag,
+          rhino1_7_13: false
```
zloirock, line 5830: Please, add `core-js` to those results.

chicoxyzzy: is it enabled by default for both Array and %TypedArray%?

zloirock: And for `String` that's missed in this PR.

PR #1754, opened 2021-09-06
feat(browsers): update versions for chromium based browsers

Eutrepe:


environments.json
```diff
@@ -3341,14 +3372,30 @@
     "equals": "chrome89",
     "short": "OP 75",
     "release": "2021-03-24",
-    "obsolete": false
+    "obsolete": true
```
ljharb: i'm confused, why are we marking all these things as obsolete?

ljharb (changes requested):
the new operas and edges and chrome canary seem fine, but let's separate out the obsoleteness changes to a different PR.

environments.json
Eutrepe: AD1. Im restored the obsolete field to its previous value in this PR.
AD2. I marked all stable versions older by 2 than the newest as obsolete for the readability of the table

ljharb (approved):

environments.json
```diff
@@ -2633,15 +2640,15 @@
     "short": "Edge 87",
     "release": "2020-11-19",
     "equals": "chrome87",
-    "obsolete": true
+    "obsolete": false
```
chicoxyzzy, line 2643: why?

Eutrepe: I forgot about it, but I do not understand why, since I added a few omitted versions to each browser, I should not add "obsolete" to "true" or "very" to the older ones. After all, without this, the readability of the page will be terribly impaired.

environments.json
```diff
@@ -2499,17 +2499,24 @@
...
-    "full": "Chrome 94 Canary",
+    "full": "Chrome 94 Beta",
     "family": "V8",
     "short": "CH 94",
-    "release": "2021-07-20",
+    "release": "2021-08-21",
+    "unstable": true
+  },
+  "chrome95": {
+    "full": "Chrome 94 Canary",
```
chicoxyzzy:
```suggestion
    "full": "Chrome 95 Canary",
```

environments.json
ljharb: right but this specific change is making an obsolete version NOT be obsolete

ljharb:
Did you delete your fork before the PR landed??

PR #1758, opened 2021-09-11
Update data for recent Duktape versions

omasanori:


ljharb (approved):

environments.json
```diff
@@ -4140,6 +4140,30 @@
     "short": "DUK 2.3",
     "platformtype": "engine",
     "release": "2017-09-23",
+    "obsolete": true
```
afmenez: This is more than 4 years old, shouldn't it be "very" instead of true?

omasanori: Thank you for reviewing! OK, I was not sure if it is very obsolete or just obsolete.

omasanori:
Changed all v1 versions to `"obsolete": "very"` as `v1-maintenance`  branch is inactive now. Also, all v2 versions released in 2018 or earlier are marked obsolete.

afmenez:
Great, but you need to rebuild for this to take effect.

omasanori:
Good catch! Rebuilt the pages.

omasanori:
Rebased.

chicoxyzzy (approved):

PR #1765, opened 2021-10-08
Add new test for destructuring param with id collision

jridgewell:
Safari 15 has issue with destructuring parameters when a function expression's id collides with a destructured parameter identifier.

https://bugs.webkit.org/show_bug.cgi?id=220517


data-es6.js
```diff
@@ -17291,6 +17291,48 @@ exports.tests = [
...
+      exec: function(){/*
+        var d = function d([d]) { return d };
+        if (!d([true])) return false;
+
+        try {
+          eval('var f = function f([id, id]) { return id }');
+          return false;
+        } catch (e) {
+          return e instanceof SyntaxError;
+        }
```
ljharb, line 17306: i wonder if we could make this test *pass* when destructuring isn't supported? that way the only `false` values will be those that implemented destructuring incorrectly

jridgewell: I can wrap everything in evals, add a check for basic destructuring, and early exit.

But, I think what's really needed is a "bug" check on top of another test. Failures would be listed, and absence would imply that it's supported (or explicitly listed as supported, as I imagine 15.1 will need).

ljharb: interesting, that sounds like a larger refactor.

I think for now this is fine, but it feels like the `false` results for engines that don't have destructuring in the first place (like opera) are a bit useless :-)

ljharb (approved):

data-es6.js
jridgewell: Removed the extra false results.

data-es6.js
```diff
@@ -17291,6 +17291,44 @@ exports.tests = [
         hermes0_7_0: true,
         rhino1_7_13: false
       }
+    },
+    {
+      name: 'duplicate identifier',
+      exec: function(){/*
+        var d = function d([d]) { return d };
+        if (!d([true])) return false;
```
JLHwung:
```suggestion
        if(!(function d([d]) {})) return false;
```
Golf down the reproducing example a bit.

jridgewell: I was actually checking that it returned the not-function binding, but I see I forgot the check explicitly for `!== true`.

chicoxyzzy (approved):

PR #1767, opened 2021-10-17
fix: undefined Safari15_0 environment in data-es6

JLHwung:


data-es6.js
```diff
@@ -17314,7 +17314,8 @@ exports.tests = [
         closure: true,
         typescript1corejs2: true,
         firefox2: true,
-        safari15_0: false,
+        safari13: false,
```
JLHwung, line 17317: I can confirm Safari 13 does not work, either.

ljharb (approved):

chicoxyzzy (approved):

PR #1773, opened 2021-10-26
es5: more Number.prototype.toExponential test cases

ljharb:


zloirock:
You forgot IE11- / Edge 14- round case and `Infinity.toExponential(Infinity)` https://github.com/zloirock/core-js/blob/2e0cc1e32652eeaf34daf3f5a47086810e1b17d0/packages/core-js/modules/es.number.to-exponential.js#L21-L40

ljharb:
Thanks; I'd skipped `Infinity.toExponential(Infinity)` since it perfectly overlaps with `NaN.toExponential(Infinity)`'s results, but it's better to be through. I'll add the round case as well.

zloirock:
IIRC some implementations (I don't remember what) passed only one of them.

ljharb:
I haven't been able to find any in my manual testing, but please let me know if you recall.

zloirock:
BTW adding FF86- to feature detection of your package can be dangerous - in too many cases, the implementation will return incorrect results with big fraction digits number, for example `123.456.toExponential(17)`. But modern (ES2018+) implementations extend the maximum fraction digits number to 100.

chicoxyzzy (approved):

PR #1779, opened 2021-11-30
Adding recent TypeScript 4.x releases

afmenez:
Adding TypeScript 4.1 to 4.5.


ljharb (approved):

chicoxyzzy (approved):
New TS versions support logical assignment and most of class features, but we can add that in another PR

PR #1791, opened 2022-02-12
Custom compat tables

### LINKED ISSUE #1198: Problem with the "platform type" header for empty categories

### LINKED ISSUE #1790: Would a PR making custom compat-tables easier be welcomed?
Hi, 

I'm wondering if a PR that would make building custom compat-tables easier be welcomed.

I'd like to add the following options to build.js:
- specify which file to use instead of environments.json
- ability to exclude the testresults for the current browser

And then changes to build.js/master.js to hide non-applicable stuff from the UI, like:
- engines from the legenda that aren't referenced in the provided environments file
- unused platformtype headings
- unused obsolete/unstable checkboxes
- etc

Currently at least NodeJS and Rhino hosts their own version of compat-tables (containing just NodeJS/Rhino versions), based on a very old version of this repo.

Not sure if NodeJS would switch, but for Rhino I'd rather be able to build our custom compat-tables using the latest version of this repo, instead of some modified fork.

I did some prototyping and got the basics going, but before putting in the effort to make it a polished PR, I'd like  to know if it has any chance of getting accepted.

Paul

p-bakker:
As discussed in #1790 a prototype to allow creating custom compat tables

Usage `node build environments=./custom.json excludecurrent`

With `custom.json` being a custom environments file, for example having the following content:
```
{
  "rhino1_7_13": {
    "full": "Rhino 1.7.13",
    "short": "Rhino 1.7.13",
    "family": "Rhino",
    "platformtype": "engine",
    "release": "2020-09-02"
  },
  "rhino1_7_14": {
    "full": "Rhino 1.7.14",
    "short": "Rhino 1.7.14",
    "family": "Rhino",
    "platformtype": "engine",
    "release": "2022-01-06"
  }
}
```

This PR is a draft, just to get the discusion going. Needs cleanup + more:
- Hide engines from the legenda that aren't used in the provided environments file
- Hide unstable/obsolete checkboxes if the provided environments file doesn't contain any unstable/obsolete browsers
- Remove sort by Engine Type, if only one engine type is is used with the environments file
- Remove entire platformType header if the browsers in the environments file are all of the same platformType

One question: the code in build.js isn't using modern JavaScript features. Is that on purpose?  

Closes #1790 
Closes #1198

### DIFF
--- a/build-utils/interpolate-all-results.js
+++ b/build-utils/interpolate-all-results.js
@@ -1,14 +1,41 @@
 var parseEnvsVersions = require("./parse-envs-versions");
 
-module.exports = function interpolateAllResults(tests, envs) {
+module.exports = function interpolateAllResults(tests, envs, customResults = {}) {
   var envsTree = buildEnvsTree(envs);
 
   tests.forEach(function (t) {
+    const customRes = customResults[t.name];
+
     // Calculate the result totals for tests which consist solely of subtests.
     [].concat(t.subtests || t).forEach(function (e) {
-      interpolateTestResults(e.res, envsTree);
+      const res = e.res;
+
+      if (customRes) {
+        if (e === t) { // not a subtest
+          Object.assign(res, customRes);
+        } else {
+          Object.assign(res, customRes.subtests[e.name]);
+          delete customRes.subtests[e.name];
+        }
+      }
+
+      interpolateTestResults(res, envsTree);
     });
+
+    if (customRes?.subtests) {
+      const invalidSubTests = Object.keys(customRes.subtests)
+      if (invalidSubTests.length) {
+        console.warn(`Invalid custom subtest results for '${t.name}'`, invalidSubTests);
+      }
+    }
+
+    delete customResults[t.name];
   });
+
+  const invalidTests = Object.keys(customResults);
+  if (invalidTests.length) {
+    console.warn(`Invalid custom results`, invalidTests);
+  }
 };
 
 function interpolateTestResults(res, envsTree) {
--- a/build.js
+++ b/build.js
@@ -463,9 +463,9 @@ function dataToHtml(skeleton, rawBrowsers, tests, compiler, customResults) {
   },{});
 
   // Adjust platformType header colspans to match # of browsers displayed by default
-  Object.keys(noOfBrowserPerPlatformType).forEach(function (platformType) {
-	  var count = noOfBrowserPerPlatformType[platformType];
-      $('table thead tr:first-child th#' + platformType + '-header').attr('colspan', count);
+  Object.keys(noOfBrowserPerPlatformType).forEach(platformType => {
+    const count = noOfBrowserPerPlatformType[platformType];
+    $(`table thead tr:first-child th#${platformType}-header`).attr('colspan', count);
   });
 
   function getHtmlId(id) {
--- a/build.js
+++ b/build.js
@@ -33,7 +33,17 @@ var cheerio = require('cheerio');
 var fl = require('fast-levenshtein');
 // var child_process = require('child_process');
 
-var useCompilers = String(process.argv[2]).toLowerCase() === "compilers";
+var useCompilers = false;
+
+process.argv.slice(2).forEach(function(arg) {
+	var parts = String(arg).toLowerCase().split('=');
+
+	switch (parts[0]) {
+		case 'compilers':
+			useCompilers = true;
+			break;
+	}
+});
 
 var STAGE2 = 'draft (stage 2)';
 
--- a/build.js
+++ b/build.js
@@ -33,11 +33,12 @@ var cheerio = require('cheerio');
 var fl = require('fast-levenshtein');
 // var child_process = require('child_process');
 
-const {values: {environments: environmentsPath = './environments.json'}, flags, positionals} = parseArgs();
+const {values: {environments: environmentsPath = './environments.json', customResults: customResultsPath }, flags, positionals} = parseArgs();
 
 const environments = JSON.parse(fs.readFileSync(__dirname + path.sep + environmentsPath, 'utf-8'));
 const useCompilers = positionals.includes('compilers');
 const excludeCurrentBrowser = flags.x === true;
+const customResults = customResultsPath && JSON.parse(fs.readFileSync(__dirname + path.sep + customResultsPath, 'utf-8'));
 
 var STAGE2 = 'draft (stage 2)';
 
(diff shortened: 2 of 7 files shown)


build.js
```diff
@@ -32,7 +31,27 @@ var path = require('path');
...
+			useCompilers = true;
+			break;
+		case 'environments':
+			environments = require(parts[1]);
+			break;
+		case 'excludecurrent':
+			excludeCurrentBrowser = true;
+			break;
+	}
+});
```
ljharb: let's use an args parser instead of rolling our own - https://www.npmjs.com/package/@pkgjs/parseargs will eventually be built into node, so that might be a good choice

build.js
```diff
@@ -32,7 +31,27 @@ var path = require('path');
...
+
+process.argv.slice(2).forEach(function(arg) {
+	var parts = String(arg).toLowerCase().split('=');
+
+	switch (parts[0]) {
+		case 'compilers':
+			useCompilers = true;
+			break;
+		case 'environments':
+			environments = require(parts[1]);
```
ljharb: dynamic requires should be avoided; can we use fs.readFileSync and JSON.parse instead of require, since this is just json?

build.js
```diff
@@ -32,7 +31,27 @@ var path = require('path');
...
+			break;
+		case 'excludecurrent':
+			excludeCurrentBrowser = true;
+			break;
+	}
+});
+
+if (!environments) environments = require('./environments');
```
ljharb: this seems like it could be handled as a default in the args parser

ljharb (commented):
seems reasonable so far

build.js
p-bakker: Looking at the example for the parseargs package, what would be your preferred setup? 
- keep `compilers` as a positional?
- excludecurrent as a flag? If so, suggestions as to which one? `-e`?

So something like `-e --environments=./custom.json compilers`?

build.js
p-bakker: I assume you mean a default value when destructuring  the returnvalue from `parseArgs(argv, options);`, as I don't see any way to provide defaults to the parseArgs method.

If so, is it ok to use stuff like desctructuring in build.js, given that pretty much all the code in there now looks like ES5

build.js
ljharb: That seems fine to me (altho maybe x for exclude)

build.js
ljharb: Yes, that’s fine with me; the runtime must be ES3 but the build script can use modern node.

build.js
p-bakker: Using [@pkgjs/parseargs](https://www.npmjs.com/package/@pkgjs/parseargs) means upping the minimal NodeJS version for running build.js to NodeJS 15(.14.0), as the package makes use of `??=` assignment operator.

That acceptible?

p-bakker:
Having to make some adjustments to the `es*/skeleton.html` files, I'm wondering if I should be bothered to look at the `es*/compiler-skeleton.html` files as well. 

Are they still being used? They don't seem to be hosted on https://kangax.github.io/compat-table/

build.js
ljharb: Yes, i think requiring node 16 is fine.

p-bakker:
@ljharb appreciate it if you could have another look at the direction of this PR: in https://github.com/kangax/compat-table/pull/1791/commits/35bd089b3c583a7b5a5c74f7ebbd36e6aad2dbee I added the option to augment the testresults already available in this repo, as to allow supplying testdata, for example for nightlies of environments that are already part of this repo or to supply test results for environments that aren't in this repo (yet)

Also updated the readme with the new options and addresses previous review comments

If this all still looks good, last step will be to hide some more things form the UI, based on which environments are included

build-utils/interpolate-all-results.js
```diff
@@ -1,14 +1,41 @@
 var parseEnvsVersions = require("./parse-envs-versions");
 
-module.exports = function interpolateAllResults(tests, envs) {
+module.exports = function interpolateAllResults(tests, envs, customResults = {}) {
```
ljharb, line 3: can you elaborate on what customResults is for?

ljharb (commented):
Seems pretty straightforward.

It'd probably be good to have separate commits here; one for the "exclude current browser" option, etc.

build-utils/interpolate-all-results.js
p-bakker: This is to be able to supply build.js with test results for browser/browserVersions that aren't part of this repo.

As the test results aren't in json files, but in an exported object in the data-esXxxx.js files, i couldn't come up with another way to provide builds.js with results for for example a nightly.

Also see the changes I made to the readme, in which I (tried to) explain how this would be used.

BTW: did you see my question about whether the es*/compiler-skeleton.html are still being user? I think they are used when using the `compilers` argument, but it looks to me that the resulting pages aren't hosted anywhere, so might this all be obsolete stuff that could be removed?

ljharb: What about instead providing an alternative list of browsers/browserVersions?

ok so "customResults" is more like "additionalResults"? it seems like maybe it'd be better for the caller to compose data-esX itself and make its own result object.

i have no idea about the compiler-skeleton stuff; before removing it, i'd want to know what it was for and why it's not being used anymore.

p-bakker: > What about instead providing an alternative list of browsers/browserVersions?

Not sure I follow: that is what what the `environments` param allow you to do

> ok so "customResults" is more like "additionalResults"? 

correct

>  it seems like maybe it'd be better for the caller to compose data-esX itself and make its own result object.

I did look at such an approach, but as the test result data is inside the data-es*.js modules, I don't see a way to 'inject' additional testresults into those, except manually or by basically maintaining a modified implementation all the data-es*.js modules, which is something I don't like, because that is verly likely to get out of sync.

Another idea I had was just to refactor all the data-es*.js files to read their test results from another json file, instead of the test results being stored inside the module. But I was thinking that might break existing tooling or process to update the restresults

> i have no idea about the compiler-skeleton stuff; before removing it, i'd want to know what it was for and why it's not being used anymore.

I haven't gotten to the bottom of it, but I think the es*/compiler-skeleton.html files are used to create .dedicated html files output files for just compilers. I think it may be that somewhere in the past, the currently hosted index.html files did not contain compiler results and they were hosted on separate pages. But it doesn't look like those separate compiler pages are still generated and hosted.

Da664458658574875567:
Parece que proporcionaste información sobre dos versiones diferentes del software Rhino  cool !!!!

p-bakker:
Closing this PR in favor of #1881 and yet to create follow-up PRS that offer a cleaner, more versatile approach

PR #1796, opened 2022-02-28
Add Chrome 98 to 100

nicolo-ribaudo:
The data changes are:
- _static class methods use [[Define]]_ is fixed in Chrome 98 (https://github.com/babel/babel/pull/14295#discussion_r814398466)
- _findLast_ and _findLastIndex_ have been added in Chrome 97 (https://developer.chrome.com/blog/new-in-chrome-97/#new-array-prototypes)


data-esnext.js
```diff
@@ -1362,6 +1362,7 @@ exports.tests = [
         ie11: false,
         chrome1: false,
         chrome90: false,
+        chrome97: true,
```
ljharb, line 1366: might be good to confirm it's false in chrome96

ljharb (approved):

data-esnext.js
nicolo-ribaudo: I cannot find where to download Chrome 96, but it's not present in V8 9.6 (which should be the one used by Chrome 96).

ljharb: verified on browserstack.

chicoxyzzy (approved):
Thanks Nicolò!

PR #1801, opened 2022-04-20
add node 18 and chrome 101

JLHwung:


data-es2016plus.js
```diff
@@ -6498,7 +6498,7 @@ exports.tests = [
         res: {
```
JLHwung, line 6498: The `actual` output in Node 16.0 is
```
[
  'global',
  'ignoreCase',
  'multiline',
  'dotAll',
  'unicode',
  'sticky',
  'hasIndices'
]
```
Yet `/a/dgimsuy.flags` does give `"dgimsuy"`. It seems to me `hasIndices` for some reason is called after other flags but the return value is somehow reordered. Has this issue been reported to Chromium?

ljharb: Yes, it's an open bug. I added the test case specifically because https://npmjs.com/regexp.prototype.flags is needed to patch it in latest node/chrome.

ljharb (approved):

data-es2016plus.js
```diff
@@ -6472,7 +6472,7 @@ exports.tests = [
         */},
         res: {
           safari15: true,
-          chrom100: true,
```
JLHwung, line 6475: We should probably throw on unknown browsers in `res`.

ljharb: sounds like a great improvement

chicoxyzzy (approved):

PR #1802, opened 2022-04-20
Throw when subtest.res includes unknown browser

JLHwung:
Per https://github.com/kangax/compat-table/pull/1801#discussion_r854570469

This PR includes commits from #1801. I will rebase when it gets merged.


package.json
```diff
@@ -23,6 +23,7 @@
     "esdown": "latest",
     "espree": "latest",
     "esprima": "latest",
+    "fast-levenshtein": "^3.0.0",
```
JLHwung, line 26: I pick this library because it's already in our n_m:
```
compat-table@0.0.0 /tmp/compat-table
└─┬ cheerio@0.20.0
  └─┬ jsdom@7.2.2
    └─┬ escodegen@1.14.3
      └─┬ optionator@0.8.3
        └── fast-levenshtein@2.0.6
```

ljharb:
I think this could just include the relevant commit, since it shouldn't conflict with the other PR?

package.json
ljharb: altho you picked v3, and v2's already what's there :-p

ljharb (approved):

package.json
JLHwung: Yeah. 🙈 My use-bleeding-edge progressive half eventually outweighs the other pick-one-from-node-modules conservative half.

ljharb:
(i intentionally rearranged the commits so that the names were all fixed before the commit the throw on bad ones landed)

chicoxyzzy (approved):
that's awesome!

PR #1803, opened 2022-04-25
Update test for ShadowRealm (nee Realms) proposal

jjgriego:
ShadowRealm is the new name for the old Realms proposal [1][2], and the API is
simplified; I adjusted the test code accordingly.

[1] https://github.com/tc39/proposal-shadowrealm (note that the old URI redirects here, too)
[2] Renaming discussion here: https://github.com/tc39/proposal-shadowrealm/issues/321

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -73,14 +73,14 @@ exports.tests = [
   ],
 },
 {
-  name: 'Realms',
+  name: 'ShadowRealm',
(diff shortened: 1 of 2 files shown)


ljharb (approved):
This is great, but it might be nice to add an actual `evaluate` and `importValue` test as well?

chicoxyzzy (approved):

PR #1804, opened 2022-05-03
Firefox 100 released

afmenez:
Firefox 100 was released today.

PR #1805, opened 2022-05-03
`includes`: note FF99+ fails; split subtests

ljharb:
See https://github.com/kangax/compat-table/pull/1804#issuecomment-1116206973 and https://bugzilla.mozilla.org/show_bug.cgi?id=1767541


chicoxyzzy (approved):
👍

PR #1808, opened 2022-05-09
Add results for Hermes 0.8.1 to 0.11.0

robhogan:
Adds results for [Hermes](https://hermesengine.dev/) versions:

0.7.0 (already mostly present)
0.8.0
0.9.0
0.10.0
0.11.0

I'm part of the React Native team at Meta, and we're looking to give more visibility to Hermes feature support as we move towards [Hermes as default](https://reactnative.dev/blog/2021/10/26/toward-hermes-being-the-default) for React Native. Thanks for providing this platform!

(NB: In practice, because Hermes is focused on bytecode performance over syntax support, JS support in React Native is heavily augmented by Babel - I'm planning to follow up this PR with the addition of "React Native" as a runtime from RN 0.68 - effectively this will be Hermes + RN's bundled Babel preset)

Preview: https://robhogan.github.io/compat-table/es6/#hermes0_11_0

CC @neildhar

(Currently in draft to give the Hermes team a chance to review footnotes, etc.)

PR #1809, opened 2022-05-09
Add an 'updateResult' utility function to upsert values into `data-*` files

robhogan:
As far as I could tell, the current accepted practice to update `data-` files is to use a script to suggest the changes to a console and then to make manual updates. 

This is a lot of work for new runtimes, and doesn't lend well to automating as part of a runtime release process.

This is a simple utility to parse and update `res` objects in place. It may be called from the test runner scripts - I used it to generate https://github.com/kangax/compat-table/pull/1808.

Because it uses an actual parser it should be robust to syntax gotchas like code-in-strings, but it makes a few basic assumptions that hold for the data files as they are now, like:
 - Data files have an `exports.tests = [ /* ... */ ]` section
 - Tests must have an object-literal `res` property, or an array-literal `subtests` property.

### DIFF
--- a/hermes.js
+++ b/hermes.js
@@ -289,7 +289,7 @@ function runTest(parents, test, sublevel) {
                 if (expect === success) {
                     // Matches.
                 } else {
-                    updateResult(parents[0], parents.slice(1).concat([test.name]), hermesKey, success ? 'true' : 'false');
+                    updateResult(parents[0], parents.slice(1).concat(test.name), hermesKey, success ? 'true' : 'false');
                     testOutOfDate++;
                     console.log(testPath + ': test result added or updated, previously: ' + expect + ', new: ' + success);
                 }
@@ -309,7 +309,7 @@ function runTest(parents, test, sublevel) {
     }
 
     if (test.subtests) {
-        test.subtests.forEach(function (v) { runTest(parents.concat([test.name]), v, sublevel + 1); });
+        test.subtests.forEach(function (v) { runTest(parents.concat(test.name), v, sublevel + 1); });
     }
 }
 
--- a/hermes.js
+++ b/hermes.js
@@ -1,7 +1,7 @@
 /*
  *  Node.js test runner for running data-*.js tests with Hermes 'hermes' command.
  *
- *  Reports discrepancies to console; fix them manually in data-*.js files.
+ *  Reports discrepancies to console and updates them automatically in data-*.js files.
  *  Expects 'hermes' to be already built.  Example:
  *
  *  $ node hermes.js --hermes-bin /path/to/hermes --suite suitename
@@ -12,6 +12,7 @@
 var fs = require('fs');
 var child_process = require('child_process');
 var console = require('console');
+var updateResult = require('./update-result');
 
 var argv = require('yargs/yargs')(process.argv.slice(2))
     .option('hermes-bin', {
@@ -21,7 +22,7 @@ var argv = require('yargs/yargs')(process.argv.slice(2))
     .option('suite', {
         alias: 's',
         type: 'string',
-        choices: ['all', 'es5', 'es6', 'es2016plus'],
+        choices: ['all', 'es5', 'es6', 'es2016plus', 'esnext'],
         default: 'all'
     })
     .option('test-name', {
@@ -61,7 +62,7 @@ var hermesKey = (function () {
         encoding: 'utf-8'
     });
 
-    var m = /Hermes release version:\s+(\d)\.(\d)(?:\.(\d))?/.exec(stdout);
+    var m = /Hermes release version:\s+(\d+)\.(\d+)(?:\.(\d+))?/.exec(stdout);
     if (m) {
         return 'hermes' + m[1] + '_' + m[2] + (m[3] ? '_' + m[3] : '');
     }
@@ -78,7 +79,7 @@ var hermesKeyList = (function () {
             continue;
         }
         res.push(k);
-        if (k === hermesKey) {
+        if (k === hermesKey) {
             // Include versions up to 'hermesKey' but not newer.
             break;
         }
@@ -287,13 +288,10 @@ function runTest(parents, test, sublevel) {
 
                 if (expect === success) {
                     // Matches.
-                } else if (expect === void 0 && !success) {
-                    testOutOfDate++;
-                    console.log(testPath + ': test result missing, res: ' + expect + ', actual: ' + success);
-                    // if (testFailure >= 2) throw new Error(testCount)
                 } else {
+                    updateResult(parents[0], parents.slice(1).concat([test.name]), hermesKey, success ? 'true' : 'false');
                     testOutOfDate++;
-                    console.log(testPath + ': test result out of date, res: ' + expect + ', actual: ' + success);
+                    console.log(testPath + ': test result added or updated, previously: ' + expect + ', new: ' + success);
                 }
 
                 // if (test.name === 'repeated parameter names is a SyntaxError') {
@@ -311,9 +309,7 @@ function runTest(parents, test, sublevel) {
     }
 
     if (test.subtests) {
-        var newParents = parents.slice(0);
-        newParents.push(test.name);
-        test.subtests.forEach(function (v) { runTest(newParents, v, sublevel + 1); });
+        test.subtests.forEach(function (v) { runTest(parents.concat([test.name]), v, sublevel + 1); });
     }
 }
 
--- a/jerryscript.js
+++ b/jerryscript.js
@@ -1,7 +1,7 @@
 /*
  *  Node.js test runner for running data-*.js tests with JerryScript 'jerry' command.
  *
- *  Reports discrepancies to console; fix them manually in data-*.js files.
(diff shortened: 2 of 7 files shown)


ljharb:
Most runtimes we support are browsers, so the primary workflow is to manually update results as they’re checked. Command-line runtimes sometimes have a script that updates the files automatically.

package.json
```diff
@@ -13,6 +13,7 @@
   "private": true,
   "dependencies": {
     "@babel/standalone": "latest",
+    "acorn": "8.7.1",
```
ljharb:
```suggestion
    "acorn": "^8.7.1",
```
no dep anywhere should ever be pinned; that’s what lockfiles are for

ljharb (changes requested):
How is this intended to be used, exactly?

robhogan:
Thanks for looking at this so quickly. I work with the Hermes engine team so we have an interest in (semi-)automating updates to `compat-table` as part of the engine release process.

I've pushed my changes to `hermes.js` to show how I've used the utility - basically we modify the data file in place on a `compat-table` checkout. The changes can be reviewed by eye, footnotes manually added if necessary, and then pushed up for a PR. It eliminates the error-prone step of taking the console output and manually transcribing the results into the data files.

ljharb:
https://github.com/kangax/compat-table/blob/gh-pages/duktape.js, https://github.com/kangax/compat-table/blob/gh-pages/jerryscript.js, https://github.com/kangax/compat-table/blob/gh-pages/nashorn.js, and https://github.com/kangax/compat-table/blob/gh-pages/rhino.js seem like they would benefit from the same approach?

robhogan:
I’d say so. Happy to update those too if the approach looks good to you.

robhogan:
(Done)

hermes.js
```diff
@@ -78,7 +79,7 @@ var hermesKeyList = (function () {
             continue;
         }
         res.push(k);
-        if (k === hermesKey) {
+        if (k === hermesKey) {
```
ljharb, line 82: i'm confused what the change is here

hermes.js
```diff
@@ -311,9 +309,7 @@ function runTest(parents, test, sublevel) {
     }
 
     if (test.subtests) {
-        var newParents = parents.slice(0);
-        newParents.push(test.name);
-        test.subtests.forEach(function (v) { runTest(newParents, v, sublevel + 1); });
+        test.subtests.forEach(function (v) { runTest(parents.concat([test.name]), v, sublevel + 1); });
```
ljharb:
```suggestion
        test.subtests.forEach(function (v) { runTest(parents.concat(test.name), v, sublevel + 1); });
```
unless test.name can be an array

jerryscript.js
```diff
@@ -197,12 +198,10 @@ function runTest(parents, test, sublevel) {
 
             if (expect === success) {
                 // Matches.
-            } else if (expect === void 0 && !success) {
-                testOutOfDate++;
-                console.log(testPath + ': test result missing, res: ' + expect + ', actual: ' + success);
             } else {
+                updateResult(parents[0], parents.slice(1).concat([test.name]), jerryKey, success ? 'true' : 'false');
```
ljharb:
```suggestion
                updateResult(parents[0], parents.slice(1).concat(test.name), jerryKey, success ? 'true' : 'false');
```

nashorn.js
```diff
@@ -116,12 +117,10 @@ function runTest(parents, test, sublevel) {
 
             if (expect === success) {
                 // Matches.
-            } else if (expect === void 0 && !success) {
-                testOutOfDate++;
-                console.log(testPath + ': test result missing, res: ' + expect + ', actual: ' + success);
             } else {
+                updateResult(parents[0], parents.slice(1).concat([test.name]), 'nashorn' + jjsKey, success ? 'true' : 'false');
```
ljharb:
```suggestion
                updateResult(parents[0], parents.slice(1).concat(test.name), 'nashorn' + jjsKey, success ? 'true' : 'false');
```

rhino.js
```diff
@@ -119,12 +120,10 @@ function runTest(parents, test, sublevel) {
 
             if (expect === success) {
                 // Matches.
-            } else if (expect === void 0 && !success) {
-                testOutOfDate++;
-                console.log(testPath + ': test result missing, res: ' + expect + ', actual: ' + success);
             } else {
+                updateResult(parents[0], parents.slice(1).concat([test.name]), 'rhino' + rhinoKey, success ? 'true' : 'false');
```
ljharb:
```suggestion
                updateResult(parents[0], parents.slice(1).concat(test.name), 'rhino' + rhinoKey, success ? 'true' : 'false');
```

update-result.js
```diff
@@ -0,0 +1,161 @@
...
+    indentLength,
+    resNode,
+    resultKey,
+    result
+  );
+  fs.writeFileSync(
+    suiteFile,
+    srcString.substring(0, resNode.start) +
+      newRawResString +
+      srcString.substring(resNode.end)
```
ljharb:
```suggestion
    srcString.slice(0, resNode.start) +
      newRawResString +
      srcString.slice(resNode.end)
```
same throughout; slice is preferred over substr/substring.

update-result.js
```diff
@@ -0,0 +1,161 @@
...
+) {
+  var newResRaw;
+  var existingValue = maybeGetPropertyValue(resNode, resultKey);
+  if (existingValue) {
+    newResRaw =
+      srcString.substring(resNode.start, existingValue.start) +
+      result +
+      srcString.substring(existingValue.end, resNode.end);
+  } else {
+    var indent = new Array(indentLength).fill(" ").join("");
```
ljharb: `new Array` should never be used; `fill` doesn't untaint a sparse array.
```suggestion
    var indent = Array.from({ length: indentLength }, () => ' ').join('');
```
altho in this case:
```suggestion
    var indent = ' '.repeat(indentLength);
```

update-result.js
```diff
@@ -0,0 +1,161 @@
...
+      bodyNode.type === "ExpressionStatement" &&
+      bodyNode.expression.type === "AssignmentExpression" &&
+      bodyNode.expression.left.type === "MemberExpression" &&
+      bodyNode.expression.left.object.type === "Identifier" &&
+      bodyNode.expression.left.object.name === "exports" &&
+      bodyNode.expression.left.property.type === "Identifier" &&
+      bodyNode.expression.left.property.name === "tests" &&
+      bodyNode.expression.right.type === "ArrayExpression"
+    );
+  });
```
ljharb, line 129:
```suggestion
  var testsArrayNode = bodyNodes.find((bodyNode) => (
    bodyNode.type === "ExpressionStatement"
    && bodyNode.expression.type === "AssignmentExpression"
    && bodyNode.expression.left.type === "MemberExpression"
    && bodyNode.expression.left.object.type === "Identifier"
    && bodyNode.expression.left.object.name === "exports"
    && bodyNode.expression.left.property.type === "Identifier"
    && bodyNode.expression.left.property.name === "tests"
    && bodyNode.expression.right.type === "ArrayExpression"
  ));
```

update-result.js
```diff
@@ -0,0 +1,161 @@
...
+  if (!property || !property.value) {
+    return null;
+  }
+  return property.value;
+}
+
+function getTestName(testNode) {
+  var nameNode = maybeGetPropertyValue(testNode, "name");
+  return nameNode ? nameNode.value : null;
```
ljharb, line 152:
```suggestion
  return nameNode?.value;
```

hermes.js
robhogan: Weirdly, it was an NBSP previously. Webstorm flagged it.

robhogan:
Thanks for the review @ljharb - the reason for the retro syntax though is the jshint lint setup for this project. For example, applying one of the suggestions above (which I'd normally very much agree with!):

```
> npm run lint
update-result.js: line 147, col 52, 'arrow function syntax (=>)' is only available in ES6 (use 'esversion: 6').
```

I'm not exactly sure what that implies about runtime feature support, but I've tried to be as conservative as I remember how to be (my ES3 is a little rusty!), mostly for consistency's sake - hence avoiding `Array.from`, `repeat` etc - it feels a bit awkward to use newer runtime features while working with a linter that forbids `const`. 

These constraints make sense in the test code where we obviously need to cater for the lowest common denominator, but probably shouldn't apply to scripts written for Node JS, like this one or the various engine harnesses. I'm not sure how best to reconcile that - different lint settings for test code vs scripts? Different linter? Exclude scripts from lint?

ljharb:
ha, ok fair enough :-)

just ignore those suggestions for now; we should update from jshint to eslint anyways, and then we can use overrides so that scripts like this don't have to meet the same standards.

robhogan:
Sounds good to me - I'll assume modern runtime support and apply all the suggestions the linter is happy with

p-bakker:
Not sure why this PR went nowhere, but #1881 achieves the same, without having to use a parser to modify the module code AND will allow custom Compat Tables, like the (vary outdated) custom compat Table we have at https://rhino.github.io/compat/engines.html

PR #1821, opened 2022-07-03
'Array find from last' and 'Hashbang Grammar' moved to Stage 4

afmenez:
https://github.com/tc39/proposals/commit/3abfcf0e96878be0f20084904f25e113e0b0e73b
https://github.com/tc39/proposals/commit/1482a975504428518b44fe0b6adbc5ffe5b54a79


ljharb:
Let’s wait until the upstream PR is merged tho.

zloirock:
Missed tests for typed arrays.

afmenez:
There's already a "finished (stage 4)" category, would it be OK to move the features there until the upstream merge?

ljharb:
Both are now merged. I’m not sure we should bother having that category, since the window for stage 4+unmerged is rarely long.

ljharb (approved):

chicoxyzzy (approved):

zloirock (changes requested):
> Missed tests for typed arrays.

ljharb:
@zloirock those weren't there before - why would we block updating the stage on adding missing tests?

zloirock:
@ljharb they were missed in the early stages of this proposal. If it's moved to stable - such a significant component should not be missed. It can be added as copy-past with smoke testing for some minutes, so I have no ideas why it's still ignored.

ljharb:
I agree such a significant component shouldn't be missed - but that has no bearing on the currently incorrect stage information.

You are more than welcome to add them yourself, if it's so quick to write, but it would be highly inappropriate to demand that the author of this PR do something entirely unrelated to the purpose of the PR, I think.

As for "stable", stage 4 is no more stable than stage 3 - in both stages it's unlikely to change, but in both it could change, for the same reasons, so I think this is an incorrect characterization.

ljharb:
Given that this has two approvals, and the missing Typed Array tests are well out of scope for this PR, I'm going to merge this - that way someone can put up a separate PR to add the Typed Array tests without dealing with excessive merge conflicts.

zloirock:
This proposal was changed on the way to stage 4 - typed array methods are one of those changes. They are missed in some environments - for example, in your polyfills.

Now I have no desire to add anything to this repository more than a review.

zloirock:
Awesome.

ljharb:
That's fine, since this repo didn't have any tests to begin with, nobody would be relying on anything that was changed along the way.

As for my polyfills, I simply don't have any Typed Array polyfills nor intend to (polyfilling something that's only a good idea to use for performance reasons doesn't make any sense), so nothing is missing from them.

zloirock:
Let's remember why I stopped actively contributing to this project https://github.com/kangax/compat-table/pull/1556 - and now you merged a PR with a changes request. Hypocrisy -)

Typed array constructors / logic and methods that are not related to performance are a few different things. However, you can think whatever you want - you have a strange opinion about many different things.

zloirock:
BTW results that should be fixed by #1556 3 years ago are still wrong and Babel uses incorrect compat data for old Node versions.

ljharb:
This is all off topic for this PR. Manually verified data is welcome in another PR.

zloirock:
No, your behavior is not off-topic - why did you merge it before fixes? -)

ljharb:
I explained why, and there's no need for further back-and-forth argument here.

Missing tests is not "broken" and doesn't need to be "fixed" in a PR that's not adding a feature.

zloirock:
It's only your opinion - and I explained why it's needed. It's not only your repo - why do you think that the opinions of other maintainers make no sense?

PR #1824, opened 2022-07-25
Adds STAGE4 on the ESNext tab

afmenez:
Adds STAGE4 on the ESNext tab, and moves "Hashbang Grammar" and "Array find from last" proposals there.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -11,6 +11,7 @@ exports.skeleton_file = 'esnext/skeleton.html';
 
 var STAGE2 = 'Stage 2';
 var STAGE3 = 'Stage 3';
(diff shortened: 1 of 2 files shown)


zloirock (commented):
It should be added to `es2016plus` table to the new `2023 features` section.

ljharb (changes requested):
as mentioned, stage 4 features go into their ES year, and don't need their own category.

afmenez:
Yep, I tried to move to ES year first on #1821, but you said it was too soon...

ljharb:
Right - it’s either stage 3, or it moves to stage 4 when it’s merged into the spec. We don’t need to ack proposals that are stage 4 but not yet merged.

PR #1827, opened 2022-08-10
Add results for GraalVM 21.3.3 and GraalVM 22.2

jirkamarsik:
This PR adds a new runner for GraalVM and updates the test results to include the two latest supported versions of GraalVM.

The GraalVM results haven't been updated in a while. This PR cleans up the GraalVM-related data:
* Removed redundant entries in `data-*.js` files. A new line is added only when the status of a test changes from version to version.
* GraalVM versions which are no longer supported are marked as obsolete.
* The `npm` script for running the `node.js` test runner against GraalVM, `graalvm:node`, has been updated to:
    - cope with the unhandled rejection in the `es2016plus -> Promise.prototype.finally -> basic support` test and
    - to use a `node` binary specified by the user to avoid problems caused by `npm` inserting its own `node` from `node_modules`

This PR adds a new runner, `graalvm.js`, and a corresponding `npm` script, `graalvm:js`. This runner uses the GraalVM's `js` launcher directly, without having to go through GraalVM's Node.js implementation. This has the advantage of letting us run the tests using just the GraalJS standalone available here (https://github.com/oracle/graaljs/) and it lets us use flags which are incompatible with the Node.js mode of executing GraalVM JavaScript. The runner is also better at reporting discrepancies than the `node.js` runner and it can figure out which flags are needed for some specific tests while generating the relevant flagged annotations.

Finally, and most importantly, this PR adds test results for GraalVM 21.3.3 (the latest release in the GraalVM 21 line) and GraalVM 22.2 (the latest release in the GraalVM 22 line).


ljharb:
The runner seems to duplicate a lot of code, but i'm not sure if that's avoidable - can graal understand required modules?

ljharb:
Otherwise, pending some rebasing, this looks good to me.

jirkamarsik:
When using the `js` launcher, GraalVM understands ECMAScript module imports. When running inside of Node.js (using the `node` launcher available in GraalVM), it also understands `require`.

ljharb:
@jirkamarsik in node, `import` can bring in CJS modules. Is that the case in GraalVM?

jirkamarsik:
@ljharb There is an option that you can set when using the `js` launcher, `--js.commonjs-require`, which lets you import CJS modules using either `import` or `require`.

ljharb:
@jirkamarsik then could we use that to share code between the test runners, and avoid the duplication?

jirkamarsik:
@ljharb I've unified the standalone runners for GraalVM, Nashorn, Rhino, JerryScript and Hermes. They now share the vast majority of their code in `runner_support.js`. The individual runners mostly provide ways on how to sniff the engine's version using the launcher, how to run the engine with the correct flags and also allow for custom options to be passed in and direct the running of the tests.

As an added bonus, all of these runners now run the tests in an identical test harness (all the provided test builtins and shims for `setTimeout` are now unified). I've tested this on all 5 of the engines involved.

jirkamarsik:
I rebased my branch on top of the current `gh-pages` branch, so this should now be possible to merge.

chicoxyzzy (approved):

environments.json
```diff
@@ -5019,7 +5019,7 @@
     "note_id": "graalvm-node-mode",
     "note_html": "Executed in Node.js/JVM mode via <code>graalvm/bin/node --jvm</code>.",
     "release": "2021-01-19",
-    "equals": "graalvm19",
```
ljharb, line 5022: why are all these "equals" removed?

ljharb (commented):
I'm assuming the node, nashorn, rhino, duktape, hermes, and jerryscript runners all still work?

environments.json
jirkamarsik: Because they were redundant, they only pointed to the previous version in the line of GraalVM releases.

Newer versions of an engine already inherit the results of previous versions so such `"equals"` annotations are not needed. In fact, they are never used like this in the `environments.json` file. Instead, they are used when one version of an engine/browser is to be marked as equivalent to a version of a different engine/browser (usually because they are based on the same JavaScript implementation).

jirkamarsik:
> I'm assuming the node, nashorn, rhino, duktape, hermes, and jerryscript runners all still work?

Yes, I downloaded/built all of these engines and tested the runners on them.

ljharb (approved):

p-bakker:
Sry to post on an already merged PR, but am noticing that this Pr changed the output when running rhino.js

Whereas before you'd get an output like the one below, now with this PR we're getting all the exceptions logged that occur while running the tests, as a result of which you don't get a clean overview anymore of which test results got changed, because those lines are somewhere in between tons and tons of logged lines with errors

```
$ node rhino.js
rhino version is: Rhino 1.7.14

rhino result key is: test.res.rhino1_7_14
rhino key list for inheriting results is: [ 'rhino1_7_13', 'rhino1_7_14' ]

**** data-es2016plus ****

data-es2016plus -> static class fields -> static class fields use [[Define]]: test result missing, res: undefined, actual: false
data-es2016plus -> RegExp Match Indices (`hasIndices` / `d` flag) -> constructor supports it: test result missing, res: undefined, actual: false
data-es2016plus -> RegExp Match Indices (`hasIndices` / `d` flag) -> shows up in flags: test result missing, res: undefined, actual: false

**** data-es5 ****

data-es5 -> Array methods -> Array.prototype.unshift: [].unshift(0) returns the unshifted count: test result out of date, res: undefined, actual: true
data-es5 -> String properties and methods -> String.prototype.substr: test result out of date, res: undefined, actual: true

**** data-es6 ****

data-es6 -> well-known symbols -> Symbol.isConcatSpreadable, spreadable object with poisoned getter: test result out of date, res: undefined, actual: true
data-es6 -> Array.prototype methods -> Array.prototype.splice: test result out of date, res: undefined, actual: true

**** data-esintl ****

data-esintl -> Intl.Collator -> rejects invalid language tags: test result out of date, res: true, actual: false
data-esintl -> NumberFormat -> rejects invalid language tags: test result out of date, res: true, actual: false
data-esintl -> DateTimeFormat -> rejects invalid language tags: test result out of date, res: true, actual: false

**** data-esnext ****


**** data-non-standard ****

1147 tests executed: 411 success, 736 fail
10 tests are out of date (data-*.js file .res)
```

p-bakker:
I think it's due to the old rhino.js adding a try/catch around the testcode, whereas the new code in runner_support does not add that, see https://github.com/compat-table/compat-table/pull/1827/files#diff-9ae6f9d9f0960230222e4bfb5f9282bcb02bd4e8e4220272bdfe4bd7ef1db3a7L87 and  https://github.com/compat-table/compat-table/blob/gh-pages/runner_support.js#L173

Any chance of getting the try/catch back?

p-bakker:
Another issue: the testcode that the runner_support generates contains the print command: https://github.com/compat-table/compat-table/pull/1827/files#diff-9b9bd34c626fbbf7f6eb2705b73526127454f46fd93984d11209aa351f2d9183R175

I think this is something native to Rhino and most likely won't work on any of the other engines that now utilize the runner

And Rhino (up to the latest version) doesn't support console.log (which isn't part of the EcmaScript specification)

ljharb:
hmm, thanks. Would you be willing to send a PR to fix those two things?

p-bakker:
yeah, think I can, already got some fixes in place locally in a branch that builds on top of #1881.

Can extract them a bit later

ljharb:
Thanks. I’ve got that one on my list to review, but a smaller more focused fix for these issues should be landed first :-)

p-bakker:
One question though (already asked [here](https://github.com/compat-table/compat-table/pull/1881#issuecomment-1753250411) as well): can we upgrade the NodeJS dependency to a much newer version, so we can start making use of newer EcmaScript features like arrow functions, Template Literals, const, let etc. in the runners?

ljharb:
Then the runners wouldn’t be able to run on older engines, which is important. Those new features don’t add any new capabilities, it’s just style, so that doesn’t seem worth it.

p-bakker:
Why is it important that the runners are able to run on older engines? From my pov, when using the Rhino runner, NodeJS is just used as an automation tool, not for executing tests. and I think the same goes for the other runners (that now all make use of runner_support.js)

p-bakker:
And the @pkgjs/parseArgs package ([which you suggested me to use before](https://github.com/compat-table/compat-table/pull/1791#discussion_r805181740)) also has a minimum requirement of NodeJS14

ljharb:
Hmm, that's a fair point; as long as the tests themselves can run on older engines (including old node) then it's fine.

p-bakker:
@JLHwung sorry to drag you into this discussion, but I see you have updated results for NodeJS is several PR's in the past and I'm wondering exactly how the results are gathered for NodeJs

There's a node.js script in the repo, which seems to parse the index.html pages looking for scripts, executes them and writes the results to the console.

Is that script still used to collect test results for NodeJS?

And regardless of the answer to the question above: I'm assuming that the actual building of the compat table (using build.js) could be done with any NodeJS version?

Asking because am trying to determine if there's anywhere in the repo (outside the actual test code) we are tied to a specific NodeJs version or not. And so far my conclusion is that we're not, except for the node.js script if it indeed collects the test code from parsing the generated index.html files. Which means that build.js, master.js, runner_support.js and the .js files in build-utils could start using newer EcmaScript & NodeJS features.

ljharb:
I'll stress that there shouldn't be any urgency to use the newer language features; it's mostly just aesthetic.

JLHwung:
@p-bakker The nodejs results are mostly inferred from the chromium test results. There are a few Unicode tests that I have manually run on node.js as they are different to the chromium support. I don't know how to generate node.js test results either.

PR #1828, opened 2022-08-19
Add Deno info in environments.json

Wetinee:
Deno env is added here, which is sourced from https://github.com/mdn/browser-compat-data/blob/main/browsers/deno.json .
After updating the Deno env and running build, there's a problem here - The Deno column is added under the Mobile category, but it should be under the category of Server/runtimes. I've no idea why it behaves like that. Can anyone help me fix it?
image


zloirock (changes requested):
Some results should be overwritten. For example, https://github.com/denoland/deno/blob/2164f6b1eb7216c1045d547c94f26fe3ceaa9403/runtime/js/99_main.js#L6

data-es6.js
```diff
@@ -18316,7 +18319,7 @@ exports.tests = [
         var desc = Object.getOwnPropertyDescriptor(Object.prototype,"__proto__");
         var A = function(){};
 
-        return (desc
+        return (desc !== undefined
```
ljharb: this change makes it less strict; `desc` must be an object, all of which are truthy

ljharb (changes requested):
Seems like many of the older deno versions should be marked as "very" obsolete?

Wetinee:
@ljharb Why CI failed? Are there anything I can do?
image

ljharb:
That means the PR needs to be rebased on top of the latest `gh-pages` branch, and `npm run build` needs to be rerun to resolve the conflicts in index.html

data-es6.js
```diff
@@ -18193,6 +18193,7 @@ exports.tests = [
         ie11: true,
         firefox2: true,
         chrome5: true,
+        deno1_24: false,
```
zloirock: It's still marked as supported in 1.23-.

Wetinee: I changed it to deno1_3, it's not supported since deno 1.2.1 https://github.com/denoland/deno/commit/fa61956f03491101b6ef64423ea2f1f73af26a73
but I'm not sure whether older version should marked as unsupported or not.

data-es6.js
```diff
@@ -18193,6 +18193,7 @@ exports.tests = [
         ie11: true,
         firefox2: true,
         chrome5: true,
+        deno1_3: false,
```
zloirock: It was removed in [0.37](https://github.com/denoland/deno/releases/tag/v0.37.0), so I think in the scope of this PR it can be marked as from 1.0.

zloirock (approved):

data-es6.js
```diff
@@ -18316,7 +18319,7 @@ exports.tests = [
         var desc = Object.getOwnPropertyDescriptor(Object.prototype,"__proto__");
         var A = function(){};
 
-        return (desc
+        return (typeof desc === "object"
```
ljharb: again, this check needs to remain unchanged, because this change allows `null` while the previous line does not (i'm aware the `in` check will throw on null)
```suggestion
        return (desc
```

ljharb (changes requested):

data-es6.js
Wetinee: If it remains unchanged, the test result will be `undefined` as deno deleted `Object.prototype.__proto__`

ljharb: Then it should fail this test, or, we should be changing the test separately.

JLHwung:
```suggestion
        return (!!desc
```
I think the author's concern is that a test should always return a boolean value. In that case we can cast `desc` to boolean so that the test returns `false` instead of `undefined` on Deno.

ljharb: returning false and undefined are treated the same. Why does it need to return a boolean?

zloirock:
> Please run `npm run build` and commit the changes.

PR #1831, opened 2022-09-28
add new safari, edge, chrome, opera versions

Eutrepe:
Add new browser versions:

chrome: https://en.wikipedia.org/wiki/Google_Chrome_version_history
edge: https://learn.microsoft.com/en-us/deployedge/microsoft-edge-relnote-stable-channel
safari: https://developer.apple.com/safari/technology-preview/release-notes/
opera: https://blogs.opera.com/desktop/
opera android: https://forums.opera.com/category/20/opera-for-android?sort=newest_to_oldest


zloirock (changes requested):
Without adding compat data for them, it makes no sense.

environments.json
```diff
@@ -3311,11 +3339,18 @@
     "release": "2021-12-13",
     "obsolete": false
   },
+  "safari16": {
```
zloirock, line 3358: Between 15 and 16, 15.4 added some new ES features.

environments.json
```diff
@@ -5879,7 +5934,16 @@
     "platformtype": "mobile",
     "equals": "safari15",
     "release": "2020-06-08",
-    "unstable": true
+    "unstable": false
+  },
+  "ios16": {
```
zloirock, line 5977: The same about 15.4.

Eutrepe:
I checked the test results and made changes.

zloirock:
Safari 16 and the latest Safari TP added many new ES features, however, they are not covered by the `compat-table` test case. It's not related to this PR, so LGTM.

zloirock (approved):

zloirock (changes requested):
I checked one more time. You missed some iOS versions and forgot to mark some browser versions as obsolete.

ljharb (approved):
We generally don’t block new versions on compat data since they auto inherit the previous version’s results. Why do we need to do so here?

zloirock:
@ljharb ~~we~~ you.

And it's the reason why `compat-table` no longer can be considered as a proper compat data source.

And, as always, you had read only a couple of first comments and missed the last review.

ljharb:
No, me and @chicoxyzzy.

I agree with blocking on incorrect claims, but it’s perfectly fine if there’s a version whose data is accidentally incorrect - it’ll get fixed eventually.

Compat data doesn’t have to be, and simply can’t be, 100% correct at all times - it just has to be eventually correct, since it’s a moving target.

zloirock:
As always, you missed a part of the questions and comment only on all that you want. So is this fine for you? -)

image

---

With this approach, `compat-table` has lost its relevance - to 2019 it was almost complete - and it is not clear to me how projects like Babel now can still use a data source containing such outdated data.

ljharb:
If it’s incorrect, please file PRs to fix it. It’s not worsening the problem to frustrate contributors with atypical requirements.

Adding versions should always be done in a different PR than adding the data for them.

zloirock:
I'm curious why you bring this topic up here where it's fixed? Here already another issue -) I don't give a damn about your opinion, just like you don't care about mine - issues like this will be a pull request block.

ljharb:
That’s a very hostile way to interact with other maintainers. I do care about your opinion, but caring about someone’s opinion doesn’t mean one has to agree with it.

The majority of maintainers on this project have been running it a certain way. Until a majority of maintainers prefer something different, that will be the way the project is run. @chicoxyzzy, can you weigh in?

zloirock:
Even after the picture above, he still doesn't understand (or pretends?) that this is another issue...

zloirock:
> I do care about your opinion

[You have perfectly shown that you consider your opinion more important than the opinion of others.](https://github.com/kangax/compat-table/pull/1821#issuecomment-1228818107)

Eutrepe:
I added "obsolete" for more more browsers and added ios15.2 and ios15.4 (I think we don't need more)

zloirock:
@Eutrepe yes, sure. It seems you forgot to push it -)

Eutrepe:
Sorry, I have pushed now

zloirock (approved):

zloirock:
LGTM, thanks.

PR #1852, opened 2023-03-07
add `regexp-v-flag` compat data

JLHwung:


afmenez:
I have run the tests on Firefox 91 and 102, they don't work there either.

JLHwung:
@afmenez Thanks! I will add `firefox91` results and hopefully the `firefox102` result can be interpolated from 91 and 110.

JLHwung:
Moved to `es2016plus` now that the proposal is stage 4 as of the may meeting.

data-esnext.js
```diff
@@ -1437,7 +1438,7 @@ exports.tests = [
       }
     }
   ]
-},
+}
 ];
```
zloirock: It seems all changes in this file should be discarded.

esnext/index.html
```diff
@@ -10300,7 +11082,7 @@
...
   </div>
   <pre class="info-tooltip" style="display:none"></pre>
 </body>
```
zloirock: In this file - too.

zloirock (changes requested):
image

ljharb:
It achieved stage 4 this morning.

zloirock:
@ljharb it's es2016+ table, so it should be renamed.

ljharb:
Why? It contains ES2016 and later, including ES2023 and the eventual ES2024.

zloirock:
Like the rest features, it should indicate the version of ES where it's added or will be added.

ljharb:
That version is unknown until the release is cut (which for ES2024 happens in March 2024), which is why the page has worked this way for the better part of a decade.

JLHwung:
The category here is copied from

https://github.com/kangax/compat-table/blob/09d24e559a39884eee5749e5a60c1e3cc859196b/data-es2016plus.js#L6782

I think `finished (stage 4)` was a thing, of course it is a transient category and we will probably replace it by "es2024 features" in the future. It is an editorial issue, do you mark ES2024 after a proposal is advanced to stage 4? Or do you mark ES2024 after the spec is actually updated.

ljharb:
@JLHwung we usually mark ESX in June X, when the edition is approved by Ecma, or in March X, when I've cut the candidate.

zloirock:
IIRC previous features were marked before that. Let it be, I don't care.

zloirock (approved):

ljharb (approved):

PR #1859, opened 2023-06-02
Add missing Safari versions

afmenez:
Adding Safari 15.1, 15.3, 15.5, 15.6, 16.1..16.4 and respective iOS derivatives.


zloirock (approved):
The actual version is 16.5, so it makes sense to add it too.

ljharb (approved):

PR #1861, opened 2023-06-07
Firefox 116 now implements RegExp 'v' flag

afmenez:
Enabled on today's nightly build.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -6792,7 +6792,9 @@ exports.tests = [
           chrome111: chrome.harmony,
           chrome112: true,
           firefox91: false,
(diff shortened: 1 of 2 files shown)


zloirock (commented):
Are you sure that it will be enabled by default in stable releases? I can't find this info.

afmenez:
The implementation is here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1826574
There's no flag, so the only way to disable it would be reverting the merge.

zloirock (approved):
Let it be if you are sure that it is in the stable version.

ljharb (approved):

PR #1862, opened 2023-06-08
add compat data for safari bug

JLHwung:
Tested on BrowserStack Safari sessions. I guess the bug was fixed in Safari 16.0 but I don't have one in hand for testing.

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -16864,9 +16864,12 @@ exports.tests = [
     {
       name: 'duplicate identifier',
       exec: function(){/*
-        var d = function d([d]) { return d };
-        if (d([true]) !== true) return false;
-
+        try {
+          eval('var d = function d([d]) { return d };if (d([true]) !== true) return false;');
+        } catch (e) {
+          return !(e instanceof SyntaxError);
+        }
+        
         try {
           eval('var f = function f([id, id]) { return id }');
           return false;
@@ -16883,8 +16886,9 @@ exports.tests = [
         closure: true,
         typescript1corejs2: true,
         firefox2: true,
-        safari13: false,
-        safari15: false,
+        safari11: false,
+        safari16_2: false,
+        safari16_3: true,
         edge13: edge.experimental,
         edge14: true,
         xs6: true,


ljharb:
In Safari 16.2, the first test throws a SyntaxError, as does the second.

ljharb:
To be clear, that means the test *fails* on Safari 16.2, because the first thing in it isn't wrapped in a try/catch like the second one is.

JLHwung:
@ljharb Thanks. I thought the second test is also throwing SyntaxError on 16.2, which would return true in that case, updated  16.2 data now.

data-es6.js
```diff
@@ -16883,8 +16883,9 @@ exports.tests = [
         closure: true,
         typescript1corejs2: true,
         firefox2: true,
-        safari13: false,
-        safari15: false,
+        safari11: false,
+        safari15_6: false,
+        safari16_2: true,
```
ljharb:
```suggestion
        safari16_2: false,
```

ljharb: Note that it's also possible the test is wrong, since it's not in strict mode?

JLHwung: I think the test is fine: although it's not in strict mode, the array pattern `[id, id]` disqualifies simple parameter anyway. So duplicate bound names check should still kick in.

ljharb: right, but the syntax error in 16.2 i get is that you can't have a non-simple parameter list in sloppy mode at all.

JLHwung: In other destructuring tests, e.g.

https://github.com/kangax/compat-table/blob/2c4d894c9a9aa48d6f23f186d884c8d23e307e13/data-es6.js#L16734C1-L16742

the test functions don't have strict directive either. Does Safari 16.2 throw syntax error on them as well? Or does it only throw SyntaxError on that specific case?

ljharb: That code returns true, so I'm not sure what the difference is.

ljharb: ok so it definitely doesn't have to do with strict mode - `var d = function d([e]) { return d };` works fine while `var d = function d([d]) { return d };` throws a syntax error. I think the issue is that this test is just wrong, because the var d declaration needs to be inside the `try` (at which point it passes on Safari 16.2)

JLHwung: > var d = function d([e]) { return d }; works fine while var d = function d([d]) { return d }; throws a syntax error.

Yeah that's exactly the Webkit bug behaviour. I have updated the test, can you take a look if fails (expectedly) on Safari 16.2?

data-es6.js
```diff
@@ -16864,9 +16864,12 @@ exports.tests = [
     {
       name: 'duplicate identifier',
       exec: function(){/*
-        var d = function d([d]) { return d };
-        if (d([true]) !== true) return false;
-
+        try {
+          eval('var d = function d([d]) { return d };if (d([true]) !== true) return false;');
+        } catch (e) {
+          return !(e instanceof SyntaxError);
```
ljharb, line 16870:
```suggestion
          return e instanceof SyntaxError;
```
since this should be a syntax error, no?

JLHwung: No it should not be a syntax error. In
```js
var d = function d([d]) { return d };
```
Safari thought the parameter `d` is duplicated, but the other `d` is just the function name, not an actual parameter.
See also https://bugs.webkit.org/show_bug.cgi?id=220517.

ljharb: ahhhh ok, thanks, that clarifies things.

JLHwung:
@ljharb @zloirock Any suggestions?

zloirock (approved):

ljharb (approved):

Issue #1863, opened 2023-06-11
The verification of `RegExp.prototype.hasIndices` is incorrect


axiac:
The generated compatibility table says that `RegExp.prototype.hasIndices` is not available in any version of Node.js
The [JavaScript documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/hasIndices) says that `RegExp.prototype.hasIndices` is available since Node.js version 16.0.

I did not check with all versions newer than 16.0 but on Node.js 16.14, 16.18 and 18.12 it is available.
Check it here: https://onecompiler.com/nodejs/3zbayqw9a (Node.js 16.14.2).

The code fails because it compares two unsorted arrays. The arrays have the same elements but in different order.

ljharb:
Note that MDN isn't "the javascript documentation", it's just MDN.

You're right that node v16.0 introduced hasIndices.

axiac:
Several years ago, Mozilla, Microsoft, Google, Samsung and several other parties involved in standardization and development of the web technologies [agreed to keep the web documentation on MDN](https://blog.chromium.org/2017/10/building-unified-documentation-for-web.html).
For the JavaScript developers, MDN **is** the documentation of JavaScript (maybe except for the developers that have never read a page of documentation in their life and use Stack Overflow to _"learn how to code"_).
The ECMAScript documentation is difficult to read and targets the implementors, anyway.

ljharb:
It's not official Ecma/TC39 documentation; how it's considered by devs isn't relevant.

axiac:
Without those devs, your site is irrelevant.

PR #1864, opened 2023-06-11
fix detection of `RegExp.prototype.hasIndices`

axiac:
Solves issue #1863.


data-es2016plus.js
```diff
@@ -6641,6 +6641,8 @@ exports.tests = [
           var actual = [];
           var p = new Proxy({}, { get: function(o, k) { actual.push(k); return o[k]; }});
           Object.getOwnPropertyDescriptor(RegExp.prototype, 'flags').get.call(p);
+          expected.sort();
+          actual.sort();
```
ljharb, line 6645: I'm confused why this fixes anything - part of the spec requirements is that the list is already alphabetized here.

zloirock (commented):
Here missed some flags like `.unicodeSets` - this test is incorrect because of this.

Old V8 (< 11.1 -> Node < 20.0) implementations had a bug with an order of getting keys in `.flags` - it's a fair reason why this test fails in old Node versions.

data-es2016plus.js
axiac: Before sorting, on Node.js 16.19.1 `expected` and `actual` look like this:
```
expected: [
  'hasIndices',
  'global',
  'ignoreCase',
  'multiline',
  'dotAll',
  'unicode',
  'sticky'
]
actual: [
  'global',
  'ignoreCase',
  'multiline',
  'dotAll',
  'unicode',
  'sticky',
  'hasIndices'
]
```
Given that the order should not matter, isn't an object a better container for these values?

ljharb: order should matter.

axiac: Why should order matter? The flags are independent.

ljharb: The spec mandates the order.

axiac: The specs also require that the flag `hasIndices` is implemented but not all JavaScript engines support it at this moment. That's why this compatibility table exists, isn't it?

Node.js 16 implements `hasIndices` despite not returning the flag names in the expected order.
In fact, `RegExp.prototype.flags` returns them in the expected order. It's only the property descriptor the one that lists them in a different order.

ljharb: Which means node 16 isn’t compliant.

I’d be fine splitting this test into two - one that checks the functionality and existence of the accessors, and another that checks the ordering - but spec compliance is binary, and if the order is wrong, it fails.

axiac: I don't get it. Why is the order so important?

ljharb: Every single thing the spec mandates is equally critically important - compliance is binary, either something does, or doesn’t.

axiac:
I thought that the goal of this project is to help the developers find out what ES functionalities are implemented by each JavaScript engine.
Apparently you do not care about the developers.
:-(

ljharb:
@axiac it is, and the ordering of the flags is a part of that.

PR #1872, opened 2023-08-04
Update Closure Compiler v20230228

teppeis:
v20230228 passes the following ES2015 tests
- spread syntax for iterable objects > spreading non-iterables is a runtime error
- generators > yield * on non-iterables is a runtime error

> - A for-of loop transpiled down to ES5 should throw a TypeError if the object to be iterated has neither a Symbol.iterator property nor a length property.
>
> https://github.com/google/closure-compiler/wiki/Releases#february-28th-2023-v20230228


zloirock (approved):

ljharb (commented):
LGTM

PR #1875, opened 2023-09-12
Update environments

JLHwung:
Added the following environments:
- Chrome 116 - 120
- Deno 1.35, 1.36
- Safari & iOS 16.6
- Samsung 22
- Opera Mobile 77

PR #1876, opened 2023-09-12
Add unicode support subtests

JLHwung:
Added Unicode support subtests for the Unicode property escape feature and the Unicode `v` flag feature.

The Babel issue https://github.com/babel/babel/issues/15868 justifies the demand of the browser compat data on implemented Unicode versions. For example, `/\p{Extended_Pictographic}/u;` is supported on Chrome 68+ while the basic unicode property escape is supported on Chrome 64+.

In this PR we add compat data for popular engines: Chrome, firefox, Safari and Node.js.

- The Chrome data is extracted from the commit history of https://github.com/mathiasbynens/caniunicode and tested on the chromium build snapshot Windows builds available from https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html
- The firefox data is tested on firefox macOS builds available from https://download-installer.cdn.mozilla.net/pub/firefox/releases/
- The Safari data is not actually tested, they are generated from the history of the WebKit source file https://github.com/WebKit/WebKit/blob/6ec1bbcc639f159ceb4986135267b1b609c4acdb/Source/JavaScriptCore/ucd/PropertyValueAliases.txt, if you have access to the old Safari versions, please help me correct the compat data
- The Node.js data is tested on the official macOS builds.

/cc @mathiasbynens 

This PR is based on #1875 due to new environments dependency. I will rebase once that PR gets merged.


mathiasbynens:
This is awesome! Thank you so much for working on this :)

data-es2016plus.js
```diff
@@ -3281,28 +3281,191 @@ exports.tests = [
...
+          jerryscript2_3_0: false,
+          graalvm19: true,
+          hermes0_7_0: false,
+          reactnative0_70_3: true,
+          rhino1_7_13: false
+        }
+      },
+      {
+        name: 'Unicode 11',
+        // 2018	June 5
```
JLHwung, line 3312: Added the release date here so if we add new browser in the future, it can help determine the first supported version since any browser released earlier than 2018 June 5 probably does not support Unicode 11, unless they are built against the Unicode beta release.

data-es2016plus.js
```diff
@@ -3281,28 +3281,191 @@ exports.tests = [
...
+      */},
+        res: {
+          chrome67: false,
+          chrome68: true,
+          firefox77: false,
+          firefox78: true,
+          safari14: false,
+          safari14_1: true,
+          node10_9: false,
+          node11_0: true,
```
JLHwung, line 3324: The node.js support is different to the chromium support because node.js maintainers bump their ICU deps earlier than the chromium repo.

environments.json
```diff
@@ -7518,6 +7598,15 @@
...
     "release": "2023-06-26",
+    "obsolete": true
+  },
+  "opera_mobile77": {
+    "full": "Opera Mobile for Android 77",
+    "family": "Opera Mobile",
+    "short": "Opera Mobile 77",
+    "platformtype": "mobile",
+    "equals": "chrome115",
+    "release": "2023-09-01",
```
zloirock, line 7623: Awesome, but why does it add new environments?

nicolo-ribaudo:
It includes commits from https://github.com/kangax/compat-table/pull/1875

environments.json
nicolo-ribaudo: It includes commits from https://github.com/kangax/compat-table/pull/1875

zloirock: Oops, I missed it at the top.

zloirock (approved):

JLHwung:
Updated Chrome 117 unicode supports, same as Chrome 116.

JLHwung:
Also added Unicode subtests to the `v` flag feature since `/^\p{RGI_Emoji}$/v` also depends on Unicode versions. It is very likely that this subtest will share the same support data with the property escape subtests.

ljharb (approved):

Issue #614, opened 2015-08-31
Add a way to automatically generate & apply browser results


kangax:
As per https://github.com/kangax/compat-table/issues/556#issuecomment-136159253

I see it like that:
1. You'd click "export" in a browser and save the file as browser-data.json. You might be asked a few questions to be able to fill the browser definition (like the short/long name etc.).
2. You'd then run node ./merge-data.js --to es6-data.js --from browser-data.json and you'd get a modified es6-data.js with a new browser added. There would have to be a merge-data.js script that would be able perform that action.

PR #1881, opened 2023-10-09
Externalize tests

### LINKED ISSUE #1493: Store Test Results In JSON

### LINKED ISSUE #1879: Externalize results from tests
ATM each test and its results on all the different environments are co-located together in the data-*-js modules

This poses 2 problems:
- prevents automated updates of the results when tests start passing/failing
- prevents generating compat tables for a completely different set of environments

The later of these 2 points I've already have a (draft) PR for (#1791), but the flaw in that PR to allow passing in custom results is that the approach breaks down in linking the custom results to a test, if the name of a test would ever change. I'm not sure how often this happens, but with the current setup that would not be an issue, but once we'd allow providing external results, it would become problematic as it would cause the externally recorded/stored results to get disconnected from the tests in data-*.js.

As for the first point: as the results are stored in an object inside a module, unless you somehow modify the sourcecode of the module through code, there's no way afaics to automate the updating of the results, making it manual labour

So, I would propose the following:
- rename the data-*.js files to test-*.js
- externalize the results from these test-*.js files to plain json files, using the results-*.json naming convention
- use the name  of (sub)test to link the test (from test-*.js) to the results in results-*.js
- optional: allow overriding the displayed testname in the generated HTMl by providing the optional `displayName` in addition to the `name` on a (sub)test in test-*.js

This achieves the following goals:
- a testrunner could now be made to automatically modify the results-*.json file, as to allow to make updating the Compat Table less manual
- forces the name of a (sub)test to not get changed without thinking about it
- prepares the codebase for the next step, which is to allow providing alternative environments and results from the command line, which in turn allows [creating custom compat tables](https://github.com/kangax/compat-table/issues/1790)

Eventually this repo could then  even be spilt in a package to create Compat Tables in general and a repo for the kangas.github.io gh pages (as suggested by @ljharb in #1791)

Thoughts?

p-bakker:
This PR splits the tests and their results into 2 separate files, with the results being stored in JSON format, allowing automated modification, as to open up the possibility to eliminate the manual updating of test results

Additionally, this split makes it possible to create custom Compat Tables, for example with just all the different versions of one environment, including for example the nightly version. Note that are more enhancements possible for custom Compat Tables, even going as far as splitting this repo in a compat-table library and a Compat Table website (github pages) repo, but this are next steps imho

Is does mean that data-common.json is obsolete (and thus removed) as all the test results are not stored in JSON 

Closes #1879
Closes #1493
Checks off on of the actions in #965
Building block for #614
Building block for #1790

Did the conversion of data-*.js files > tests.*.js & results-*.json using the following script:
```
const fs = require('fs');

const extractResults = (entry, result) => {
	if (Array.isArray(entry)) {
		const mapping = {};

		entry.forEach(entry => {
			mapping[entry.name] = extractResults(entry);
		});

		return mapping;
	}

	return entry.res || extractResults(entry.subtests);
}

[
	'es5',
	'es6',
	'es2016plus',
	'esnext',
	'esintl',
	'non-standard',
].forEach(fileName => {
	console.log(`Processing 'data-${fileName}'`);

	const data = require(`./data-${fileName}`);

	// Extract just the results from the data-*.js modules, convert their format and export to result-*.json
	fs.writeFileSync(
		`results-${fileName}.json`, 
		JSON.stringify(
			extractResults(JSON.parse(JSON.stringify(data.tests))),
			null,
			'\t'
		),
		'utf8');

	// Read data-*.js, remove test results and write to tests-*.js
	fs.writeFileSync(
		`./test-${fileName}.js`,
		fs
			.readFileSync(`./data-${fileName}.js`, 'utf8')
			.replace(/,$(\s*)res *: \{.*?$\1}/gms, ''),
		'utf8'
	);

	// Check the conversion
	const tests = require(`./test-${fileName}`).tests;
	const results = require(`./results-${fileName}`);

	if (tests.length !== Object.keys(results).length) console.error(`Test mismatch: tests.length !== Object.keys(results).length (${tests.length} vs. ${Object.keys(results).length})`);

	if (!tests.every((entry, idx) => {
		return !entry.subtests || entry.subtests.length === Object.keys(results[entry.name]).length
	})) console.error(`Subtest mismatch`);
});
```


p-bakker:
Would it be ok to update the min nodejs version to 14 or 18? Want to use parseArgs either natively (Node 18) or via @pkgjs/parseArgs (Node 14)

p-bakker:
(re)marked this as Draft again, as I'm contemplating some  additional changes, but would already appreciate some feedback on the general direction of extracting the results into their own json files. 

Has that any change of getting accepted?

p-bakker:
Any feedback anyone?

Don't want to spend more time on this if the general direction of this PR, namely splitting the tests and results into separate files, is gonna be a no-go to begin with

p-bakker:
I guess the change in where the test results are stored would break [the mechanism in Babel that downloads and converts the compat data](https://github.com/babel/babel/blob/c446ff85c28e117ebf3cd72cd34ef358f1077aa8/packages/babel-compat-data/scripts/utils-build-data.js), but imho it would be easy to adapt Babel accordingly

p-bakker:
@ljharb @chicoxyzzy any chance to get some feedback on the general idea to extract the results out from the data-*.js files and into .json files? 

Is it worth my time to finish this PR and get it out of draft status so it can be considered for merge? Or is this general idea dead in the water?

ljharb:
It's a pretty major change so it's hard to develop an opinion about it, I think :-)

p-bakker:
Ok, but what does that mean exactly? 🙂

Does it mean you and/or others are actively thinking about whether or not this change is something you're willing to accept? 

Or do you need something additional from me to be able to make that decision? 

Or ....?

ljharb:
It means I'm actively thinking about it. I can't speak for the other maintainers.

p-bakker:
@ljharb how's the thinking going?

p-bakker:
@chicoxyzzy @ljharb sorry to nag again, but would really like to move on this...

So how about it? Will you support a change in the proposed direction?

ljharb:
I think I do like this direction, and if @chicoxyzzy is in favor of it, then this PR can move forward.

p-bakker:
Great!

So @chicoxyzzy what do you say?

chicoxyzzy:
This is great!

chicoxyzzy:
But I'd also want to know what @JLHwung and @nicolo-ribaudo think since such a change will affect Babel

nicolo-ribaudo:
This is fine for Babel: we need to update our build scripts, but I like this direction.

JLHwung:
Thank you, this PR looks good to me.

p-bakker:
Great!

I'll try and rebase & finish this PR asap, but it might be a week or two before I get to that 

Well keep you posted

p-bakker:
Closing this one in favor of a new, rebased PR: #1992

PR #1882, opened 2023-10-11
Updates to pr1827

p-bakker:
Fixes issues introduced in #1827, namely:
- runner_support.js mixing usage of the `print` and `console.log` commands, neither of which will both work on any environment
- exceptions occurring during running of tests cluttering the stdout

Includes extracting `createIterableHelper` out from runner_support, as it's used by the `node.js` script, which must remain working on older NodeJS versions (and thus should be authored in ES3), while runner_support.js does not have that ES3 requirement  (and thus can use newer syntax)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,9 +3,11 @@ es*/compilers/*
 .DS_Store
 .ruby-version
 _site/
-.idea/
 
 # Only apps should have lockfiles
 npm-shrinkwrap.json
 package-lock.json
 yarn.lock
+
+// temp file created by runner_support
+test.js
--- a/runner_support.js
+++ b/runner_support.js
@@ -1,25 +1,43 @@
+/* jshint esversion: 11 */
 var fs = require('fs');
+var createIterableHelper = require('./test-utils/testHelpers').createIterableHelper;
 
-exports.createIterableHelper =
-'function __createIterableObject(arr, methods) {\n' +
-'    methods = methods || {};\n' +
-'    if (typeof Symbol !== "function" || !Symbol.iterator)\n' +
-'        return {};\n' +
-'    arr.length++;\n' +
-'    var iterator = {\n' +
-'        next: function() {\n' +
-'            return { value: arr.shift(), done: arr.length <= 0 };\n' +
-'        },\n' +
-'        "return": methods["return"],\n' +
-'        "throw": methods["throw"]\n' +
-'    };\n' +
-'    var iterable = {};\n' +
-'    iterable[Symbol.iterator] = function(){ return iterator; };\n' +
-'    return iterable;\n' +
-'}\n' +
-'if (typeof global !== "undefined") {\n' +
-'    global.__createIterableObject = __createIterableObject;\n' +
-'}\n';
+function getKeys(key, family, environments) {
+    var res = [];
+    for (var k in environments) {
+        var env = environments[k];
+        if (env.family !== family) {
+            continue;
+        }
+        res.push(k);
+        if (k === key) {
+            // Include versions up to 'key' but not newer.
+            break;
+        }
+    }
+    return res;
+}
+
+function removeIndent(str) {
+    var indentation = /^[\t ]+/m.exec(str);
+    return str.replace(new RegExp('^' + indentation[0], 'gm'), '');
+}
+
+function testCode(exec) {
+	if (typeof exec === 'function') {
+        var src = exec.toString();
+        var functionBody = /^function\s*\w*\s*\(.*?\)\s*\{\s*\/\*([\s\S]*?)\*\/\s*\}$/m.exec(src);
+        if (functionBody) {
+            return '(function () { ' + removeIndent(functionBody[1]) + ' })()';
+        } else {
+            return '(' + src + ')()';
+        }
+    } else if (Array.isArray(exec)) {
+        return exec.map(function (e) { return testCode(e.script); }).join("; ");
+    } else {
+        return undefined;
+    }
+}
 
 exports.runTests = function runTests(runner, key, family, options) {
     options = options === undefined ? {} : options;
@@ -37,6 +55,7 @@ exports.runTests = function runTests(runner, key, family, options) {
     }
     var testName = options.testName;
     var bail = options.bail;
+    var logCommand = options.logCommand || 'console.log';
 
     var testCount = 0;
     var testSuccess = 0;
@@ -48,7 +67,7 @@ exports.runTests = function runTests(runner, key, family, options) {
 
     var asyncTestHelperHead =
         'function asyncTestPassed() {\n' +
-        '   print("[SUCCESS]");\n' +
+        '   ' + logCommand + '("[SUCCESS]");\n' +
         '}\n' +
         '\n' +
         'var jobQueue = [];\n' +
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -4,8 +4,13 @@ es*/compilers/*
 .ruby-version
 _site/
 .idea/
+.history/
+.project
```
ljharb: i'm not sure what these are, but specific tools' files should be in your global gitconfig, not added to every project you touch

test-utils/testHelpers.js
```diff
@@ -0,0 +1,22 @@
+// this file must only contain js code written in ES3 as it's used in tests
```
ljharb, line 1: this ideally would be enforced using eslintrc and "overrides" - is there a reason this *must* be a string? could it be an actual JS function, thus lintable, and when concatenated into a string elsewhere, use Function.prototype.toString?

p-bakker: If we don't store it as a string, we're relying on the (proper) implementation of Function.prototype.toString() in the executing environment, whatever that may me.

I cannot judge whether it's safe to rely on that... thoughts?

.gitignore
p-bakker: .project is from Eclipse and it's required to have any project in your workspace. In this case it can be omitted, as there's no relevant info in it, but that is not always the case, so it cannot be excluded globally, I guess similarly to `.idea/`?

.history is a VS Code (plugin?) thing which I think makes sense to exclude globally, will check

test-utils/testHelpers.js
ljharb: yeah, that's a fair point :-/ probably best to keep it as a string just in case.

.gitignore
ljharb: `.idea/` is just a default from github, it shouldn't be there either.

why can't `.project` be excluded in your global `.gitignore`?

p-bakker: See https://einverne.github.io/post/2016/04/should-i-git-ignore-project-file-under-eclipse.html

Basically: for many projects in Eclipse you want the .project file to be committed, hence it shouldn't be excluded globally

Should I remove `.idea/` then?

ljharb: That seems like an unfortunate design in Eclipse :-/

sure, let's remove `.idea/`

runner_support.js
```diff
@@ -37,6 +17,7 @@ exports.runTests = function runTests(runner, key, family, options) {
     }
     var testName = options.testName;
     var bail = options.bail;
+    var logCommand = options.logCommand ?? 'console.log';
```
ljharb: `||` would work fine here, ftr

runner_support.js
```diff
@@ -157,23 +138,33 @@ exports.runTests = function runTests(runner, key, family, options) {
...
                 } else if (/\bglobal\.test\b/.test(evalcode)) {
-                    script += 'global.test = function (expression) {\n' +
-                              '    if (expression) {\n' +
-                              '        console.log("[SUCCESS]");\n' +
-                              '    }\n' +
-                              '}\n' +
-                              evalcode;
+                    script += `
+                        global.test = function (expression) {
+                            if (expression) ${logCommand}("[SUCCESS]");
```
ljharb: let's keep the curly braces; don't want another gotofail :-)
```suggestion
                            if (expression) {
                                ${logCommand}("[SUCCESS]");
                            }
```

runner_support.js
```diff
@@ -157,23 +138,33 @@ exports.runTests = function runTests(runner, key, family, options) {
...
+                        } catch (e) {}
+                    `;
                 } else {
-                    script += 'var result = ' + evalcode + ';\n' + 
-                              'if (result) {\n' + 
-                              '    print("[SUCCESS]");\n' +
-                              '}\n';
+                    script += `
+                        try {
+                            if (eval(testCode)) ${logCommand}("[SUCCESS]");
```
ljharb:
```suggestion
                            if (eval(testCode)) {
                                ${logCommand}("[SUCCESS]");
                            }
```

runner_support.js
```diff
@@ -157,23 +138,33 @@ exports.runTests = function runTests(runner, key, family, options) {
...
+                        } catch(e) {}
```
ljharb:
```suggestion
                        } catch (e) {}
```

ljharb (commented):
It seems like you're adding try/catches in a number of places that didn't have them before; won't that mask errors in the test code we want caught?

p-bakker:
> It seems like you're adding try/catches in a number of places that didn't have them before; won't that mask errors in the test code we want caught?

Prior to #1827 there also were try/catched around the eval of the testcode:
- https://github.com/compat-table/compat-table/pull/1827/files#diff-9ae6f9d9f0960230222e4bfb5f9282bcb02bd4e8e4220272bdfe4bd7ef1db3a7L88
- https://github.com/compat-table/compat-table/pull/1827/files#diff-48b230a7f01f4595f2a9df8af1db79063c8a0e0458de67e4873877d98880cbb6L83
- https://github.com/compat-table/compat-table/pull/1827/files#diff-3744f0f40c1d8d08319f4dce83919405e334cf138b46cc16e9848cb88a1a5489L159
- etc.

.gitignore
p-bakker: done

runner_support.js
p-bakker: replaced

runner_support.js
p-bakker: done

runner_support.js
p-bakker: done

runner_support.js
p-bakker: done

.gitignore
ljharb: `.project` still should be removed, though, despite the inconvenience for yourself.

ljharb:
gotcha, so it's restoring them, good.

.gitignore
p-bakker: Done

ljharb (approved):
LGTM, but i'll wait for another approval before i merge

p-bakker:
I see the pretest stage fails due to linting. For now I fixed it to the specific file that had warnings, but going forward what's the desired way?

1. Globally allow newer js, limit to es3 only on the data-*.js files, node.js en master.js? Or
2. The other way around: globally limit to es3, allowing newer js on runner_support.js/build.js & the various runner .js files?

p-bakker:
Is there anything I must do now to move this PR along?

ljharb:
> I see the pretest stage fails due to linting. For now I fixed it to the specific file that had warnings, but going forward what's the desired way?
> 
> 1. Globally allow newer js, limit to es3 only on the data-*.js files, node.js en master.js? Or
> 2. The other way around: globally limit to es3, allowing newer js on runner_support.js/build.js & the various runner .js files?

i would go with option 2; the default should be the safest thing.

p-bakker:
> i would go with option 2; the default should be the safest thing.

k, that's what I figured as well and thus what I already did :-)

Just now waiting for that second approval...

p-bakker:
@chicoxyzzy any chance you can review this one?

chicoxyzzy (approved):

PR #1883, opened 2023-10-24
Add Node.js 21 and Node.js 22 data

JLHwung:
Also added Chrome 121-124 and updated Unicode 15.1 support.


chicoxyzzy (approved):

JLHwung:
@ljharb ping.

data-es2016plus.js
chicoxyzzy, line 1: Why bump versions of Chrome, Firefox and Node here? Is the support in previous versions unknown?

JLHwung: I bumped the versions here when I tested on new browsers and confirmed that they didn't support this feature. For example, if Chrome 118 does not support this feature, Chrome 117 will probably not support it either, but not vice versa.

Do you prefer both `chrome117` and `chrome118` entries here? Or I can add recent support data as well (this PR was drafted half a year ago).

data-es2016plus.js
```diff
@@ -3453,10 +3453,10 @@ exports.tests = [
       */},
         res: {
           ie11: false,
-          chrome117: false,
-          node20_0: false,
+          chrome118: false,
+          node21_0: false,
```
JLHwung, line 3459:
```suggestion
          node21_0: false,
          node22_0: true,
```
I can confirm Node.js 22 installed from `nvm` supports this feature.

ljharb (approved):
thanks for the ping, not sure how i missed this one 3 weeks ago.

any additional changes can be done in a followup (specifically, only manually verified entries should be present in the list, and none should ever be removed - so before merging this, i'm going to undo the removals)

PR #1891, opened 2024-05-08
Add Rhino 1.7.15

gausie:
- **Add new version environment**
- **Update instructions at the top of rhino.js**
- **Update newly passing tests for Rhino 1.7.15**
- **Fix some old tests that were erroneously marked as true, thus the new falses are not regressions**

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -6991,7 +6991,7 @@ exports.tests = [
           ie11: false,
           safari16_6: false,
           safaritp: true,
-          rhino1_7_15: true,
+          rhino1_7_15: false,
         }
       },
       {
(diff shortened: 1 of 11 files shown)


chicoxyzzy (approved):
LGTM
Can you please rebase and regenerate HTML files?

ljharb:
@chicoxyzzy ill take care of it :-) reminder that PRs still need two green checks before they can be landed.

ljharb:
(also, the consecutive falses are necessary and desired; every version that a human has verified should have an entry in the data)

ljharb (approved):

gausie:
Noted, thanks. Everywhere an old one was turned to false needs an extra
false added now then. I am ill at the moment so can get to it over the
weekend.

ljharb:
sounds good, we'll fix that in a followup.

PR #1892, opened 2024-06-10
Add 'instance class fields / resolving identifier in parent scope'

davidtaylorhq:
This is to detect https://bugs.webkit.org/show_bug.cgi?id=236843, which affects Safari 15 and below.

Discussion in https://github.com/babel/babel/issues/14289

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -5776,6 +5776,26 @@ exports.tests = [
           reactnative0_70_3: true,
           rhino1_7_13: false
         }
+      },
+      {
+        name: 'resolving identifier in parent scope',
+        exec: function () {/*
+          {
+            let a = ["hello world"];
+            class MyClass {
+              // The parenthesis below are required to trigger https://bugs.webkit.org/show_bug.cgi?id=236843
+              c = a[(0)]; 
+            }
+            return new MyClass().c === a[0];
+          }
+        */},
+        res: {
+          chrome124: true,
+          safari15: false,
+          safari16: true,
+          safari17: true,
+          safaritp: true,
+        }
       }
     ]
   },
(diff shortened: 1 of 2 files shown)


data-es2016plus.js
```diff
@@ -5776,6 +5776,39 @@ exports.tests = [
...
           rhino1_7_13: false
         }
+      },
+      {
+        name: 'resolving identifier in parent scope',
+        exec: function () {/*
+          {
+            let a = ["hello world"];
+            class MyClass {
+              c = a[(0)];
```
ljharb, line 5787: are the parens needed here to trigger the issue?

data-es2016plus.js
```diff
@@ -5776,6 +5776,39 @@ exports.tests = [
...
+          safari15: false,
+          safari16: true,
+          safaritp: true,
+          duktape2_0: false,
+          graalvm19: false,
+          graalvm20: true,
+          hermes0_7_0: false,
+          reactnative0_70_3: true,
+          rhino1_7_13: false
+        }
```
davidtaylorhq: Here, I have copy/pasted the data from 'instance class fields / public instance class fields', and then manually modified the entries for safari. Is that the correct approach? Or is there some automated system to collect this data?

data-es2016plus.js
davidtaylorhq: Yes, the parentheses are important. The JIT bug only occurs when parenthesis are included in the initializer.

ljharb: A comment to that affect on this line would be good, to ensure nobody removes them in the future.

data-es2016plus.js
ljharb: Ideally, results should only be present in the file if a human has manually verified them.

data-es2016plus.js
davidtaylorhq: Added :+1:

data-es2016plus.js
davidtaylorhq: Ok that makes sense. Updated to only include browsers I have manually tested

nicolo-ribaudo:
I can confirm that this passes in Firefox 125 and Chrome 74.

nicolo-ribaudo:
Also Firefox 69

chicoxyzzy (approved):

ljharb (approved):

nicolo-ribaudo:
Note: Firefox 69 and Chrome 74 are _not_ included in this PR yet! @davidtaylorhq do you mind opening a new one? :)

davidtaylorhq:
Thanks for the merge! Added those two extra data points in https://github.com/compat-table/compat-table/pull/1899

PR #1896, opened 2024-06-25
Typescript 5.5 released

afmenez:
Ditto.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -808,7 +808,7 @@
     "family": "TypeScript",
     "platformtype": "compiler",
     "release": "2022-08-25",
-    "obsolete": true,
+    "obsolete": "very",
     "test_suites": [
       "es6",
       "es2016plus",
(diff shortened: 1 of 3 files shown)


chicoxyzzy (approved):

chicoxyzzy (commented):
It also supports Set methods as far as I remember.

Note: I see that Set methods are shown as Stage 2 proposal, which is not true, as it's stage 4 (we can add it to ES2025), but we can fix that in another PR

ljharb (approved):
followup PRs are welcome

ljharb:
Filed #1897 for some stage changes.

afmenez:
Yes, it is on the release notes for 5.5:
"TypeScript 5.5 declares [new proposed methods for the ECMAScript Set type](https://github.com/tc39/proposal-set-methods).

Some of these methods, like union, intersection, difference, and symmetricDifference, take another Set and return a new Set as the result. The other methods, isSubsetOf, isSupersetOf, and isDisjointFrom, take another Set and return a boolean. None of these methods mutate the original Sets."

But I was unable to complete the tests on my machine, so I didn't update the results.  I did it for Chrome(#1895) and Firefox(#1898).

PR #1902, opened 2024-07-01
Updating ShadowRealm results for Firefox

afmenez:
Ditto.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -92,6 +92,12 @@ exports.tests = [
     ie11: false,
     firefox10: false,
     firefox52: false,
+    firefox103: false,
+    firefox107: {
+      val: 'flagged',
+      note_id: 'ff-shadow-realm',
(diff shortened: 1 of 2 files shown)


afmenez:
Firefox 128 will be released tomorrow, it would be nice if this could be merged before that to avoid the conflicts.

data-esnext.js
```diff
@@ -92,6 +92,12 @@ exports.tests = [
     ie11: false,
     firefox10: false,
     firefox52: false,
+    firefox103: false,
+    firefox104: {
+          val: 'flagged',
+          note_id: 'ff-shadow-realm',
+          note_html: 'The feature has to be enabled via <code>javascript.options.experimental.shadow_realms</code> setting under <code>about:config</code>.'
+        },
```
chicoxyzzy: Hmm... Probably spec has changed since then. Maybe we need to reflect that in the comment

afmenez: The last commit for this feature was on Firefox 107, I updated the patch to this version.

ljharb (approved):

afmenez:
Thank you!

PR #1903, opened 2024-07-01
Update Firefox results for the Set Methods feature

afmenez:
Ditto.


afmenez:
Firefox 128 will be released tomorrow, it would be nice if this could be merged before that to avoid the conflicts.

ljharb:
If that doesn't happen, you can either add to this PR, or base the new PR on top of this one?

afmenez:
It's 3 different PRs that would be affected.  I'll hold the new one until all three are committed or rejected.

afmenez:
Closing in favor of #1908 , which is more complete.

PR #1905, opened 2024-07-04
Add regexp modifiers tests

JLHwung:


JLHwung:
@chicoxyzzy @ljharb Rebased, PTAL. Now that Chrome has shippped regexp modifiers, Babel plans to add regexp modifiers support to preset-env, it will be great if we can get the compat data ready so that we don't have to enable regexp modifiers for latest chrome versions.

chicoxyzzy (approved):

ljharb (approved):

PR #1907, opened 2024-07-06
Add es2023 change-array-by-copy

### LINKED ISSUE #1874: add new es2023 Array.prototype methods
[ECMAScript 2023 introduced](https://tc39.es/ecma262/multipage/#sec-intro) new `Array.prototype` methods: `toSorted`, `toReversed`, `with`, `findLast`, `findLastIndex`, and `toSpliced`. The `findLast` and `findLastIndex` methods are already present on [the compatibility table](https://kangax.github.io/compat-table/es2016plus/), but the rest are missing. This request is to add them to the table for completion. Thank you.

cdrikd:
Add the 4 methods from https://github.com/tc39/proposal-change-array-by-copy. 

Fixes: #1874

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -6994,6 +6994,110 @@ exports.tests = [
     }
   },
   {
+    name: 'Change Array by copy',
+    category: '2023 features',
+    significance: 'small',
+    spec: 'https://github.com/tc39/proposal-change-array-by-copy',
+    subtests: [
+      {
+        name: "Array.prototype.toReversed",
+        exec: function () {/*
+          var arr = [1, 2, 3];
+          return arr.toReversed()[0] === 3 && arr[0] === 1;
+        */},
+        res: {
+          babel7corejs3: babel.corejs,
+          typescript3_2corejs3: typescript.corejs,
+          ie11: false,
+          chrome109: false,
+          chrome110: true,
+          edge109: false,
+          edge110: true,
+          firefox114: false,
+          firefox115: true,
+          firefox102: false,
+          safari15_5: false,
+          safari16: true,
+          node16_11: false,
+          node18_3: false,
+          node20_0: true,
+        }
+      },
+      {
+        name: "Array.prototype.toSorted",
+        exec: function () {/*
+          var arr = ['C', 'A', 'B'];
+          return arr.toSorted()[0] === 'A' && arr[0] === 'C';
+        */},
+        res: {
+          babel7corejs3: babel.corejs,
+          typescript3_2corejs3: typescript.corejs,
+          ie11: false,
+          chrome109: false,
+          chrome110: true,
+          edge109: false,
+          edge110: true,
+          firefox114: false,
+          firefox115: true,
+          firefox102: false,
+          safari15_5: false,
+          safari16: true,
+          node16_11: false,
+          node18_3: false,
+          node20_0: true,
+        }
+      },
+      {
+        name: "Array.prototype.toSpliced",
+        exec: function () {/*
+          var arr = ['A', 'C'];
+          return arr.toSpliced(1, 0, 'B')[1] === 'B' && arr[1] === 'C';
+        */},
+        res: {
+          babel7corejs3: babel.corejs,
+          typescript3_2corejs3: typescript.corejs,
+          ie11: false,
+          chrome109: false,
+          chrome110: true,
+          edge109: false,
+          edge110: true,
+          firefox114: false,
+          firefox115: true,
+          firefox102: false,
+          safari15_5: false,
+          safari16: true,
+          node16_11: false,
+          node18_3: false,
+          node20_0: true,
+        }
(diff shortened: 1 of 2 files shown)


data-es2016plus.js
```diff
@@ -6993,6 +6993,118 @@ exports.tests = [
...
+    name: 'Change Array by copy',
+    category: '2023 features',
+    significance: 'small',
+    spec: 'https://github.com/tc39/proposal-change-array-by-copy',
+    subtests: [
+      {
+        name: "Array.prototype.toReversed",
+        exec: function () {/*
+          var arr = [1,2,3];
+          return arr.toReversed()[0] === 3;
```
```suggestion
          var arr = [1, 2, 3];
          return arr.toReversed()[0] === 3 && arr[0] === 1;
```
to ensure the original isn't mutated

data-es2016plus.js
```diff
@@ -6993,6 +6993,118 @@ exports.tests = [
...
+          node16_11: false,
+          node18_3: false,
+          node20_0: true,
+        }
+      },
+      {
+        name: "Array.prototype.toSorted",
+        exec: function () {/*
+          var arr = ['C','A','B'];
+          return arr.toSorted()[0] === 'A';
```
```suggestion
          var arr = ['C', 'A', 'B'];
          return arr.toSorted()[0] === 'A' && arr[0] === 'C';
```

data-es2016plus.js
```diff
@@ -6993,6 +6993,118 @@ exports.tests = [
...
+          node16_11: false,
+          node18_3: false,
+          node20_0: true,
+        }
+      },
+      {
+        name: "Array.prototype.toSpliced",
+        exec: function () {/*
+          var arr = ['A','C'];
+          return arr.toSpliced(1, 0, 'B')[1] === 'B';
```
```suggestion
          var arr = ['A', 'C'];
          return arr.toSpliced(1, 0, 'B')[1] === 'B' && arr[1] === 'C';
```

data-es2016plus.js
```diff
@@ -6993,6 +6993,118 @@ exports.tests = [
...
+          node16_11: false,
+          node18_3: false,
+          node20_0: true,
+        }
+      },
+      {
+        name: "Array.prototype.with",
+        exec: function () {/*
+          var arr = ['A', 'X', 'C'];
+          return arr.with(1, 'B')[1] === 'B';
```
```suggestion
          return arr.with(1, 'B')[1] === 'B' && arr[1] === 'X';
```

ljharb (dismissed):

data-es2016plus.js
```diff
@@ -6993,6 +6993,118 @@ exports.tests = [
...
+          var arr = [1, 2, 3];
+          return arr.toReversed()[0] === 3 && arr[0] === 1;
+        */},
+        res: {
+          babel7corejs3: babel.corejs,
+          typescript3_2corejs3: typescript.corejs,
+          ie11: false,
+          chrome109: false,
+          chrome110: true,
+          edge109: false,
```
afmenez, line 7014: Recent Edge results are mirrored from chrome, no need to add them here.

data-es2016plus.js
```diff
@@ -6993,6 +6993,118 @@ exports.tests = [
...
+          edge110: true,
+          firefox114: false,
+          firefox115: true,
+          firefox102: false,
+          opera95: false,
```
afmenez: Recent Opera results are mirrored from chrome, no need to add it here.

data-es2016plus.js
ljharb: (unless they're manually confirmed, in which case it's preferred to add them here)

afmenez (changes requested):

data-es2016plus.js
cdrikd: I have removed the opera browser in commit [8e40474](https://github.com/compat-table/compat-table/pull/1907/commits/8e40474707acbcea4b396e922eca44df2e9b0ccb).

data-es2016plus.js
cdrikd: In `environments.json` the edge version have theirs corresponding chrome version, would you like that I remove this edge entries like opera ?

ljharb: If you haven't manually verified a result yourself, it shouldn't be present in the file.

chicoxyzzy (approved):

ljharb (approved):
if any results need to be added or removed, let's do that in a followup PR.

data-es2016plus.js
cdrikd: Ok, I understand. For most of the browser I only check on these sites : https://caniuse.com/?search=tosorted and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted .
You always validate  by running code on real browser ?

ljharb: Yes, otherwise there's no point in having the table, everyone could just use caniuse :-)

PR #1931, opened 2024-10-15
Implement the RegExp Escaping test

afmenez:
I am not sure about the 'significance' field.


chicoxyzzy (approved):
I think that significance should be `medium` here

ljharb (approved):

PR #1932, opened 2024-10-15
Implement the "Uint8Array to/from base64 and hex" tests

afmenez:
I am not sure about the 'significance' field.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -958,6 +958,97 @@ exports.tests = [
     firefox115: false,
   }
 },
+{
+  name: 'Uint8Array to/from base64 and hex',
+  category: STAGE3,
+  significance: 'small',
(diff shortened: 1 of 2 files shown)


ljharb:
The tests from https://github.com/es-shims/es-arraybuffer-base64/tree/main/test could help make these more thorough.

(There's also 6 methods, and you've only got 4 methods added here)

afmenez:
I have added the 2 missing tests, but I don't believe the tests here need to be so complete... are we going for conformance tests or basic functionality tests?

ljharb:
They definitely don't need to be fully complete - more like, find some of the edge cases, and exercise all of the options.

chicoxyzzy (approved):

ljharb (approved):
I'd still like to see tests that exercise some of the options but this is a great start :-)

PR #1936, opened 2024-10-18
Fixed the `Promise.try` test.

boenrobot:
Instead of "arguments", "args" was used.

An array was used instead of a variadic, as per spec.

Test now fails early if "try" is not found in Promise.

The results of the two promises are checked on a second tick and call asyncTestPassed() on success.


ljharb:
Good catch, thanks

afmenez:
Fixes #1957.

ljharb (approved):

PR #1938, opened 2024-10-18
Fix RegExp flags tests

boenrobot:
Fixed the ```RegExp Match Indices (`hasIndices` / `d` flag)``` / `shows up in flags` test. It was testing for all flags being in the RegExp prototype, in order. This is already done by a separate test (`Proxy, internal 'get' calls` / `RegExp.prototype.flags`), which is still present in `data-es6.js`, and still failing on newer V8 versions.

Support for the now fixed test is assumed to start the same versions as the constructor subtest, and is copied from there. Actually verified on Edge 129, Chrome 130 and Firefox 131.

Added a note on `Proxy, internal 'get' calls` / `RegExp.prototype.flags` about newer Chrome and NodeJS versions failing due to wrong property order.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -6936,22 +6936,10 @@ exports.tests = [
       {
         name: 'shows up in flags',
         exec: function () {/*
-          var expected = ['hasIndices'];
-          // Sorted alphabetically by shortname – "dgimsuy".
-          if ('global' in RegExp.prototype) expected.push('global');
-          if ('ignoreCase' in RegExp.prototype) expected.push('ignoreCase');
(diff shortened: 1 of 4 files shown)


ljharb:
I'm confused what's being fixed here - it's perfectly fine if the same bug causes multiple tests to fail.

ljharb:
The ordering is part of the support - if the order is wrong, it doesn't support hasIndices.

ljharb:
I can see the value, though, of separating the tests - so that for both hasIndices and unicodeSets, there's one test that doesn't care about the order, and one that does?

ljharb:
Yes, exactly.

PR #1942, opened 2024-10-19
Fix links and Firefox results for Promise.try

afmenez:
Ditto.  The test is not working, the results here are after the fix on #1936.


data-es2016plus.js
```diff
@@ -8326,7 +8328,13 @@ exports.tests = [
...
     res: {
       chrome128: true,
-      firefox132: firefox.nightly,
+      firefox115: false,
+      firefox131: false,
+      firefox132: {
+            val: 'flagged',
+            note_id: 'ff-promise-try',
+            note_html: 'The feature is only available on Nightly builds, and has to be enabled via <code>javascript.options.experimental.promise_try</code> setting under <code>about:config</code>.'
+          },
```
```suggestion
        val: 'flagged',
        note_id: 'ff-promise-try',
        note_html: 'The feature is only available on Nightly builds, and has to be enabled via <code>javascript.options.experimental.promise_try</code> setting under <code>about:config</code>.'
      },
```

afmenez:
I had to rebase to commit the requested changes, but now the indentation is fixed.

chicoxyzzy (approved):

ljharb (approved):

PR #1949, opened 2024-11-11
Fix some tests which use "function" instead of "Function"

gergocs:
The following tests failed because of using `function` instead of `Function`
- es2016plus -> exponentiation (**) operator -> early syntax error for unary negation without parens: test result out of date, res: true, actual: false
- es2016plus -> strict fn w/ non-strict non-simple params is error: test result out of date, res: true, actual: false
- es2016plus -> async functions -> no line break between async and function: test result out of date, res: true, actual: false
- es2016plus -> async functions -> must await a value: test result out of date, res: true, actual: false
- es2016plus -> async functions -> cannot await in parameters: test result out of date, res: true, actual: false
- es2016plus -> globalThis -> "globalThis" global property is global object: test result out of date, res: true, actual: false
- es2016plus -> globalThis -> "globalThis" global property has correct property descriptor: test result out of date, res: true, actual: false

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -82,7 +82,7 @@ exports.tests = [
         exec: function () {/*
           if (2 ** 3 !== 8) { return false; }
           try {
-            function ("-5 ** 2")();
+            Function ("-5 ** 2")();
           } catch (e) {
             return true;
           }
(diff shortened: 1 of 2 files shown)


ljharb (approved):
oops, good catch, this was my mistake.

ljharb:
Since this is hopefully an uncontroversial bug fix, I'll go ahead and land it ASAP.

PR #1951, opened 2024-11-18
Firefox 134 has full support for Promise.try

afmenez:
Note that the test is currently broken, it doesn't work on Firefox or Chrome.  Tested with the fix in #1936.

### DIFF
--- a/data-es2016plus.js
+++ b/data-es2016plus.js
@@ -8343,6 +8343,7 @@ exports.tests = [
         note_id: 'ff-promise-try',
         note_html: 'The feature is only available on Nightly builds, and has to be enabled via <code>javascript.options.experimental.promise_try</code> setting under <code>about:config</code>.'
       },
+      firefox134: true,
       node23: true,
     }
   }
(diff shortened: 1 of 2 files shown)


ljharb:
let's fix it in this PR - `s/args/arguments`, basically

afmenez:
Done the replacement, but the test still fails on both Chrome and Firefox.

data-es2016plus.js
```diff
@@ -8328,9 +8328,8 @@ exports.tests = [
       var p = Promise.try(function () { called = true; })
       var p2 = Promise.try(function () {
         'use strict';
-        argsMatch = this === undefined && arguments.length === 2 && args[0] === p && args[1] === 2;
+        argsMatch = this === undefined && arguments.length === 2 && arguments[0] === p && arguments[1] === 2;
       }, [p, 2]);
```
```suggestion
      }, p, 2);
```
oops, one more fix.

afmenez:
Isn't the fix in #1936 complete already?

ljharb:
ah, yes - in that case no need to include those changes here.

afmenez:
Removed the partial fix, back to the original version.

ljharb:
→ #1952

PR #1962, opened 2025-01-09
Rhino v1.8.0

gausie:
- **Updates for Rhino v1.8.0**
- **Update instructions**
- **Build**

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -5151,6 +5151,14 @@
     "release": "2024-05-03",
     "obsolete": true
   },
+  "rhino1_8_0": {
+    "full": "Rhino 1.8.0",
+    "short": "Rhino 1.8.0",
+    "family": "Rhino",
+    "platformtype": "engine",
+    "release": "2025-01-03",
+    "obsolete": false
+  },
   "besen": {
     "full": "Bero's EcmaScript Engine (version 1.0.0.489)",
     "short": "BESEN",
--- a/rhino.js
+++ b/rhino.js
@@ -1,6 +1,5 @@
 /*
- *  Node.js test runner for running data-*.js tests with Rhino's interpreter
- *  running in es6 mode (a.k.a.: -version 200).
+ *  Node.js test runner for running data-*.js tests with Rhino's interpreter (use rhino-all.jar distribution).
  * 
  *  If the environment variable JAVA_HOME is defined it will use it to
  *  construct the path to 'java' as $JAVA_HOME/bin/java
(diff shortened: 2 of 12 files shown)


rhino.js
```diff
@@ -31,7 +30,7 @@ if (process.env.JAVA_HOME) {
 }
 
 function executeScript(scriptName) {
-    return child_process.execFileSync(javaCommand, ['-jar', 'rhino.jar', '-version', '200', scriptName], {
```
ljharb, line 34: can you elaborate on this change?

gausie: Yes! From the release notes:
> The default language level is "VERSION_ES6". That means that modern JavaScript features supported by Rhino will work by default.

And the relevant PR: https://github.com/mozilla/rhino/pull/1594

environments.json
```diff
@@ -5033,6 +5033,14 @@
     "release": "2024-05-03",
     "obsolete": true
   },
+  "rhino1_8_0": {
+    "full": "Rhino 1.8.0",
+    "short": "Rhino 1.8.0",
+    "family": "Rhino",
+    "platformtype": "engine",
+    "release": "2025-01-03",
+    "obsolete": true
```
afmenez: Since this is the latest release, shouldn't it be `"obsolete": false,`?

gausie: I have previously been informed that `obsolete` means that the engine itself is considered outmoded, not that the version is old. Is that no longer the case?

gausie: See: https://github.com/compat-table/compat-table/issues/1687#issuecomment-729878415

ljharb: If Rhino is continuing to have new releases, despite the existence of Nashorn and Graal, then perhaps my comment there is no longer correct.

gausie: Ok! Delighted to set that to false

tonygermano:
Would love to see some movement on this one if there are no other pending requests?

razedator1999:
> Would love to see some movement on this one if there are no other pending requests?

Yeah, @chicoxyzzy hasn't been around since April 9, 2025, I wonder - is the Kangax compatibility table even alive? If not, how do you use test262 on Linux and Windows so that it passes all its tens of thousands of tests quickly, not in days, while only selecting tests of Stage 4 offerings, ES5, ES6 and ES2016+ standards: all 4 options individually. If the Kangax compatibility table is no longer relevant, make a fork of this Kangax compatibility table and develop it separately, as I haven't seen an update to this table in almost 5 months. I use this Kangax compatibility table to evaluate the compatibility level of JS-ES2016+ in various browsers including mobile browsers.

ljharb:
@razedator1999 yes, it is alive. One month isn't all that much time for an unpaid maintainer to take a break. 

I will be proposing some governance changes soon that will allow some kinds of PRs to be added with much less friction, which should allow me to land the majority of the open ones, and moving forward.

afmenez (approved):
LGTM

ljharb (approved):

PR #1970, opened 2025-05-14
fix Promise.try

ilogico:
`Promise.try` is not supposed to receive the arguments to the callback packed in an array.
See the [Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/try) and the [specification](https://tc39.es/proposal-promise-try/#sec-promise.try)


ilogico:
#1936

PR #1974, opened 2025-05-21
Node.js 24 released

afmenez:
Node.js 24 has been released this month.
Obsoleting ancient versions.

### DIFF
--- a/environments.json
+++ b/environments.json
@@ -5368,7 +5368,7 @@
     "note_id": "harmony-flag-old",
     "note_html": "Flagged features have to be enabled via <code>--harmony</code> flag",
     "release": "2011-08-26",
-    "obsolete": true,
+    "obsolete": "very",
     "test_suites": [
       "es5",
       "es6",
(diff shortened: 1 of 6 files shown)


environments.json
```diff
@@ -6150,7 +6152,7 @@
     "platformtype": "engine",
     "equals": "chrome113",
     "release": "2023-04-18",
-    "obsolete": false,
+    "obsolete": true,
```
ljharb: node 20 is still in LTS, so it is decidedly not obsolete.

environments.json
```diff
@@ -6100,6 +6101,7 @@
     "platformtype": "engine",
     "equals": "chrome102",
     "release": "2022-06-02",
+    "obsolete": "very",
```
ljharb: node 18 just went EOL one month ago; i think `true` is more appropriate.

environments.json
```diff
@@ -6201,6 +6203,23 @@
     "platformtype": "engine",
     "equals": "chrome129",
     "release": "2024-10-16",
+    "obsolete": true,
```
ljharb: node 23 is still in maintenance, per https://nodejs.org/en/about/previous-releases

ljharb (approved):

PR #1986, opened 2025-08-18
Map.prototype.upsert moved to Stage 3

afmenez:
Ditto.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -288,7 +288,7 @@ exports.tests = [
 },
 {
   name: 'Map.prototype.upsert',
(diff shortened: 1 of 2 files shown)


nicolo-ribaudo:
The proposal does not have an `.upsert` method anymore: https://github.com/tc39/proposal-upsert

ljharb:
indeed; let's update the tests before we update the stage

afmenez:
OK. I'll let someone with more knowledge on the matter do this.

afmenez:
The proposal was now moved to stage 4.

PR #1992, opened 2025-10-07
Externalize tests and add auto update option

### LINKED ISSUE #1879: Externalize results from tests

p-bakker:
Rebased continuation of #1881, ready for review and hopefully merge

Sry it took a while

@ljharb @chicoxyzzy @JLHwung @nicolo-ribaudo 

Closes #1879 
Makes #1809 obsolete


p-bakker:
The pretest stage seems to fail, due to using somewhat more modern EcmaScript features like const and arrow functions in the .js files that are used to run builds with NodeJS

As part of this PR I've updated the NodeJs dependency to 18.3, as to be able to use the parseArgs feature, as per earlier discussions, see https://github.com/compat-table/compat-table/pull/1791#discussion_r805181740

I think it ought to be fine to use more modern syntax for build.js, runner_support.js, the runners (rhino, duktape, hermes, jerryscript, nashorn and graalvm) and the .js files in build-utils, as I assume those will be run with a modern version of NodeJs anyway?

p-bakker:
@ljharb @chicoxyzzy any input on the questions posed in https://github.com/compat-table/compat-table/pull/1992#issuecomment-3375503095?

Would like to wrap up this PR and move onto the next stage

p-bakker:
@ljharb @chicoxyzzy any chance to get some input on the questions posted above wrt allowing newer ES syntax for the runner files, so this PR can be moved forwards?

p-bakker:
@ljharb @chicoxyzzy just another ping... Would be so nice to be able to progress on this....

ljharb:
I rebased this, but I think there's been some conflicts. Can you confirm?

PR #1999, opened 2025-11-18
Node 25 released and add "using bound function" sub test case

### LINKED ISSUE #1994: Add `Symbol.asyncDispose` bound function bug (`TypeError: Symbol(Symbol.asyncDispose) is not a function`)
The following code errors with `TypeError: Symbol(Symbol.asyncDispose) is not a function` on v8 13.6.233.10-node.28 (Node 24.10.0), but passes with v8 14.1.146.11 (Chrome 141.0.7390.70). I guess there was a bug between these versions. I haven't found a bug report in chromium's issue tracker though.
```js
{
	async function close() {
		console.log('called')
	}
	await using n = {
		[Symbol.asyncDispose]: close.bind(null)
	};
}
```
I'm not sure if it's better to add this case in the table, but thought to report this anyways.

sapphi-red:
This PR adds the subtest. This fails with Node 24.11. ~~I didn't add any `res` entry as I wasn't sure how that should be tested.~~ added node 24 & 25 results

https://nodejs.org/ja/blog/release/v25.0.0

fixes #1994

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1244,6 +1244,28 @@ exports.tests = [
       }
     },
     {
+      name: 'Symbol.dispose with bound function',
+      exec: function () {/*
+        var disposed = false;
+        var resource = {
+          [Symbol.dispose]: close.bind(null),
+        };
+        function close() {
+          disposed = true;
+        }
+        {
+          using _ = resource;
+        }
+        return disposed;
+      */},
+      res: {
+        chrome143: true,
+        firefox147: true,
+        node24_0: false,
+        node25_0: true,
+      }
+    },
+    {
       name: "SuppressedError",
       exec: function () {/*
         var err1 = new Error();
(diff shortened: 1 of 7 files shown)


ljharb:
If it fails on a version of node with manual testing, please do add the entry reflecting it :-)

sapphi-red:
I've added the results. I also added node 25 as that passed the test. Please let me know if I should split the PR into two.

nicolo-ribaudo (commented):
This passes in Firefox 147

ljharb (approved):

data-esnext.js
```diff
@@ -1243,6 +1243,28 @@ exports.tests = [
...
+          disposed = true;
+        }
+        {
+          using _ = resource;
+        }
+        return disposed;
+      */},
+      res: {
+        chrome143: true,
+        firefox147: true,
```
JLHwung, line 1263: I can confirm that firefox 141, the first version with unflagged ERM support, is not affected by this bug. I will update the compat-data later today.

PR #2001, opened 2025-12-03
Fix correct prototype chains test

### LINKED ISSUE #1996: typed arrays 'correct prototype chains' snippet broken with arraybuffer-base64 proposal
`"BYTES_PER_ELEMENT,constructor"` check at [data-es6.js:7604](https://github.com/compat-table/compat-table/blob/gh-pages/data-es6.js#L7604) is broken for many recent engines  (v8, jsc, spidermonkey, libjs, escargot, xs and kiesel among what I tested), because there are new methods being introduced: `setFromBase64,setFromHex,toBase64,toHex`

https://tc39.es/proposal-arraybuffer-base64/spec/
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/setFromBase64

ivankra:
arraybuffer-base64 proposal introduces 4 new typed array methods (setFromBase64, setFromHex, toBase64, toHex) that break equality check in this test.

Fixes #1996

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -7600,8 +7600,8 @@ exports.tests = [
           if (!(constructors[i] in global
               && Object.getPrototypeOf(global[constructors[i]]) === constructor
               && Object.getPrototypeOf(global[constructors[i]].prototype) === prototype
-              && Object.getOwnPropertyNames(global[constructors[i]].prototype).sort() + ''
-                === "BYTES_PER_ELEMENT,constructor")) {
+              && Object.getOwnPropertyNames(global[constructors[i]].prototype).indexOf('BYTES_PER_ELEMENT') !== -1
+              && Object.getOwnPropertyNames(global[constructors[i]].prototype).indexOf('constructor') !== -1)) {
             return false;
           }
         }


data-es6.js
```diff
@@ -7600,8 +7600,8 @@ exports.tests = [
           if (!(constructors[i] in global
               && Object.getPrototypeOf(global[constructors[i]]) === constructor
               && Object.getPrototypeOf(global[constructors[i]].prototype) === prototype
-              && Object.getOwnPropertyNames(global[constructors[i]].prototype).sort() + ''
-                === "BYTES_PER_ELEMENT,constructor")) {
+              && Object.getOwnPropertyNames(global[constructors[i]].prototype).indexOf('BYTES_PER_ELEMENT') !== -1
+              && Object.getOwnPropertyNames(global[constructors[i]].prototype).indexOf('constructor') !== -1)) {
```
ljharb, line 7604: let's also compare the indexes, and ensure that the former is in front of the latter?

ljharb (changes requested):

data-es6.js
ivankra: I don't think spec guarantees a specific order here. Object.getOwnPropertyNames() returns keys in insertion order but that might be engine-specific, probably that's why the original had .sort() there.

ivankra:
```
$ ./kiesel 
Kiesel 0.1.0-dev+073f5d0a6 [Zig 0.15.2] on linux
Use Ctrl+D to exit.
> Object.getOwnPropertyNames(Int32Array.prototype)
[ "BYTES_PER_ELEMENT", "constructor" ]
> Object.getOwnPropertyNames(Uint8Array.prototype)
[ "BYTES_PER_ELEMENT", "constructor", "setFromBase64", "setFromHex", "toBase64", "toHex" ]

$ ./v8
V8 version 14.4.258.13
d8> Object.getOwnPropertyNames(Int32Array.prototype)
["constructor", "BYTES_PER_ELEMENT"]
d8> Object.getOwnPropertyNames(Uint8Array.prototype)
["constructor", "BYTES_PER_ELEMENT", "toBase64", "setFromBase64", "toHex", "setFromHex"]

$ ./jsc
>>> Object.getOwnPropertyNames(Int32Array.prototype)
[BYTES_PER_ELEMENT,constructor]
>>> Object.getOwnPropertyNames(Uint8Array.prototype)
[BYTES_PER_ELEMENT,setFromBase64,setFromHex,toBase64,toHex,constructor]
```

ljharb: It's not engine-specific; the spec definitely guarantees it. If implementations don't agree, there's a bug somewhere.

ivankra: How from the spec did you infer that `constructor` must come before `BYTES_PER_ELEMENT` or vice versa here?

ljharb: Because "insertion order" is guaranteed since ES5, and the spec "inserts" properties in alphabetical order, and BYTES_PER_ELEMENT comes before constructor. If the spec is unclear, I'm happy to make it more clear (I'm on TC39) - but that is absolutely the intention and the way test262 tests are authored.

ivankra: > the spec "inserts" properties in alphabetical order

Where does it say that for internal objects their properties must be populated alphabetically?

I'm actually not even sure this check really belongs to the test. From the description, I'd say
"Object.getPrototypeOf(global[constructors[i]]) === constructor && Object.getPrototypeOf(global[constructors[i]].prototype) === prototype" is the main thing being tested, and "BYTES_PER_ELEMENT,constructor" just a side test probably added to make sure it got the correct prototype.

ljharb: The ordering in the spec determines the insertion order; it's kept alphabetically.

I'd be fine adding a separate test that confirms the order is correct - you're right that _this_ test doesn't need to cover it.

ivankra: Yea, if spec mandates an order there, it would make more sense to add a separate test for it.

Here are the results of console.log(Object.getOwnPropertyNames(Uint8Array.prototype)) for a few more engines I test - I see large variations between implementations:

```
boa:          [ "BYTES_PER_ELEMENT", "constructor" ]
duktape:      ["constructor","BYTES_PER_ELEMENT"]
engine262:    ['BYTES_PER_ELEMENT', 'constructor', 'toBase64', 'setFromBase64', 'toHex', 'setFromHex']
graaljs:      BYTES_PER_ELEMENT,constructor
hermes:       constructor,BYTES_PER_ELEMENT
libjs:        [ "BYTES_PER_ELEMENT", "setFromBase64", "setFromHex", "toBase64", "toHex", "constructor" ]
quickjs:      [ "BYTES_PER_ELEMENT", "constructor" ]
qv4:          [constructor,BYTES_PER_ELEMENT]
rhino:        ["constructor","BYTES_PER_ELEMENT"]
spidermonkey: [setFromBase64,setFromHex,toBase64,toHex,BYTES_PER_ELEMENT,constructor]
v8:           constructor,BYTES_PER_ELEMENT,toBase64,setFromBase64,toHex,setFromHex
```

ljharb: Then it'd be great to have a test :-) once we have one here, it'd be great to add it to test262 also.

Very few things in the spec are actually permitted to be left up to individual implementations, and object key ordering hasn't been one of them for 16 years.

ivankra: Sounds like that test would properly belong to data-es5.js and should cover other internal objects as well, as I doubt Uint8Array.prototype would be the only one having this problem. I think it should be a separate PR.

ljharb: The data-es5 test would need to test the relative order of only the properties present in ES5; and newer editions that added properties would need to test the relative orderings of the added ones.

It's fair that it should perhaps be a separate PR.

ivankra:
Rebased and ran `npm run build`

ljharb (approved):

PR #2004, opened 2025-12-26
Improve ERM subtests

JLHwung:
As a follow-up to https://github.com/compat-table/compat-table/pull/1999, this PR adds more support data to the test results.

### DIFF
--- a/data-esnext.js
+++ b/data-esnext.js
@@ -1263,7 +1263,10 @@ exports.tests = [
         return disposed;
       */},
       res: {
+        chrome140: false,
+        chrome141: true,
         chrome143: true,
+        firefox141: true,
         firefox147: true,
         node24_0: false,
         node25_0: true,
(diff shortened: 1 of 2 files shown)


ljharb (approved):

JLHwung:
@ljharb Is this PR ready to merge? Thanks.

data-esnext.js
```diff
@@ -1259,7 +1259,10 @@ exports.tests = [
         return disposed;
       */},
       res: {
+        chrome140: false,
+        chrome141: true,
         chrome143: true,
+        firefox141: true,
```
afmenez, line 1269: I tested with Firefox 115, it passed.

JLHwung: This is very strange. The ERM implementation was landed in FF 134 behind a flag, which was then unflagged in FF 141. The test should not have passed in FF 115. I will check if there is something wrong in the test itself.

JLHwung: Update: I can not reproduce the result on FF 115 mac version, all subtests fail as expected:
image

afmenez: My bad. I installed FF115 to run the test but it updated itself.  I reinstalled it and got the same results as you.

JLHwung: Thank you. Yeah the auto-update is pretty annoying for test purpose. I wish FF can offer a "Firefox for testing" build that always disables the auto-update.

ljharb (approved):

PR #2008, opened 2026-02-04
Add safari destructuring bug

JLHwung:
Added new test cases for a Safari destructuring bug, affecting Safari from 9 to 13.1.

Due to the limit of the BrowserStack, I can not verify whether this bug has been fixed in Safari 14 or not. The earliest version I can verify is Safari 14.1. If you have access to Safari 14.0, feel free to suggest the test result and I will update the PR accordingly.

/cc @magic-akari, who reported this bug to Babel.

Context: https://github.com/babel/babel/issues/17773#issuecomment-3842765704

### DIFF
--- a/data-es6.js
+++ b/data-es6.js
@@ -15218,6 +15218,30 @@ exports.tests = [
       }
     },
     {
+      name: 'rest with identical number of elements in RHS',
+      exec: function(){/*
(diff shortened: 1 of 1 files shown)


ljharb (approved):
i also have 14.1 and 13, but not 14.0.

